PR #61, opened 2019-04-15
BlueZ fixes for develop branch

dlech:
I've been trying out the develop branch and ran into a few issues with BlueZ. See commit messages for details. I can split these into individual pull requests if that is helpful.

### DIFF
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 import logging
 import asyncio
+import os
 import re
 import subprocess
 from functools import wraps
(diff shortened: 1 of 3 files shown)


hbldh:
It is good to have them gathered like this. These are all interesting fixes for hard to detect problems!

A question: the write without response problem in 94163fa, is it not better to have the old solution (WriteValue without type in options dict) instead of the AcquireWrite? The AquireWrite is also added first in ~5.46 so it is not a backwards compatible solution either. But I had missed that the type parameter was added in 5.50; I had just assumed that it had been present since the beginning...

dlech:
> is it not better to have the old solution (WriteValue without type in options dict) instead of the AcquireWrite?

I would say no. If the user requested a write without response and we give them a write with response, it can cause unexpected bugs. For example, the device I am working with would get disconnected because of a Bluetooth protocol bug in the device. I think it would be better to just raise a RuntimeError for older BlueZ versions that don't support this writing without response.

> But I had missed that the type parameter was added in 5.50

I think it was added *after* 5.50 - it is not in a released version yet.

PR #66, opened 2019-05-21
Fix IndexError in bluez client

5frank:
There seems to be no guarantee that services are listed before
characteristics in bluez dbus API. This caused (sometimes) a IndexError
in get_services() as adding a characteristic to a service that did not
exist yet failed.
Not sure if it's the same with descriptors, could possible be in the same 
loop as characteristics?

### DIFF
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -176,6 +176,9 @@ class BleakClientBlueZDBus(BaseBleakClient):
             self._bus, self.loop, self._device_path + "/service"
         )
 
+        # no guarantee service listed before charachteristics in objs
+        # need multiple iterations
+
         for object_path, interfaces in objs.items():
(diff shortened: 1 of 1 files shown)


5frank:
Traceback on the issue (line numbers might be off):
'''
  [...]
    async with BleakClient(address, loop=loop) as client:
  File "bleak/bleak/backends/client.py", line 41, in __aenter__
    await self.connect()
  File "bleak/bleak/backends/bluezdbus/client.py", line 107, in connect
    await self.get_services()
  File "bleak/bleak/backends/bluezdbus/client.py", line 199, in get_services
    char, object_path, _service[0].uuid
IndexError: list index out of range
'''
API usage example from bluez that iterate twice on the same objects but without a explanation on why: https://github.com/hadess/bluez/blob/master/test/example-gatt-client

hbldh:
Merged this to develop and rewrote it slightly. Will be merging this to master during next week after I have tried it a bit more. Thank you!

PR #80, opened 2019-07-03
Added development for macOS

kevincar:
I've started putting together a macOS implementation here. 

The biggest difference between the other backends, however, is that CoreBluetooth doesn't identify devices using [MAC addresses](https://forums.developer.apple.com/thread/8442), but rather through device UUIDs.


hbldh:
This looks amazing! Just yesterday, I set up my Mac mini in preparation of this summer's main project: the Core Bluetooth backend. I am very happy to start from this instead! I will merge this and start looking at it tonight and get back to you here as soon as possible.

Implementation of #3

hbldh:
Generally this works wonderfully, thank you so much! It can definitely be merged to `develop` in the current state. I plan on continuing the discussion about this implementation and its further development in #3.

Issue #82, opened 2019-07-07
Connection handling enhancements


JoeHut:
In Bluez it should be possible to listen for an event if a device has disconnected (https://stackoverflow.com/questions/53664144/device-connect-disconnect-notification-via-bluez-dbus-api). Are there plans to add an disconnection callback to the client API (I have no idea whether this is possible on the other platforms) or even try to implement something like auto-reconnection in the library itself?

hbldh:
There are no plans, no. It is a good idea though and I do not think it is very hard to do, at least not in the BlueZ backend. 

The auto-reconnection is somewhat breaking the idea of BLE, is it not? If the connection is not used, better to let it die and let the peripheral save energy until it is actually needed and let the client reconnect then?

JoeHut:
I could provide a PR for the API and a BlueZ implementation, but I have no idea about .NET or the Darwin backend...

Out of experience there are multiple reasons why a BLE connection might drop, especially if the connection intervals are not properly set up. Typical cases are interferences and reflections in crowded environments and getting out of range. We always try to have a very resilient auto-reconnection on the peripheral side since you typically want to have long-lived connections where you can react to events from the peripheral in real time. But this is of course very use-case specific.

Anyways, this logic can be definitely pushed to the application if we have a proper disconnect callback.

hbldh:
This is implemented and merged into version 0.5.0 which has now been deployed to PyPI.

PR #85, opened 2019-07-10
Add a disconnect callback for BlueZ

arthur-proglove:
This tries to fix issue #82 for BlueZ implementation only.
Other implementations (Core Bluetooth and .NET) will fail with a
NotImplementedError exception.

The disconnect event is found in the DBus _PropChanged_:
> path: /org/bluez/hci0/dev_XX_XX_XX_XX_XX_XX, domain: org.bluez.Device1, body: {'ServicesResolved': False, 'Connected': False}

The callback will be called only on unsolicited disconnect event from
peer. An initiated disconnect from the host will not call it.

### DIFF
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -537,6 +537,12 @@ class BleakClientBlueZDBus(BaseBleakClient):
                 the new data on the GATT Characteristic.
 
         """
+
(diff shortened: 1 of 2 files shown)


arthur-proglove:
I am not an expert of `async` and `await` usage. 
How can we call `self._bus.delMatch('PropChanged')` in the `_properties_changed_callback` routine to act the same as the `disconnect` routine?

PR #86, opened 2019-07-12
Register a disconnected callback and clean up subscriptions on disconnect

### LINKED ISSUE #82: Connection handling enhancements

### LINKED ISSUE #83: Duplicated notifications when reconnecting
* bleak version: 0.4.3
* Python version: Python 3.7.3
* Operating System: Ubuntu 19.04 (Kernel 5.0.0-20-generic)

### Description

I'm trying to maintain a connection to a BLE device permanently. In case of connection lost (caused by an out of range o standby situation), I am reconnecting and resubscribing for the same notifications callback.

To accomplish this, my code looks something like this (simplified):

```
while self.running
    async with BleakClient(device_mac, device="hci0", timeout=3) as bleak:
        bleak.start_notify("FANCY UUID", callback=notification_handler)
        while self.running and await bleak.is_connected():
            await asyncio.sleep(2)
            
def notification_handler(self, uuid, value):
    # do neat stuff
```
This seems to work, but every time I use start_notify  the **notification_handler** method is triggered one additional time. Just to clarify: First connection, **notification_handler** is called 1 time per notification; second connection, **notification_handler** is called twice; an so on.

I verified the code and seems the problem is that DBUS is sending the notifications duplicated N times because of the resubscriptions. The cause of this is that when it's connecting, the library subscribes to **PropertiesChanged** rule which is only removed in case of controlled disconnection. I also tried to call **disconnect** function in case of uncontrolled disconnection but an exception is raised in such case.

```Traceback (most recent call last):
  File "/snap/pycharm-community/125/helpers/pydev/pydevd.py", line 1741, in <module>
    main()
  File "/snap/pycharm-community/125/helpers/pydev/pydevd.py", line 1735, in main
    globals = debugger.run(setup['file'], None, None, is_module)
  File "/snap/pycharm-community/125/helpers/pydev/pydevd.py", line 1135, in run
    pydev_imports.execfile(file, globals, locals)  # execute the script
  File "/snap/pycharm-community/125/helpers/pydev/_pydev_imps/_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "/home/marcx/btest/main.py", line 142, in <module>
    asyncio.get_event_loop().run_until_complete(main())
  File "/usr/lib/python3.7/asyncio/base_events.py", line 584, in run_until_complete
    return future.result()
  File "/home/marcx/btest/main.py", line 93, in main
    await task
  File "/home/marcx/btest/bleak_heart_rate_device.py", line 83, in start
    await self.bleak.disconnect()
  File "/home/marcx/btest/venv/lib/python3.7/site-packages/bleak/backends/client.py", line 45, in __aexit__
    await self.disconnect()
  File "/home/marcx/btest/venv/lib/python3.7/site-packages/bleak/backends/bluezdbus/client.py", line 131, in disconnect
    await self._bus.delMatch(rule_id).asFuture(self.loop)
  File "/home/marcx/btest/venv/lib/python3.7/site-packages/txdbus/client.py", line 185, in delMatch
    rule = self.match_rules[rule_id]
KeyError: 1
```

Not sure what to do here. I'm doing something wrong? It is a bug?

Thanks!

JoeHut:
This patch contains two changes:
* Add API and Bluez implementation to set a callback fired when a device disconnects without active intervention from the client
* Remove all existing subscriptions to characteristic notifications when a device disconnects

Replaces #85 
Fixes #83 
Closes #82

### DIFF
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -134,6 +134,11 @@ class BleakClientBlueZDBus(BaseBleakClient):
         )
         return True
 
+    async def _cleanup(self) -> None:
+        for rule_name, rule_id in self._rules.items():
+            logger.debug("Removing rule {0}, ID: {1}".format(rule_name, rule_id))
+            await self._bus.delMatch(rule_id).asFuture(self.loop)
+
     async def disconnect(self) -> bool:
         """Disconnect from the specified GATT server.
 
@@ -142,9 +147,9 @@ class BleakClientBlueZDBus(BaseBleakClient):
 
         """
         logger.debug("Disconnecting from BLE device...")
-        for rule_name, rule_id in self._rules.items():
-            logger.debug("Removing rule {0}, ID: {1}".format(rule_name, rule_id))
-            await self._bus.delMatch(rule_id).asFuture(self.loop)
+
+        await self._cleanup()
+
         await self._bus.callRemote(
             self._device_path,
             "Disconnect",
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -537,6 +537,12 @@ class BleakClientBlueZDBus(BaseBleakClient):
                 the new data on the GATT Characteristic.
 
         """
+
+        logger.debug('DBUS: path: {}, domain: {}, body: {}'
+                     .format(message.path,
+                             message.body[0],
+                             message.body[1]))
+
         if message.body[0] == defs.GATT_CHARACTERISTIC_INTERFACE:
             if message.path in self._notification_callbacks:
                 logger.info(
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -40,6 +40,8 @@ class BleakClientBlueZDBus(BaseBleakClient):
         self._bus = None
         self._rules = {}
 
+        self._disconnected_callback = None
+
         self._char_path_to_uuid = {}
 
         # We need to know BlueZ version since battery level characteristic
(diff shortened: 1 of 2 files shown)


bleak/backends/bluezdbus/client.py
```diff
@@ -39,6 +39,9 @@ def __init__(self, address, loop=None, **kwargs):
         self._device_path = None
         self._bus = None
         self._rules = {}
+        self._subscriptions = list()
+
+        self.disconnect_callback = None
```
hbldh: Could you change this to an internal attribute, please?

bleak/backends/client.py
```diff
@@ -46,6 +46,20 @@ def __repr__(self):
 
     # Connectivity methods
 
+    @abc.abstractmethod
+    async def set_disconnect_callback(
```
hbldh: I would like some documentation of an example of a `disconnect_callback` that shows how it is used. Place it here, in the docstring of the interface class.

bleak/backends/bluezdbus/client.py
hbldh: Could you change this to an internal attribute, please?

bleak/backends/client.py
JoeHut: Makes sense. I'll add it tomorrow.

JoeHut:
@hbldh I implemented the changes and rebased on the develop branch. I was not aware of this branch before...

PR #90, opened 2019-07-16
Fix leaking of callbacks from DBus on multiple calls to discover()

### LINKED ISSUE #89: Memory leaks through missing callback deregistration in Bluez
* bleak version: master
* Python version: 3.7.3
* Operating System: Linux
* Bluez version: 5.50

### Description

In my use case I have a long running application that is looking for devices with certain properties (like a name prefix, RSSI etc.) and connects to them. To accomplish this I have an infinite loop that calls discover and checks the list of devices until the proper one is available. Something like:

```python
while True:
    devices = await discover(timeout=1)
    for d in devices:
        if d.name.startswith("PREFIX"):
            return d
```

I observed that if it is running for a while, the discover takes not only around a second but up to multiple minutes.

### Explanation

The issue is that the DBus callbacks in discovery() for InterfacesAdded, InterfacesRemoved and PropertiesChanged are never deleted again. Therefore they pile up and the parse_msg callback is called thousands of times after a couple of minutes.

The issue is therefore similar to #83 where the notifications are sent multiple times after a reconnection. I think it makes sense to go through all Bluez callback registrations and make sure they don't leak.

JoeHut:
If the rules are only added to Bluez via DBus but never removed again,
the calls pile up if discover() is called multiple times in one session.
This is especially harmful for long-running applications that connect
and disconnect to the same or different servers over and over. The time
to await discovery() becomes very long (up to minutes) even with smaller
timeouts.

Make sure the callbacks registered are cleaned up before discover() is
left.

Fixes #89

### DIFF
--- a/bleak/backends/bluezdbus/discovery.py
+++ b/bleak/backends/bluezdbus/discovery.py
@@ -72,6 +72,7 @@ async def discover(timeout=5.0, loop=None, **kwargs):
     loop = loop if loop else asyncio.get_event_loop()
     cached_devices = {}
     devices = {}
+    rules = list()
 
     def parse_msg(message):
         if message.member == "InterfacesAdded":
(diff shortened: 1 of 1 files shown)


hbldh:
Merged this into develop branch. Will close this when I release 0.5.0.

PR #92, opened 2019-07-16
Prevent leaking of DBus connections on discovery

### LINKED ISSUE #91: Leak of active DBus connections
* bleak version: master
* Python version: 3.7.3
* Operating System: Linux
* Bluez version: 5.50

### Description

DBus connections are not cleaned up after usage. Multiple calls to discover() result eventually in following error:
```
  File "discover.py", line 29, in <module>
    loop.run_until_complete(run())
  File "/usr/lib64/python3.7/asyncio/base_events.py", line 584, in run_until_complete
    return future.result()
  File "discover.py", line 22, in run
    devices = await discover(timeout=1)
  File "/home/joe/code/proglove/python/bleak/bleak/backends/bluezdbus/discovery.py", line 125, in discover
    bus = await client.connect(reactor, "system").asFuture(loop)
txdbus.error.RemoteError: org.freedesktop.DBus.Error.LimitsExceeded: The maximum number of active connections for UID 1000 has been reached
```

For me this happens at the 246th call to discover()

JoeHut:
During a call to discover() on the Bluez backend, a DBus connection is
opened, but never closed. If discover() is called multiple times while
the application is running, DBus at some point prevents opening
additional connections.

Make sure the bus connection is closed after discovery.

Fixes #91

### DIFF
--- a/bleak/backends/bluezdbus/discovery.py
+++ b/bleak/backends/bluezdbus/discovery.py
@@ -199,4 +199,7 @@ async def discover(timeout=5.0, loop=None, **kwargs):
         manufacturer_data = props.get('ManufacturerData', {})
         discovered_devices.append(BLEDevice(address, name, {"path": path, "props": props}, uuids=uuids,
                                   manufacturer_data=manufacturer_data))
+
+    bus.disconnect()
+
     return discovered_devices


hbldh:
Merged this to `develop` branch. It will be released with version 0.5.0.

PR #96, opened 2019-07-26
Use raw literals for regex patterns

dwolfi:
Should fix this Python 3 related warning:
```
 src/bleak/bleak/backends/bluezdbus/utils.py:10: DeprecationWarning: invalid escape sequence \d
    _hci_device_regex = re.compile("^hci(\d+)$")
```

### DIFF
--- a/bleak/backends/bluezdbus/utils.py
+++ b/bleak/backends/bluezdbus/utils.py
@@ -7,7 +7,7 @@ from bleak.backends.bluezdbus import defs
 from bleak.exc import BleakError
 
 _mac_address_regex = re.compile("^([0-9A-Fa-f]{2}[:-]){5}([0-9A-Fa-f]{2})$")
-_hci_device_regex = re.compile("^hci(\d+)$")
+_hci_device_regex = re.compile("^hci(\\d+)$")
 
 
 def validate_mac_address(address):


bleak/backends/bluezdbus/utils.py
```diff
@@ -6,8 +6,8 @@
 from bleak.backends.bluezdbus import defs
 from bleak.exc import BleakError
 
-_mac_address_regex = re.compile("^([0-9A-Fa-f]{2}[:-]){5}([0-9A-Fa-f]{2})$")
-_hci_device_regex = re.compile("^hci(\d+)$")
+_mac_address_regex = re.compile(r'^([0-9A-Fa-f]{2}[:-]){5}([0-9A-Fa-f]{2})$')
```
hbldh: This row does not need to be changed. The code formatter `black`, which I use for bleak, will change your single quotes to doubles very soon. The string also does not need to be binary. I.e. Remove this change.

bleak/backends/bluezdbus/utils.py
```diff
@@ -6,8 +6,8 @@
...
+_hci_device_regex = re.compile(r'^hci(\d+)$')
```
hbldh: I would rather see this changed to 
```
_hci_device_regex = re.compile("^hci(\\d+)$")
```
That woudl remove the warning and keep it as a regular string.

bleak/backends/bluezdbus/utils.py
dwolfi: done

bleak/backends/bluezdbus/utils.py
dwolfi: fair enough, done

PR #100, opened 2019-08-13
Avoid infinite loop by raising an exception if not connected

arthur-proglove:
Raise an exception in the infinite loop of `get_services` if the field `Connected` is not found in the properties.

Sometimes we can get stuck in this loop if the connection drops between the connect and the service discovery.

Either we should raise an exception there or return and raise it in the caller.


arthur-proglove:
Good point, I added a 5 seconds timeout before failing.

PR #117, opened 2019-10-08
set verbose option to true for python exception to be raised from ObjC

Arrowana:


hbldh:
I will have to try this out on my Mac first before I merge anything. I agree that it seems better to have the Python exception printed above the Core Bluetooth one, but if it is still uncatchable in Python or breaks the execution of the Python program, then it is of little use. 
I will still merge it probably, but I want to examine the situation first.

Arrowana:
I agree that it is of little use for operational purposes, but at least we get to see the source of the crash by default rather than having to wrap the notification handler code into a try except all.

Alternatively, because async/await is in use, we could register that the callback fired and store the arguments then execute the callback once we are back into python, as a result the callback can raise naturally. But then that would mean that we potentially have to do that for every callback from objc => python

hbldh:
Agreed.

The storing of executing methods and callbacks would be too brittle to support I think. This will have to do for now. 

Merged this into develop branch. Will close this modification is relesed with the upcoming 0.5.1 version.

PR #119, opened 2019-10-21
Backend/BlueZ Filter scan by UUID

arthur-proglove:
New extra argument to filter the advertisement data with a UUIDs list.


arthur-proglove:
@hbldh  I created a new PR for that #119 124

PR #122, opened 2019-10-30
Update to txdbus v1.1.1

dlech:
Upstream bug fix has finally been merged and released, so we can
bump the version and remove the comment about the bug.


hbldh:
Very nice. This I will merge this directly. Thank you!

PR #124, opened 2019-11-04
backend(BlueZ): add a discovery filter argument

arthur-proglove:
Allow the user to add any BlueZ filter for the discovery.


hbldh:
This looks better. Will merge this!

PR #126, opened 2019-11-11
Fix macOS BLEDevice.metatadata values

dlech:
This contains some fixes for the corebluetooth backend so that the format of the `BLEDevice.metatadata` values are the same on all platforms.

See commit messages for details.


hbldh:
Amazing and thank you. I will happily accept this!

PR #134, opened 2019-12-06
Asynchronous discovery of devices

JoeHut:
This patch includes two changes:
* Prevent overhead of process change for discovery if timeout = 0 is passed
* asynchronous discovery of nearby BLE devices. While
the blocking discover() function with a timeout is still available, an
additional class provides support for starting and stopping the
discovery session individually with a callback on each device change.

@hbldh I don't know whether you are interested in something like this or how you would want this to be integrated in the library. Just let me know if I should close the PR or change it to be merged upstream

### DIFF
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -106,7 +106,8 @@ class BleakClientBlueZDBus(BaseBleakClient):
         # A Discover must have been run before connecting to any devices. Do a quick one here
         # to ensure that it has been done.
         timeout = kwargs.get("timeout", self._timeout)
-        await discover(timeout=timeout, device=self.device, loop=self.loop)
+        if timeout > 0:
+            await discover(timeout=timeout, device=self.device, loop=self.loop)
 
         # Create system bus
         self._bus = await txdbus_connect(reactor, busAddress="system").asFuture(
--- a/bleak/backends/bluezdbus/discovery.py
+++ b/bleak/backends/bluezdbus/discovery.py
@@ -2,6 +2,7 @@
 
 import asyncio
 import logging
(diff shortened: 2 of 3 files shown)


hbldh:
I like this as an idea and realize that it might be highly useful, but I am not fully onboard with the implementation. I think a `__aenter__` and `__aexit__` solution and using this for the "non-async" discover would work better.

I will not add this to the 0.5.2, I need to think about it some more, and see if the other backends could be used for the same kind of async returning of discovered elements. 

I had never envisioned that one would like to run the discover method for prolonged amounts of time and for much else than finding devices. I think I need to rethink this a bit.

JoeHut:
Well, my use case is basically to have a daemon running and you don't know when a device you want to connect to comes into range. In this case I have to poll all the time with the current discover function, but actually I rather want to have proper filtering in the BlueZ backend and get notified when a device comes into range.

mickael9:
An async generator would be awesome too, so you can do:

```python
async for device in scanner.iter(timeout=5):
    pass
```

hbldh:
Yes, that is my ultimate goal with this, after getting a scanner working for each backend. It does however make restrictions on using python>=3.6. I generally would prefer that, but I have had requests on keeping it compliant with 3.5. Not sure how to do that with a asynchronous iterators yet though.

tannewt:
Hi! I'm interested in this as well. In CircuitPython, our scan function returns a generator that yields results as they come in. CircuitPython doesn't do async yet so it'd be great to work as a non-async generator or iterator too.

Anything I can do to help move this along? Overall, bleak looks really promising.

hbldh:
I will close this in favour of the `BleakScanner` interface added in version 0.6 instead.

Issue #137, opened 2019-12-20
write_gatt_char error. correct bluez version check?


5frank:
* bleak: 0.5.1
* python: 3.5.3
* bluez: 5.43
* os: Linux-4.9.0-11-amd64-x86_64-with-debian-9.11

### Description

`write_gatt_char` (repsonse=False) fails with the following error `txdbus.error.RemoteError: org.freedesktop.DBus.Error.UnknownMethod: Method "AcquireWrite" with signature "a{sv}" on interface "org.bluez.GattCharacteristic1" doesn't exist`.
The following change resolvs the issue:

```diff
diff --git a/bleak/backends/bluezdbus/client.py b/bleak/backends/bluezdbus/client.py
index ac779f8..3585659 100644
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -399,7 +399,7 @@ class BleakClientBlueZDBus(BaseBleakClient):
                 % str(_uuid)
             )
 
-        if response or (self._bluez_version[0] == 5 and self._bluez_version[1] > 50):
+        if response or (self._bluez_version[0] == 5 and self._bluez_version[1] > 42):
             # TODO: Add OnValueUpdated handler for response=True?
             await self._bus.callRemote(
                 characteristic.path,
```

Is the check on bluez version correct? or perhaps debian have some patches in the bluez 5.43 version that causes this to work?

dlech:
The check in your diff is correct because it is for support for the `type` option of the `WriteValue` method, not the `AcquireWrite` that is causing your error.

`AcquireWrite` was added in Bluez 5.46. So in older Bluez versions, it is simply not possible to write without response.

5frank:
ok thanks! It all makes sense now! :)

Issue #129, opened 2019-11-17
wrong version requirement for BlueZ on Linux


afflux:
Since 94163fa1d4929b243311988c420b2bcffb204270 (#61), `write_gatt_char` requires at least BlueZ 5.46, because `AcquireWrite` was not introduced before.

The current version checking logic in [`__init__.py`](https://github.com/hbldh/bleak/blob/11e7adf1ae1774b22667fcc4e6533ad69597951a/bleak/__init__.py#L41) and the readme files all claim that 5.43 is sufficient.

dlech:
To clarify, BlueZ 5.46 is only required to be able to "Write without response".

hbldh:
Documentation in v0.5.2 of Bleak will provide clearer instructions regarding BlueZ version requirements. Thank you for requesting clearer specification of this.

As @dlech says,  the version requirements is higher than 5.43 only in the case of the `write_gatt_char` method, and only if the "Write without response" is desired.

hbldh:
Adressed and fixed in #138, and further in commit b767f9d

afflux:
I'd argue that the default `response=False` did not really make sense, because that's the stricter version requirement - but then again, with the exception in #138, chances are higher for people to find the parameter.

hbldh:
@afflux In my mind, the `response=False` was the simpler choice and thus the default behaviour. The other OS backends have no such version requirement demands, and I don't want to change default behaviour for all of them. 

It is now a regrettable byproduct of a choice I made in the advent of this project.

Issue #106, opened 2019-08-28
Wrong BlueZ version check in write_gatt_char


arthur-proglove:
* bleak version: 0.5.1.dev1
* Python version: 3.7.3
* Operating System: Ubuntu 19.04
* BlueZ version: 5.50

### Description

In the method `write_gatt_char`, the `WriteValue` command is only called if the write request is with response or if the BlueZ version is strictly greater then `5.50`. Shouldn't this be greater or equal to `5.50`?

```
if response or (self._bluez_version[0] == 5 and self._bluez_version[1] > 50):                                                                          
```

hbldh:
Hello, no I think the version boundary is correct. See the `WriteValue` in the docs of the `5.50`-tagged [commit](https://git.kernel.org/pub/scm/bluetooth/bluez.git/tree/doc/gatt-api.txt?h=5.50) and contrast that to the [current master branch docs](https://git.kernel.org/pub/scm/bluetooth/bluez.git/tree/doc/gatt-api.txt).

PR #138, opened 2019-12-20
bluez: add comment about version check

dlech:
This version check gets questioned somewhat frequently. Each time,
it is found to be correct. So, let's add a comment with a detailed
explanation about what is going on.


dlech:
A also updated it to add a more useful error message in the exception.

dlech:
#137 #129 #106

hbldh:
Very good; thank you very much!

Issue #125, opened 2019-11-06
discover and connect never complete from non-MainThread


derekgorcz:
* bleak version:
0.5.0
* Python version:
3.6.8
* Operating System:
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.3 LTS
Release:	18.04
Codename:	bionic

* BlueZ version (`bluetoothctl -v`) in case of Linux: 
bluetoothctl: 5.50

### Description
discover and connect never complete from non-mainthread

It appears that this may be failing to connect to the system dbus: https://github.com/hbldh/bleak/blob/master/bleak/backends/bluezdbus/discovery.py#L128

### What I Did

```
import asyncio
import threading
from bleak import discover

async def run(loop):
    print('1'*80)
    devices = await discover(loop=loop)  # discover never completes
    print('2'*80)
    for d in devices:
        print(d)

def threaded_func():
    # loop = asyncio.get_event_loop()  # this will fail with "RuntimeError: There is no current event loop in thread 'Thread-1'."
    loop = asyncio.new_event_loop()
    asyncio.set_event_loop(loop)
    loop.run_until_complete(run(loop))

thread = threading.Thread(target=threaded_func)
thread.start()
thread.join()
```

hbldh:
Similar to #93. You should not create a new event loop using the BlueZ. Use the `asyncio.get_event_loop` and skip the `threading` module. Are you sure you really need it? If you are using the `asyncio` module, then you are better off using it for all you distributed tasks.

Use something like this instead:

```python
import asyncio
from bleak import discover

async def run(loop):
    print('1'*80)
    devices = await discover(loop=loop) 
    print('2'*80)
    for d in devices:
        print(d)

loop = asyncio.get_event_loop()
loop.run_until_complete(run(loop))
```

derekgorcz:
I am not able to execute from the main thread. As a workaround I am currently having my thread spin up a new process.

Issue #93, opened 2019-07-17
Discover not working with new event loops.


Mithrandir2k18:
* bleak version: 0.4.3
* Python version: 3.5.3
* Operating System: Debian/Raspian-Stretch

### Description
I want to write a small module that deals with a bluetoothle connection.
Everything works fine until I try to use a new event loop together with bleak.discover().
Please see MWE below for details.

### What I Did

```
import asyncio
from bleak import discover
import logging

async def printer2(loop: asyncio.AbstractEventLoop):
    print("Start Printer2")
    devices = await discover(loop=loop)  # This function gets stuck here.
    for d in devices:
        print(d.name)
    print("Hey")
    return 42

async def printer(loop: asyncio.AbstractEventLoop):  # This function works as expected
    print("Start Printer")
    await asyncio.sleep(5, loop=loop)
    print("Hey")
    return 42


def create():
    loop = asyncio.new_event_loop() #  replacing this with asyncio.get_event_loop() works..
    asyncio.set_event_loop(loop)
    return loop

logging.basicConfig(level=logging.DEBUG)
loop = create()
retval = loop.run_until_complete(printer(loop))
print(retval)

retval = loop.run_until_complete(printer2(loop))
print(retval)

```

hbldh:
I will have to look into this, but I think it is due to the `twisted` implementation used in `txdbus` package that bleak depends on. Check the [`__init__.py`](https://github.com/hbldh/bleak/blob/master/bleak/backends/bluezdbus/__init__.py) to see how I create a loop there.
Is it the same you use a new loop for the client class?

djwmarcx:
I can confirm this situation in my side. No additional info yet.

Mithrandir2k18:
Sorry, didn't have the time to test it right away. Here's a test with my polar h10, but I don't see how a different BLE server would yield different results. Anyway, if the line where create is called is comented out it works as expected, if we create and use a new loop, it won't work. It just waits forever.

```
import asyncio
import bleak
import logging

async def printer2(loop: asyncio.AbstractEventLoop):
    print("Start Printer2")
    devices = await bleak.discover(loop=loop)  # This function gets stuck here.
    for d in devices:
        if "Polar" in d.name:
            return d
    print("Hey")
    return 42


def create():
    loop = asyncio.new_event_loop() #  replacing this with asyncio.get_event_loop() works..
    asyncio.set_event_loop(loop)
    return loop

logging.basicConfig(level=logging.DEBUG)

[...]

PR #143, opened 2020-01-06
backends/bluezdbus: use an a new reactor for each client

mickael9:
This makes things much cleaner :
- The default asyncio event loop is not modified
- No default twisted reactor is installed (a new one will be created every time)
- The reactor will use the provided event loop

Related to: #93 #125

### DIFF
--- a/bleak/backends/bluezdbus/__init__.py
+++ b/bleak/backends/bluezdbus/__init__.py
@@ -5,11 +5,3 @@ __init__.py
 Created on 2017-11-19 by hbldh <henrik.blidh@nedomkull.com>
 
 """
-
-import asyncio
-from twisted.internet import asyncioreactor as _asyncioreactor
-
-loop = asyncio.new_event_loop()
-asyncio.set_event_loop(loop)
-_asyncioreactor.install(eventloop=loop)
-from twisted.internet import reactor  # noqa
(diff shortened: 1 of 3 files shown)


hbldh:
This. This is how it should have been from the start. Thank you very much, this I will merge immediately!

mickael9:
Thanks. I have not much experience with asyncio and multiple event loops. I'm still wondering exactly how useful the "loop" argument is, since we're already in an async function, there must already be a running event loop, and I don't think you can await on tasks from another event loop

hbldh:
If no loop is specified, the default loop will be used. There are multiple people you have asked about sending in non-default loops, This enables running multiple BleakClients in the same process, which is highly desireable.

JoeHut:
actually this is problematic... It looks like the reactor leaks file descriptors (2-3 for each connected client). So if you are connecting and disconnecting devices continuously you will run out of file descriptors and crash. I am not really familiar with twisted, but as I understand it, there are wakers installed represented by file descriptors and neither closed by calling ``stop()``, ``disconnectAll()`` or ``removeAll()`` nor affected by the garbage collector... So I guess the only way is to try and instantiate as few reactors as possible and pass them. If there is someone who can explain me more, I would be happy to help

JoeHut:
The only reference I found is this really old ticket: https://twistedmatrix.com/trac/ticket/3063
But I don't quite understand whether it was resolved or what the general development state of twisted is

Issue #142, opened 2020-01-04
CoreBluetooth AdvertisementData should be handled in callback!?


5frank:
* bleak: 0.5.1
* python: 3.7.5
* os: macOS 10.14.6 (Darwin-18.7.0-x86_64-i386-64bit)

### Description
printing AdvertisementData keys in the `centralManager_didDiscoverPeripheral_advertisementData_RSSI_` callback: 
```
DEBUG:bleak.backends.corebluetooth.CentralManagerDelegate:184: advDataKey: kCBAdvDataIsConnectable
DEBUG:bleak.backends.corebluetooth.CentralManagerDelegate:184: advDataKey: kCBAdvDataLocalName
DEBUG:bleak.backends.corebluetooth.CentralManagerDelegate:184: advDataKey: kCBAdvDataChannel
DEBUG:bleak.backends.corebluetooth.CentralManagerDelegate:184: advDataKey: kCBAdvDataIsConnectable
DEBUG:bleak.backends.corebluetooth.CentralManagerDelegate:184: advDataKey: kCBAdvDataServiceUUIDs
DEBUG:bleak.backends.corebluetooth.CentralManagerDelegate:184: advDataKey: kCBAdvDataChannel
```
but later in discovery, printing the same keys for same device shows some of them disappeared:
```
DEBUG:bleak.backends.corebluetooth.discovery:126: advDataKey: kCBAdvDataIsConnectable
DEBUG:bleak.backends.corebluetooth.discovery:126: advDataKey: kCBAdvDataLocalName
DEBUG:bleak.backends.corebluetooth.discovery:126: advDataKey: kCBAdvDataChannel
```
i.e. AdvertisementData, RSSI etc should be handled in the callback and copied.

hbldh:
I do not own a mac myself, so the fixes for macOS issues happen not so very often. It would be best for you to implement this yourself and send a PR if it is important. 

Would it suffice to convert the `advertisementData: NSDictionary` input to a Python dict and store that in `self.advertisement_data_list` here: [`CentralManagerDelegate.py#L173`](https://github.com/hbldh/bleak/blob/master/bleak/backends/corebluetooth/CentralManagerDelegate.py#L173)?

5frank:
working on a PR. Found some other issues while debugging this. For example that devices are never removed only added.

5frank:
Half related changes to the `BLEDevice` class that I would appreciate some help with (if accepted):

[...]

PR #144, opened 2020-01-07
Advertisement data

5frank:
Fixes the following issues on MacOS/CoreBluetooth backend:

- #142. advertisementData such as service UUIDs lost as multiple calls to
  centralManager_didDiscoverPeripheral_advertisementData_RSSI_
  for same device (but different advertisementData) was not handled.

- Devices never removed only added while scanning and long running
  applications would have get incorrect scan result.

- Scan was never stoped if timeout was 0 or None.

### DIFF
--- a/bleak/backends/corebluetooth/CentralManagerDelegate.py
+++ b/bleak/backends/corebluetooth/CentralManagerDelegate.py
@@ -10,6 +10,7 @@ import asyncio
 import logging
 from enum import Enum
 from typing import List
+from copy import deepcopy
 
 import objc
 from Foundation import (
@@ -24,8 +25,9 @@ from Foundation import (
 )
 
 from bleak.backends.corebluetooth.PeripheralDelegate import PeripheralDelegate
+from bleak.backends.device import BLEDevice
 
-# logging.basicConfig(level=logging.DEBUG)
+
 logger = logging.getLogger(__name__)
 
 CBCentralManagerDelegate = objc.protocolNamed("CBCentralManagerDelegate")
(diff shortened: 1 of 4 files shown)


hbldh:
I am reluctant to add backend specific implementations on BLEDevices right now. This PR will have to wait for a later release, there are too much to test for the 0.6.0 release as it is...

5frank:
Currently the `rssi` and `metadata['uuids']` attributes of  the `BLEDevice` class are broken on corebluetooth (they are always nul/None/empty for me but might behave differently for others). 

As I see it this PR  do not introduce any backend specific changes to `BLEDevice` as all attributes are the same but not nul/None/empty. 
Suggested changes are welcome!

hbldh:
That they are broken is indeed a very unsatisfactory result... I would like to integrate the `BLEDeviceCoreBluetooth` into the existing `bleak.backends.device.BLEDevice` class if possible.

I am not, in general, adversly positioned into including a Device interface and having separate implementations for each backend, but I would like an `abc.ABC` inheritance for the `BLEDevice` and making a separate implementation of functionality in each backend. This is a halfway point that solves it for one backend and leaves the proper structure for later.

If you break out the `BLEDeviceCoreBluetooth` to a separate file called `device.py` in the `corebluetooth` folder and make sure it works well in the macOS world, then I will merge this.

5frank:
note: the PR branch do not work for me without https://github.com/hbldh/bleak/commit/f2aa8d939ec8c06418fb2dc95275bbc409319098 did not include as I assumed it would make any merge easier.

Issue #114, opened 2019-09-20
write with/without response is not valid 


fanshuaiqi:
* bleak version:0.5.0
* Python version:3.7
* Operating System:mac os 10.14.6

### Description
base on my understanding, write_gatt_char can set write with/without response, but the arg of this function is not passed to next level call.It uses self.peripheral.writeValue_forCharacteristic_type_(value, characteristic, 0) to write char eventually.

hbldh:
Is the lack of propagation a problem for you or is it a general comment on the proper way of doing this? I agree that it is a lack of coherence, but I have never experienced myself that the lack of using a write instead of a write-with-response or vice versa is that much of an obstacle?

5frank:
The fix raises the following error for me. 
```
objc.BadPrototypeError: Objective-C expects 3 arguments, Python argument has 4 arguments for <unbound selector writeCharacteristic:value: of PeripheralDelegate at 0x1036ccce0>
```

dlech:
Made some (untested) suggestions on possibly how to fix: https://github.com/hbldh/bleak/commit/75f2ff9f3a1ff7775a062072138b33d588d418aa

5frank:
the underscores seems to be of importance for pyobjc. should match number of args. (I'm new to this).
The following seems to fix it but I'm on a messy branch.
```diff
diff --git a/bleak/backends/corebluetooth/PeripheralDelegate.py b/bleak/backends/corebluetooth/PeripheralDelegate.py
index 9fa73b3..0e2d0c5 100644
--- a/bleak/backends/corebluetooth/PeripheralDelegate.py
+++ b/bleak/backends/corebluetooth/PeripheralDelegate.py
@@ -145,10 +145,10 @@ class PeripheralDelegate(NSObject):
 
         return descriptor.value()
 
-    async def writeCharacteristic_value_(
+
+    async def writeCharacteristic_value_type_(
         self, characteristic: CBCharacteristic, value: NSData, response: int
     ) -> bool:
-
         cUUID = characteristic.UUID().UUIDString()
         self._characteristic_write_log[cUUID] = False
 
diff --git a/bleak/backends/corebluetooth/client.py b/bleak/backends/corebluetooth/client.py
index 8ac39ba..63e8b37 100644
--- a/bleak/backends/corebluetooth/client.py
+++ b/bleak/backends/corebluetooth/client.py
@@ -220,7 +220,7 @@ class BleakClientCoreBluetooth(BaseBleakClient):
[...]

PR #146, opened 2020-01-14
macOS: Adapt characteristic write log to account for WriteWithoutResponse.

synicalsyntax:
f2aa8d9 fails to account for an edge case the issues found in #114, which results in the while loop in `writeCharacteristic_value_type` never resolving because the peripheral never calls the `didWriteValueForCharacteristic` method.

> On the other hand, if you specify the write type as CBCharacteristicWriteWithoutResponse, Core Bluetooth attempts to write the value but doesn’t guarantee success. If the write doesn’t succeed in this case, you aren’t notified and you don’t receive an error indicating the cause of the failure.

[Apple docs for writeCharacteristic_value_type](https://developer.apple.com/documentation/corebluetooth/cbperipheral/1518747-writevalue?language=objc#discussion)

> **Core Bluetooth invokes this method only when your app calls the writeValue:forCharacteristic:type: method with the CBCharacteristicWriteWithResponse constant specified as the write type.** If successful, the error parameter is nil. If unsuccessful, the error parameter returns the cause of the failure.

[Apple docs for didWriteValueForCharacteristic](https://developer.apple.com/documentation/corebluetooth/cbperipheraldelegate/1518823-peripheral?language=objc#discussion)

### DIFF
--- a/bleak/backends/corebluetooth/PeripheralDelegate.py
+++ b/bleak/backends/corebluetooth/PeripheralDelegate.py
@@ -150,7 +150,7 @@ class PeripheralDelegate(NSObject):
     ) -> bool:
 
         cUUID = characteristic.UUID().UUIDString()
-        self._characteristic_write_log[cUUID] = False
+        self._characteristic_write_log[cUUID] = False if not response else True
 
         self.peripheral.writeValue_forCharacteristic_type_(value, characteristic, response)
 


hbldh:
Thank you, this was a very nice catch indeed! I was not aware of this at all. Will merge this immediately.

PR #148, opened 2020-01-14
backends/dotnet: run callbacks from main thread

mickael9:
Since we're an async library, we probably don't want to run the
callbacks from a different thread than the one the event loop is running
from.


hbldh:
Hmm, yes, I agree. I have only had very small callbacks in my code and apparently never thought about them being cumbersome to run in the thread that received the dotnet notification. Will merge this and test it. Thank you!

PR #152, opened 2020-01-27
bluez: only subscribe to BlueZ messages on DBus

### LINKED ISSUE #151: Callback execution for unrelated DBus messages
* bleak version: master
* Python version: 3.8.1
* Operating System: Arch Linux
* BlueZ version (`bluetoothctl -v`) in case of Linux: 5.52

During debugging, I found that the DBus matches are not only registered for relevant Bluez messages, but for all senders that use the same interface (like the org.freedesktop.DBus.Properties). So if you have logging enabled and NetworkManager running, you will see that messages WiFi Access Point RSSI updates are logged. Those messages are subsequently discarded, but shouldn't even be received in the first place.

PR will be opened soon.

JoeHut:
The management of the discovery and clients on Linux is primarly done
through subscribing and listening to DBus messages. The subscriptions
have been identified only through the Interface names and their members.
However, those interfaces are generic and can be implemented by all kind
of applications that expose their API through DBus. This led to the
situation where for example NetworkManager events are forwarded to the
bleak application and handled in the callback.

Although those messages have been ignored, they shouldn't even reach the
application. This patch limits the subscriptions to messages with the
"/org/bluez" path prefix and therefore to relevant BlueZ messages.

Fixes #151

### DIFF
--- a/bleak/backends/bluezdbus/discovery.py
+++ b/bleak/backends/bluezdbus/discovery.py
@@ -145,6 +145,7 @@ async def discover(timeout=5.0, loop=None, **kwargs):
             parse_msg,
             interface="org.freedesktop.DBus.ObjectManager",
             member="InterfacesAdded",
+            path_namespace="/org/bluez",
         ).asFuture(loop)
     )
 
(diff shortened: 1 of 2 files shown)


hbldh:
You are very correct. These were definitely missing earlier.

PR #154, opened 2020-02-04
bluez: Cleanup all resources even if the client disconnects

### LINKED ISSUE #130: BlueZ bus interface is not cached and can not be cleaned up properly
* bleak version: 0.5.1
* Python version: 3.5
* Operating System: osmc
* BlueZ version (`bluetoothctl -v`) in case of Linux: 5.43

### Description

Repeated calls to `client.connect()` (both on the same instance, as well as newly created instances) will create new bus-objects (see [client.py:96](https://github.com/hbldh/bleak/blob/11e7adf1ae1774b22667fcc4e6533ad69597951a/bleak/backends/bluezdbus/client.py#L96)) for each call, without the means to clean them up again.

### What I Did
Auto-Reconnect loop like this:

```python
while True:
    client = bleak.BleakClient(addr, loop=loop)
    try:
        await client.connect()
    except bleak.exc.BleakError as exc:
        logging.error('failed to connect: %s', exc)
        continue
    [... logic here ....]
```

### What I expected
Either `client.__del__` or an explicit `client.cleanup` to call `client._bus.disconnect`, or `client._bus` to be cached.

### What happened
```
WARNING failed to connect: org.bluez.Error.Failed: Software caused connection abort
[previous message repeats 244 times]
  File "/usr/local/sbin/ble.py", line 156, in reconnect
    await client.connect()
  File "/opt/bleak-venv/lib/python3.5/site-packages/bleak/backends/bluezdbus/client.py", line 93, in connect
    await discover(timeout=timeout, device=self.device, loop=self.loop)
  File "/opt/bleak-venv/lib/python3.5/site-packages/bleak/backends/bluezdbus/discovery.py", line 128, in discover
    bus = await client.connect(reactor, "system").asFuture(loop)
  File "/usr/lib/python3.5/asyncio/futures.py", line 380, in __iter__
    yield self  # This tells Task to wait for completion.
  File "/usr/lib/python3.5/asyncio/tasks.py", line 304, in _wakeup
    future.result()
  File "/usr/lib/python3.5/asyncio/futures.py", line 293, in result
    raise self._exception
txdbus.error.RemoteError: org.freedesktop.DBus.Error.LimitsExceeded: The maximum number of active connections for UID 0 has been reached
```

JoeHut:
Recently there have been changes to make sure we clean up all DBus and
Twisted resources when we disconnect from a server to prevent memory
leaks and running out of available DBus connections. However, this was
only implemented for the case where the application actively disconnects
form the server. If the disconnection was initiated by the server for
example or by going out of range, the dbus connections still leaked.

This patch splits the cleanup into two different methods. One for
cleaning the subscriptions to DBus messages (for example through
notifications) and one for cleaning up the DBus and Twisted resources.
The split is necessary because we want to first remove all subscriptions
before initiating a disconnection and eventually clean up the resources
in the active case, but in the inactive disconnection case we want to
cleanup all resources immediately.

It also fixes issues seen in the recent changes to have a twisted reactor for
each client connection. This actually leaks resources like file descriptors
that can not be cleaned up properly. Therefore a helper is provided to cache
the used reactor.

Fixes #130

### DIFF
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -164,7 +164,11 @@ class BleakClientBlueZDBus(BaseBleakClient):
         )
         return True
 
-    async def _cleanup(self) -> None:
+    async def _cleanup_notifications(self) -> None:
+        """
+        Remove all pending notifications of the client. This method is used to
+        free the DBus matches that have been established.
+        """
         for rule_name, rule_id in self._rules.items():
             logger.debug("Removing rule {0}, ID: {1}".format(rule_name, rule_id))
             try:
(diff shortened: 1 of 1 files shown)


hbldh:
Merged this, with additions into develop. I changed things aroudna bit, so the `get_reactor` could be accessed from `client`, `discovery` and `scanner` instead of being used just in `client`.

Issue #184, opened 2020-05-12
Feature request: Add disconnection callback in CoreBluetooth


hardik-prajapati:
* bleak version: 0.6.1
* Python version: 3.7.4
* Operating System: Mac OS Mojave
* BlueZ version (`bluetoothctl -v`) in case of Linux: 

### Description
Disconnection callback not implemented in CoreBluetooth

### What I Did
```
client.set_disconnected_callback(disconnection_callback)
```

From the code it seems we can call callback function from the function  `centralManager_didDisconnectPeripheral_error_` from `CentralManagerDelegate.py`.

I can do that raise PR for it if you want.

hbldh:
No, it is not implemented. Feel free to do it if you want it.

bsiever:
For what it's worth,
-  I've implemented this (disconnect callbacks of macOS)
- fixed `is_connected` on macOS
- added support for multiple clients on macOS
- partially implemented scanning on macOS (need to test and add support for filters)

I'm going to do a bit more testing and try to issue a pull request in the coming weeks.  In the meantime, here's my fork/branch: https://github.com/bsiever/bleak/tree/macOS-multiple-clients

(And it is a great library --- thanks!)

hardik-prajapati:
Created pull request here : https://github.com/hbldh/bleak/pull/186

PR #186, opened 2020-05-13
Added disconnected callback feature for CoreBluetooth

hardik-prajapati:
Added disconnected callback variable in `CentralManagerDelegate` class, which being set by the `set_disconencted_callback` function.

CentralManagerDelegate calls the callback function without argument, so I added wrapper function `disconnected_callback_client` in `client` that pads the `client` as argument and calls actual disconnection callback. 

Issue #184


hbldh:
Checked this and it seems to work fine. Thank you!

Issue #207, opened 2020-05-28
Unit Testing w/ Devices


bsiever:
Hi, 

As I've been experimenting with (and enhancing) bleak I found testing to be bit cumbersome.  I thought I'd: 
- Ask other developers how you're testing your work --- Any info to share?
- Explain how I'm doing my testing and enhance/contribute my work if there's sufficient interest.

As for me: I'm mainly focused on the macOS backend and haven't been testing my work on other platforms.   I think my current approach should be portable with a little effort. Here's what I'm doing/developing:

1. I'm using BBC Micro:Bits
   - They're pretty cheap test devices ([Amazon Link](amazon.com/micro-bit-BBC2546862-Micro-go/dp/B01G8X7VM2) : ~$18 US each).  
   - They're easy to program with custom BLE services and advertising.  I've been using PlatformIO's micro:bit support (with some heavy modifications). The resulting firmware is a single file. They enumerate as a USB drive and firmware can be changed by simply uploading a new firmware file.
2. I've created a custom service with a lot of BLE features that I wanted to test.  (Here: https://github.com/bsiever/microbit-ble-testservice)
   - I'll be creating a few variations with other things to test (different advertising, services with "duplicate" UUIDs, etc).  My test fixtures upload the firmware needed for the set of tests.
3. I've working on a variety of unit tests using pytest and alt-pytest-asyncio.  
   - Current tests and testing approaches are still a bit rough, but it's a promising start. Initial tests are here: https://github.com/bsiever/bleak/tree/scan-filter/tests/microbit1 . 
   - As of this moment I've got about 18 candidate tests and plan to do about 12+ more.  So far I've been focused on reads/writes and connects/disconnects.  
    - Next steps include:
       - Notifies and indicates 
       - Authorized Reads & Writes (success and failure)
       - Descriptors (reading and writing)
       - Multiple simultaneous clients
       - Python behavioral stuff & context management
        - Scanning & Filtering
 
Slight tangent:  As a result of the testing I've made a lot of changes to the macOS backend.  Unfortunately I've been building off the release version instead of keeping up with the devel branch.  I'll try to do a PR at some point...but it may be messy and merit rejection.

hbldh:
[...]

PR #209, opened 2020-05-30
Updates to CoreBluetooth backend

bsiever:
This includes several updates to the CoreBluetooth backend. These include:

- Support for scanning w/ filtering (Using BlueZ-style flags)
- Performance improvements (connect will use cached peripheral for faster connect following discovery and/or faster reconnecting)
- Several bug fixes
   - Some errors were causing crashes rather than raising exceptions.  (Exceptions have been moved out of delegate callbacks)
- Initial work for supporting multiple clients (still needs further testing)
  - Different approach to disconnect callbacks for multiple peripherals
- Misc. small consistency changed (Ex: importing/using CoreBluetooth constants rather than their values)
- A few of these are a bit hacky and merit revision

There are a lot of changes.  I'm happy to answer any questions and won't be hurt if this is rejected.  I'm also able to do a PR for my test suite (pretty rough) to help quickly vet the work (requires 1+ microbit).

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -17,6 +17,8 @@ jobs:
                 exclude:
                     -   os: windows-latest
                         python-version: 3.8
+                    -   os: macos-latest
+                        python-version: 3.5
         steps:
             -   uses: actions/checkout@v2
             -   name: Set up Python ${{ matrix.python-version }}
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -8,7 +8,6 @@ on:
 
 jobs:
     build:
-
         runs-on: ubuntu-latest
         strategy:
             matrix:
@@ -22,6 +21,7 @@ jobs:
                     python-version: ${{ matrix.python-version }}
             -   name: Install dependencies
                 run: |
+                    apt install -y bluez
                     python -m pip install --upgrade pip
                     if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
                     if [ -f requirements_dev.txt ]; then pip install -r requirements_dev.txt; fi
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -9,6 +9,7 @@ on:
 jobs:
     build_linux:
         name: "Build and test"
+        runs-on: ubuntu-latest
         strategy:
             matrix:
                 os: [ubuntu-latest]
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -48,7 +48,9 @@ jobs:
         strategy:
             matrix:
                 python-version: [3.5, 3.6, 3.7, 3.8]
-
+                exclude:
+                    -   os: windows-latest
+                        python-version: 3.8
         steps:
             -   uses: actions/checkout@v2
             -   name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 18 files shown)


5frank:
Nice! I have not had the time to test it yet.
Some thoughts:

Not sure I like adding variables to class instances. e.g 
```python
        self[xUUID]._error = None  # Clear error prior to new event
```
I believe it is a bit confusing and it makes pylint unhappy. Perhaps inherit `ayncio.Event` instead? This would also make it easy to add a timeout to `event.wait()` if/when needed and remove some duplicated code. Example (untested):
```python
class _BleakCorebluetoothEvent(asyncio.Event):
    def __init__(self):
        self._error = None
        super().__init__()

    def clear(self):
        self._error = None
        super().clear()

    def set(error: NSError = None): # or pass a Exception?
        self._error = error
        super().set()

    async def wait(self, timeout=None):
        if timeout:
            try:
                await asyncio.wait_for(super().wait(), timeout)
            except asyncio.TimeoutError:
                raise TimeoutError("")
        else:
            await super().wait()

        if self._error:
            raise SomeException(self._error)
```

bsiever:
My approach is a hack.  I like the idea of a subclass (and better semantics for waits/timeouts).   I'll probably try to incorporate your approach in the next few days.

hbldh:
Whoa. This was a lot of changes in one PR... I would have liked them spread out over several to be able to review them better. Nevertheless, thank you!

I will await you changes and then plan to look into this as soon as possible though.

bsiever:
Yeah, sorry for the number of changes in one batch.  It was one of my motivations for Issue #207 (Unit testing with devices).  My development path was pretty haphazard. I'll try to be a little more structured in the future.  

I've incorporated a first pass at @5frank's approach, which is much much cleaner.  It also provides an opportunity for timeouts on all device events, which could make "hanging" almost impossible.   I think this could use further review discussion, but the constructors in `initWithPeripheral_` just need to specify the timeout for events.  I've chosen to include the new  `_BleakCorebluetoothEvent` class in PeripheralDelegate.py.  It may be better as a separate file so it could be used elsewhere.

f0rodo:
+1 This fixed bleak breaking in Catalina

dlech:
Hi @bsiever, can we break this down into some smaller pull requests to start getting things merged? I started fixing things myself before I saw this PR. Changes are [here](https://github.com/hbldh/bleak/compare/master...pybricks:master) if you want to compare notes (still a bit of a work in progress though).

hbldh:
I have planned on devoting an entire day to bleak during next week, with explicit focus on macOS issues. @dlech Could you integrate your changes on with these ones (if needed)? I could merge this into a separate branch in my repo so that additional changes can be appended. Would that help?

@bsiever Thank you for the work done here. I will try it out first thing on the Bleak-day!

bsiever:
@hbldh I hope Bleak Day goes well!    If you have a micro:bit the unit tests I've made can help quickly vet that additions aren't breaking too many things.

@dlech It's clear that you caught things I missed and I restructured parts and added features you haven't hit.  We both made a lot of changes and it's definitely a bit messy to integrate.  The "big ones" for me were: 

- Moved error handling out of delegates.  That was causing un-catchable Python crashes.
- Added support for disconnect via weak references from the `cbapp` singleton.  
- Added support for multiple clients (also using the `cbapp`'s weak references for some updates)
- Used cached `CBPeripheral` objects for faster connections
- Lots of small, misc. error handling. 

I probably won't have time to make many updates for the next week or two, but I should have time to reply to any questions.

hbldh:
I merged the PR #227 before this one, since it was smaller and contained exacty what I needed for the moment. Regrettably I have to had time to merge this with that yet, so I will leave this one for a 0.7.X release in a few days.

hbldh:
@bsiever I tried to do a careful merge of your changes into a separate branch (`bsiever-develop`), but it ended up in a broken state. I do not now how to move forward from there and I think I have to abandon the attempt. Reluctantly, I will close this PR and ask you to update your changes with current `develop`. Thank you for the effort.

The things I liked in the PR, that wasn't solved in merging of #227:

1) **the scanning filter work**: If you could integrate your scanning filter work with the current develop and make a PR I would be very grateful. If there are more changes still needed for making your test suite run, please add them in a separate PR and we can address that separately.
2) **the reuse of scan results**: For the reuse of scan results I have an idea for version 0.8.0 that will collide with that, so for the time being I will go a different route.

Issue #116, opened 2019-10-03
Hanging on "async with BleakClient..."


GreatScott:
* bleak version: 0.5.0-develop
* Operating System: macOS 10.14.6 (on early 2015 Macbook Air)

### Description
In running the examples on MacOS, everything hangs after/within `async with BleakClient(address, loop=loop) as client:`  (_within_ `run`) i.e. the subsequent typical example `x = await client.is_connected` or any other calls do not appear to run. Despite this, the OS does indicate that the BLE device _is_ connected indefinitely until the script is stopped. Everything appears to be working fine on Windows.

hbldh:
I will try to replicate this and see if I can verify it, but my trials with a Mac Mini with an even older macOS version did not exibit this behaviour. Did this fail with all examples? Or just some? Which in that case?

GreatScott:
Thanks for the response! I'll run through all the examples in a bit and report back.

alamysm:
I'm having the same issue, I think. I'm at the top of the develop branch ("0.5.2a1 (2019-11-10)"?) on Mac OS 10.14.6 on a 2019 MBP 13. I've tested with Python 3.7.6 and 3.8.1 with the same results. The discover example appears to work as expected, but get_services and service_explorer exhibit the described behavior. Here are my logs for service_explorer:

`DEBUG:bleak.backends.corebluetooth.CentralManagerDelegate:Discovered device 3DBBDC7D-4F6E-401F-8154-7B8ACE865CAE: Unknown @ RSSI: -97
DEBUG:bleak.backends.corebluetooth.CentralManagerDelegate:Discovered device F44CAE74-564E-4F68-9FA4-8BA1D78EFFFB: Unknown @ RSSI: -84
DEBUG:bleak.backends.corebluetooth.CentralManagerDelegate:Discovered device 7F242CAE-B671-4530-8F2D-6F5D0B4874E6: Unknown @ RSSI: -83
DEBUG:bleak.backends.corebluetooth.client:Connecting to BLE device @ F44CAE74-564E-4F68-9FA4-8BA1D78EFFFB
DEBUG:bleak.backends.corebluetooth.CentralManagerDelegate:Successfully connected to device uuid F44CAE74-564E-4F68-9FA4-8BA1D78EFFFB
WARNING:bleak.backends.corebluetooth.PeripheralDelegate:PeripheralDelegate is not compliant
DEBUG:bleak.backends.corebluetooth.client:Retrieving services...
[...]

PR #212, opened 2020-06-07
Fix Hanging on "async with BleakClient..." on OSX

ShaharHD:
Fix for issue #116


ShaharHD:
Created fix on master by mistake

PR #213, opened 2020-06-07
Fix Hanging on "async with BleakClient..." on OSX

ShaharHD:
Fix for issue #116

### DIFF
--- a/bleak/backends/corebluetooth/CentralManagerDelegate.py
+++ b/bleak/backends/corebluetooth/CentralManagerDelegate.py
@@ -203,9 +203,10 @@ class CentralManagerDelegate(NSObject):
                 peripheral.identifier().UUIDString()
             )
         )
-        peripheralDelegate = PeripheralDelegate.alloc().initWithPeripheral_(peripheral)
-        self.connected_peripheral_delegate = peripheralDelegate
-        self._connection_state = CMDConnectionState.CONNECTED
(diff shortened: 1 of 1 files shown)


hbldh:
I just saw the changes in #209, which solves this issue in another way, which I think will yield an even better result in the end. Thank you for the PR though! I will close this and if #209 does not solve the same issue, we will ahve to revisit this later on.

ShaharHD:
@hbldh Why not incorporate the fix right now (It is a limited fix, specific to OS X)? 

And once #209 PR (which is huge) is completed, it will overwrite this fix?

Right now - every time I install Bleak, I need to have a script to manually patch it.

hbldh:
Because I have no desire to make two releases so close to each other. There will be one next week, so I will wait until then.

It is not unusable on macOS. I used it just fine a minute ago.

ShaharHD:
Look at comments of the other users on #116 which confirms my fix works...

ShaharHD:
> It is not unusable on macOS. I used it just fine a minute ago.

Be aware you were not able to reproduce the issue last year as well.

hbldh:
I have merged #227 into `develop` and will make a version 0.7.0 release later today with those changes as well as a lot of others. If the "fire twice" problem is still left after release of version 0.7.0, I will include these changes in a version 0.7.X release tomorrow.

Issue #189, opened 2020-05-13
Can't read characteristic longer than 20 bytes.  


gyamanishi:
* bleak version:0.6.1
Python version:3.7
Operating System:Windows 10
BlueZ version (bluetoothctl -v) in case of Linux:

### Description

I am trying to read a 200 byte characteristic, but the read receives 20 bytes.  The documentation on reading a blob (large characteristic)  is not clear.

### What I Did
Verified the MTU exchange happened.  I can write a 244 byte characteristic to a TI launchpad.
When I issue a read, it only receives 20 bytes.

I can't find where you specify the length of the read, or the offset.  The microsoft doc mentions an error code that is the characteristic is not long, hinting at the fact that it reads blobs.
```
Paste the command(s) you ran and the output.
Connected: True
[Service] 00001800-0000-1000-8000-00805f9b34fb: Generic Access Profile
	[Characteristic] 00002a00-0000-1000-8000-00805f9b34fb: (write-without-response,write) | Name: , Value: None 
	[Characteristic] 00002a01-0000-1000-8000-00805f9b34fb: (write-without-response,write) | Name: , Value: None 
	[Characteristic] 00002a04-0000-1000-8000-00805f9b34fb: (read) | Name: , Value: b'P\x00\xa0\x00\x00\x00\xe8\x03' 
[Service] 00001801-0000-1000-8000-00805f9b34fb: Generic Attribute Profile
[Service] f000ffd0-0451-4000-b000-000000000000: Unknown
	[Characteristic] 0000ffd6-0000-1000-8000-00805f9b34fb: (read) | Name: Log Retrieval, Value: b'\x1f\x1f\x1f\x1f\x1f\x1f\x1f\x1f\x1f\x1f\x1f\x1f\x1f\x1f\x1f\x1f\x1f\x1f\x1f\x1f' 
		[Descriptor] 00002902-0000-1000-8000-00805f9b34fb: (Handle: 12) | Value: b'\x00\x00' 
		[Descriptor] 00002901-0000-1000-8000-00805f9b34fb: (Handle: 13) | Value: b'Log Retrieval'

dlech:
Can you please fill in all of the version information?

gyamanishi:
Here is another clue.  I can read a 200 byte descriptor no problem.  When I try to read a 200 byte value I get: 
[...]

PR #217, opened 2020-06-15
Fix app crash on Windows when receiving approx 300 bytes

pawelzwronek:
OS: Win10, Python: 3.7, 3.8
For unknown for me reason after receiving approx 300 bytes (in 26 bytes messages) with 'client.start_notify(...)' my app crashes. After some investigating I found that removing line `output = Array.CreateInstance(Byte, reader.UnconsumedBufferLength)` fixes my app, so I replaced it with for loop to read one byte at a time from `args.CharacteristicValue`. I applied this scheme in 2 other places but don't know if it's necessary.

### DIFF
--- a/bleak/backends/dotnet/client.py
+++ b/bleak/backends/dotnet/client.py
@@ -307,8 +307,7 @@ class BleakClientDotNet(BaseBleakClient):
         )
         if read_result.Status == GattCommunicationStatus.Success:
             reader = DataReader.FromBuffer(IBuffer(read_result.Value))
-            output = Array.CreateInstance(Byte, reader.UnconsumedBufferLength)
-            reader.ReadBytes(output)
+            output = [reader.ReadByte() for _ in range(reader.UnconsumedBufferLength)]
             value = bytearray(output)
             logger.debug("Read Characteristic {0} : {1}".format(_uuid, value))
         else:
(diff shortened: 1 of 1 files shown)


hbldh:
This is a very interesting discovery! How is this reasonable?

Anyway, this might solve #189 and potentially other read error in the .NET backend as well. @pawelzwronek If you could please direct the PR towards the `develop` branch I will test it and merge it next week.

pawelzwronek:
I was trying to get any log or stack trace but the app just finish execution. I'm not .NET person but will try to debug pythonnet like described [here](https://github.com/pythonnet/pythonnet/wiki/Various-debugging-scenarios-of-embedded-CPython)

pawelzwronek:
Good news. After upgrade pythonnet to 2.5.1, where 5 days ago was added important fix, my problem dissipated :)

hbldh:
@pawelzwronek Thank you for telling me about `pythonnet`, that makes me very glad to hear! 

Does this mean that your PRs (#217 and #225) are not needed anymore? Was the problem only fixed in 2.5.1 and not in 2.5.0? Out of mere curiosity, I would like to see the change that actually solved this and if you know if it came specifically in 2.5.1 it narrows down my search...

pawelzwronek:
PRs closed. Yes, 2.5.1 fixed the issue. Today when I was testing this issue it only showed on python 3.8, on 3.7 it was ok. When creating this PR I was convinced it was broken on 3.7 and 3.8, but now I'm not sure or it's very nondeterministic. I think the problem was related with 'ob's ref count', see 2 commits in 2.5.1. I can test which one but because the issue is quite nondeterministic it won't tell much.

pawelzwronek:
Actually because of nondeterminism of the issue it could be coincidence it fixed my issue, hard to tell :dizzy_face:

hbldh:
Ok, create a new issue if this resurfaces. I bumped the version in commit a22245b just now. Will advise people to use this version instead.

PR #224, opened 2020-06-22
Remove passing of asyncio event loop as argument.

dlech:
This removes most cases of passing the asyncio event loop as an argument (some cases are kept in the Linux backend for interop with txdbus/twisted).

Passing the loop as an argument has been deprecated in Python 3.8 and will be removed in 3.10.

Explanation: https://stackoverflow.com/a/60315290/1976323

This is a breaking change, but it seems better to do it now rather than wait for it to break in Python itself in a few years.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -78,13 +78,13 @@ Connect to a Bluetooth device and read its model number:
     address = "24:71:89:cc:09:05"
     MODEL_NBR_UUID = "00002a24-0000-1000-8000-00805f9b34fb"
 
-    async def run(address, loop):
-        async with BleakClient(address, loop=loop) as client:
+    async def run(address):
+        async with BleakClient(address) as client:
             model_number = await client.read_gatt_char(MODEL_NBR_UUID)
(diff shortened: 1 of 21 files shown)


hbldh:
Good reasoning. This seems to cover all backends adequately. 

I would however like to wait with this breaking change until 0.8.0, i.e after some macOS issues have been resolved first. It feels like it will be unecessary problematic to separate bugs from that work with error reports due to these changes...
Does that seem resonable to you, or do you think that removing event loop stuff from the macOS backend is actually helpful for   all issues?

dlech:
It seems reasonable to wait.

hbldh:
Did you try this loop-free handling out in the .NET backend? I did no test properly before merging and I am having weird behaviour first getting services (which I remedied) and also when receiving notifications (which I have not been able to solve). I just want to see if it might be due to Windows 10 update problems reported by people or if it is a true Bleak bug.

dlech:
No, I don't think I tried this on Windows yet.

hbldh:
Thank you. Then I will continue to remedy the situation.

PR #244, opened 2020-07-02
add CHANGELOG.rst to MANIFEST.in to be able to build sdist

toppk:
running: 
   python setup.py sdist
   pip install -t foo dist/bleak-0.7.0.tar.gz

will fail because CHANGELOG.rst is required.  This also breaks source installs from pypi, for example:
   pip install bleak --no-binary :all:

included is a trivial fixup

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,5 @@
 include AUTHORS.rst
+include CHANGELOG.rst
 include CONTRIBUTING.rst
 include HISTORY.rst
 include LICENSE


hbldh:
Ach, such a clumsy mistake by me. Thank you. Will release immediately toghether with some other changes.

Why, may I ask, do you build yourself instead of pulling from PyPI?

toppk:
> Why, may I ask, do you build yourself instead of pulling from PyPI?

I do pull from pypi, but I use the source using:   pip install -r requirements.txt --no-binary :all:

I like it because it's closer to the source.  it's not in the bones of python to use source packages, but I think that's a shame.  regardless, it's taught me a bit about python's sdlc tooling.

PR #253, opened 2020-07-13
remove objc delegate compliance checks

dlech:
It seems that these expect all delegate members to be implemented, however in reality only the `CentralManagerDelegate` `didUpdateState` callback is required. All other methods are optional, so this warning is not particularly useful.


hbldh:
I trust you on this. Will merge

PR #260, opened 2020-07-30
fence post error

danaronson:
fixed code to use WriteValue to be >= bluez 5.50 (was >)


dlech:
Actually, I think the error is in the comment, not the code. It looks like the type option was implemented in BlueZ 5.51.

https://github.com/bluez/bluez/commit/aa9f9b193c51d371b5d2a863f35a034ee0377b22

hbldh:
I agree with @dlech : according to documentation it was added in 5.51.

See

https://kernel.googlesource.com/pub/scm/bluetooth/bluez/+/refs/tags/5.51/doc/gatt-api.txt

and compare to

https://kernel.googlesource.com/pub/scm/bluetooth/bluez/+/refs/tags/5.50/doc/gatt-api.txt

`type` is present in 5.51 and not in 5.50. I am sorry, but this PR will be closed.

PR #270, opened 2020-08-14
Fix get_services function of corebluetooth client

charlieda:
The get_services function incorrectly returns the underlying `_service` array rather than the services object when reading it from the cache.


hbldh:
You are very right. If you add

```python
 self._services = None
```

to the `disconnect` method and direct the PR towards the `develop` branch and I am prepared to merge it directly.

charlieda:
Thanks for getting back so quickly. I've made those changes now :)

hbldh:
Thank you for your contribution!

PR #271, opened 2020-08-17
Implementation of `find_specific_device` feature

hbldh:
A new method in the `BleakScanner` interface. Using the scanner soution, one can remove the hard wait time in `connect` calls with a much faster, variable time scanning which exits when the sought device has been detected by the scanner. This will result in generally faster connections and hopefully more stable connects as well.

### DIFF
--- a/bleak/backends/corebluetooth/CentralManagerDelegate.py
+++ b/bleak/backends/corebluetooth/CentralManagerDelegate.py
@@ -96,11 +96,8 @@ class CentralManagerDelegate(NSObject):
         """
         await asyncio.wait_for(self.powered_on_event.wait(), timeout)
 
-    async def scanForPeripherals_(self, scan_options) -> List[CBPeripheral]:
-        """
-        Scan for peripheral devices
-        scan_options = { service_uuids, timeout }
-        """
+    @objc.python_method
+    def start_scan(self, scan_options):
         # remove old
         self.devices = {}
         service_uuids = []
@@ -110,17 +107,11 @@ class CentralManagerDelegate(NSObject):
                 list(map(string2uuid, service_uuids_str))
             )
 
-        timeout = 0
-        if "timeout" in scan_options:
-            timeout = float(scan_options["timeout"])
-
         self.central_manager.scanForPeripheralsWithServices_options_(
             service_uuids, None
         )
 
-        if timeout > 0:
-            await asyncio.sleep(timeout)
-
+    async def stop_scan(self) -> List[CBPeripheral]:
         self.central_manager.stopScan()
 
         # Wait a while to allow central manager to stop scanning.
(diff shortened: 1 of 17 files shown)


hbldh:
Requires doctrings for the `BleakScanner` interface before merging to develop.

bleak/backends/scanner.py
```diff
@@ -41,3 +44,18 @@ async def set_scanning_filter(self, **kwargs):
     @abc.abstractmethod
     async def get_discovered_devices(self) -> List[BLEDevice]:
         raise NotImplementedError()
+
+    @classmethod
+    @abc.abstractmethod
+    async def find_specific_device(cls, device_identifier: str, timeout: float = 10.0) -> BLEDevice:
```
dlech, line 50: How about calling this `find_device_by_id()`? This way we could have other methods like find device with service UUID or find device with manufacturer specific data?

bleak/backends/scanner.py
```diff
@@ -41,3 +44,18 @@ async def set_scanning_filter(self, **kwargs):
...
+        """A convenience method for obtaining a ``BLEDevice`` object specified by MAC address or (macOS) UUID address.
```
dlech, line 51: If you care about being technically correct, it is a Bluetooth Address, not a MAC address. Both use the same EUI-48 format but they are not the same thing. But most people will still call a a MAC address anyway. :smile:

bleak/backends/scanner.py
```diff
@@ -41,3 +44,18 @@ async def set_scanning_filter(self, **kwargs):
...
+
+        Args:
+            device_identifier (str): The MAC/UUID address of the Bluetooth peripheral sought.
+            timeout (float): Optional timeout to maximally wait for detection of specified peripheral. Defaults to 10.0 seconds.
```
dlech, line 55: "maximally wait` sounds a bit odd to me as a native English speaker. Maybe say "time to scan before giving up" or something like that?

bleak/backends/corebluetooth/CentralManagerDelegate.py
```diff
@@ -135,6 +126,16 @@ async def scanForPeripherals_(self, scan_options) -> List[CBPeripheral]:
 
         return []
 
+    async def scanForPeripherals_(self, scan_options) -> List[CBPeripheral]:
```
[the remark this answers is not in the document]
hbldh: Very good, I will add it to this and `stop_scan` as well.

bleak/backends/scanner.py
hbldh: I do. This is good. Made change.

bleak/backends/scanner.py
hbldh: I thought a very shrt while on naming for this. Picked one and moved forward, meant to change it at the end to something more appropriate.

Yes, you are right. I meant those filtering methods to be unimplemented in Bleak and left to users to implement by themselves, to avoid having to support the code. But the name is still better than `find_specific_device`. 

`find_device_by_address`?

bleak/backends/scanner.py
hbldh: Changed the wording on this.

bleak/backends/scanner.py
dlech: > find_device_by_address

I like this name too.

hbldh:
With regards to moving the implementation to the `BaseScanner`, I agree totally. I made a note of it in f21161a, but I forgot to do. It will not be an abstract method any longer, so I wanted to think it over a while first.

Issue #266, opened 2020-08-09
Bleak: WiteTheDocs session


hbldh:
Create better, more helpful documentation on ReadTheDocs.

Proposed layout of the new documentation:

- Introduction
- What Bleak is and what it isn't
- Basic Usage
    - Scanning
    - Connecting to a dvice
    - Notifications
- Advanced usage
    - Scanning
        - BleakScanner
        - Filtering
        - Advertised data
    - Client
        - Reconnections
        - Pairing
        - Bleak SigSpec (#237)
- Differences between OSes
- Examples
- Changelog
- Contributing
- API (Interfaces)

jackjansen:
As this has the "opinions appreciated" label, here is my view as someone who has only discovered bleak last week, and likes it a lot, but is struggling.

My first problem is that the API documentation is phrased in a way that is geared towards bleak developers, i.e. people implementing/extending bleak. This is probably due to the primary API consisting of the abstract classes to be subclassed and extended from the various backends. As an example, the definition of the `BaseBleakClient`, the first class you come across, is

```
The Client Interface for Bleak Backend implementations to implement.
The documentation of this interface should thus be safe to use as a reference for your implementation.
```

Whereas it would be much friendlier for _users_ of bleak if this said something like `The BleakClient represents a connection to a BLE device`.

This may be fairly easily fixable by replacing the 
```
from xxxxxx import xxxxBleakClient as BleakClient
```

 in `__init__.py` with something like

```
from xxxxx import xxxxBleakClient as _BleakClientImplementation

class BleakClient(_BleakClientImplementation):
    """Insert user-friendly docstring here"""
```

My second problem is that the documentation is not complete. Again an example: the important bits in the `BLEDevice` object are the `name` and `address` attributes, but these are not documented anywhere, at least I could not find them. The `metadata` attribute is mentioned (at least for the MacOS backend) but the contents are not described.

My third problem (also mentioned in issue #277) as that it is difficult to work out how to combine functionality.

hbldh:
@jackjansen Thank you for the input!

I agree with the first part. It is due to API autogeneration issues, not being able to import all backends' code on the RTD build servers. I am reluctant to flip import order like you propose however. I plan on writing better docs without focusing on autogeneration instead.

The second problem is the main reason for this issue.

As for the third, I am not sure on what level to write the docs on. I do not want to write the code that people just copy. I want them to think and know what they do as well. But better, more collected examples not only in the Github folder but also on RTD is planned.

[...]

PR #273, opened 2020-08-18
RSSI feature, Notifiy by handle fix, event.wait() fix

Carglglz:
I made the changes on top on feature/find-device-850aba9 yesterday, so that's why they are here.

### DIFF
--- a/bleak/backends/corebluetooth/client.py
+++ b/bleak/backends/corebluetooth/client.py
@@ -7,6 +7,7 @@ Created on 2019-6-26 by kevincar <kevincarrolldavis@gmail.com>
 import logging
 import uuid
 from typing import Callable, Any, Union
+import asyncio
 
 from Foundation import NSData, CBUUID
 from CoreBluetooth import (
@@ -93,7 +94,7 @@ class BleakClientCoreBluetooth(BaseBleakClient):
         self.services = BleakGATTServiceCollection()
         # Ensure that `get_services` retrieves services again, rather than using the cached object
         self._services_resolved = False
-        self._services = None 
+        self._services = None
         return True
 
     async def is_connected(self) -> bool:
@@ -381,3 +382,19 @@ class BleakClientCoreBluetooth(BaseBleakClient):
             raise BleakError(
                 "Could not stop notify on {0}: {1}".format(characteristic.uuid, success)
             )
+
+    async def get_rssi(self) -> int:
+        """To get RSSI value in dBm of the connected Peripheral"""
+
+        self._device_info.readRSSI()
+        manager = self._device_info.manager().delegate()
+        RSSI = manager.connected_peripheral.RSSI()
+        for i in range(20):  # First time takes a little otherwise returns None
+            RSSI = manager.connected_peripheral.RSSI()
+            if not RSSI:
+                await asyncio.sleep(0.1)
+            else:
+                return int(RSSI)
+
+        if not RSSI:
+            return None
--- a/bleak/backends/corebluetooth/PeripheralDelegate.py
+++ b/bleak/backends/corebluetooth/PeripheralDelegate.py
@@ -36,7 +36,7 @@ CBPeripheralDelegate = objc.protocolNamed("CBPeripheralDelegate")
 class _EventDict(dict):
     def get_cleared(self, xUUID) -> asyncio.Event:
         """ Convenience method.
-        Returns a cleared (False) event. Creates it if doesen't exits.
+        Returns a cleared (False) event. Creates it if doesn't exits.
         """
         if xUUID not in self:
             # init as cleared (False)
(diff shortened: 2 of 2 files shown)


Carglglz:
The RSSI feature should be mentioned here maybe? https://bleak.readthedocs.io/en/latest/backends/macos.html

Also`notify_callback` accepts now three inputs (handle, data, uuid), this should be implemented in Linux/Windows too or referenced in some way in the api? https://bleak.readthedocs.io/en/latest/api.html

bleak/backends/corebluetooth/client.py
```diff
@@ -381,3 +379,28 @@ async def stop_notify(
             raise BleakError(
                 "Could not stop notify on {0}: {1}".format(characteristic.uuid, success)
             )
+
+    async def update_RSSI(self, rtn: bool = False) -> int:
```
hbldh, line 383: I do not want to store the rssi in the client. Better to have this method and use it when needed since the rssi is fluctuating all the time. Call it `get_rssi`, remove the `self.rssi` and the `rtn` argument.

bleak/backends/corebluetooth/client.py
```diff
@@ -381,3 +379,28 @@ async def stop_notify(
...
+            rtn (Bool): Whether to return the updated RSSI value or not
+
+        """
+        self._device_info.readRSSI()
+        manager = self._device_info.manager().delegate()
+        RSSI = manager.connected_peripheral.RSSI()
+        for i in range(20):  # First time takes a little otherwise returns None
+            RSSI = manager.connected_peripheral.RSSI()
+            if not RSSI:
+                await as_sleep(0.1)
```
hbldh, line 396: Import `asyncio` and use `asyncio.sleep` instead. That makes it adhere to how the rest of Bleak looks like.

bleak/backends/corebluetooth/client.py
```diff
@@ -381,3 +379,28 @@ async def stop_notify(
...
+            else:
+                break
+
+        if not RSSI:
+            raise BleakError(
```
hbldh, line 401: Is it really necessary with an error? Is it not enough to return a None?

hbldh (changes requested):
I would rather merge the `feature/find-device` branch separately, so please base this on the `develop` branch instead or wait for me to merge #271 first.

The notify callback has one argument in Linux, two in Windows and three in Core Bluetooth. This is documented in each backends `register_detection_callback` method. During #266, on branch `feature/wtd` this will be referenced to in the interface docstring.

I do not want to abstract argument numbers away since that could possibly create a support situation that is detrimental to my mental health. The contents of the arguments returned are also very different things (native OS components objects) and it is better left to the user to handle this. Some things Bleak cannot abstract away, and I think this is one such thing.

Carglglz:
I fixed all the requested changes  👍 .

Carglglz:
> I would rather merge the `feature/find-device` branch separately, so please base this on the `develop` branch instead or wait for me to merge #271 first.
> 
> The notify callback has one argument in Linux, two in Windows and three in Core Bluetooth. This is documented in each backends `register_detection_callback` method. During #266, on branch `feature/wtd` this will be referenced to in the interface docstring.
> 
> I do not want to abstract argument numbers away since that could possibly create a support situation that is detrimental to my mental health. The contents of the arguments returned are also very different things (native OS components objects) and it is better left to the user to handle this. Some things Bleak cannot abstract away, and I think this is one such thing.

`register_detection_callback`?  I think you mean in the docstring of `start_notify` method ?
In the three backends the docstring is like this, and in base `client.py` too:
```
"""Activate notifications/indications on a characteristic.

        Callbacks must accept two inputs. The first will be a uuid string
        object and the second will be a bytearray.
```
The fix will pass three values to `notify_callback` : characteristic handle (int), value (bytearray), uuid (string)
```
if notify_callback:
            notify_callback(c_handle, value, cUUID)
```
So that was my question, how to deal with this change in the docs, but 
> During #266, on branch feature/wtd this will be referenced to in the interface docstring

may be a better place to discuss that. 👍

hbldh:
My bad, I missed the extra argument there in the disarray of all my own changes overlaid with this.

I would like there to be only two arguments: `c_handle` and `value`. It is still a `sender, data` solution and imo Bleak should have handle returned instead of uuid in all backends, since the handle identifies uniquely and the uuid does not. One can get the characteristic from either the handle or the uuid (provided it is unique), so I think this is better.

I will merge this, make the changes in this and other backends and documentation modifications regarding this. Hopefully it will not cause too much problems.

hbldh (approved):

hbldh:
@Carglglz In #275 you can continue the discussion on the uuid/handle situation if you would like to add something.

Carglglz:
> I will merge this, make the changes in this and other backends and documentation modifications regarding this. Hopefully it will not cause too much problems.

Yes, that was my concern, it could be a potentially breaking change and therefore should be properly documented as you did.
Also I didn't drop the uuid because I didn't realise this 

> One can get the characteristic from either the handle or the uuid (provided it is unique)

Maybe it worth to be mentioned in the docs, right next to callback example? or included in `enbale_notifications.py` example?
```
# sender uuid
BleakClient.services.characteristics[c_handle].uuid

# sender description
BleakClient.services.characteristics[c_handle].description
```
```
def notification_handler(sender, data):
    """Simple notification handler which prints the data received."""
    print("{0}: {1}".format(client.services.characteristics[sender].description, data))
```


So yes `c_handle` and `data` is the way to go.

PR #275, opened 2020-08-19
Change uuid to handle in data sent to notification callbacks

hbldh:
**POTENTIAL BREAKING CHANGE**: All notifications now have the characteristic's integer **handle** instead of its UUID as a string as the first argument ``sender`` sent to notification callbacks. This provides the uniqueness of sender in notifications as well.

This dawned on me when reviewing and merging #273. I think it is more correct this way, but it might break some existing code.

### DIFF
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -56,6 +56,7 @@ class BleakClientBlueZDBus(BaseBleakClient):
 
         self._disconnected_callback = None
 
+        # This maps DBus paths of GATT Characteristics to their BLE handles.
         self._char_path_to_handle = {}
 
         # We need to know BlueZ version since battery level characteristic
@@ -611,7 +612,7 @@ class BleakClientBlueZDBus(BaseBleakClient):
         """Activate notifications/indications on a characteristic.
 
         Callbacks must accept two inputs. The first will be a integer handle of the characteristic generating the
-        data and the second will be a ``bytearray``.
+        data and the second will be a ``bytearray`` containing the data sent from the connected server.
 
         .. code-block:: python
 
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -56,7 +56,7 @@ class BleakClientBlueZDBus(BaseBleakClient):
 
         self._disconnected_callback = None
 
-        self._char_path_to_uuid = {}
+        self._char_path_to_handle = {}
 
         # We need to know BlueZ version since battery level characteristic
         # are stored in a separate DBus interface in the BlueZ >= 5.48.
@@ -229,6 +229,7 @@ class BleakClientBlueZDBus(BaseBleakClient):
         Free all the allocated resource in DBus and Twisted. Use this method to
         eventually cleanup all otherwise leaked resources.
         """
+        self._char_path_to_handle.clear()
         await self._cleanup_notifications()
         await self._cleanup_dbus_resources()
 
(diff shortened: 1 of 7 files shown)


bleak/backends/bluezdbus/client.py
```diff
@@ -56,7 +56,7 @@ def __init__(self, address, **kwargs):
 
         self._disconnected_callback = None
 
-        self._char_path_to_uuid = {}
+        self._char_path_to_handle = {}
```
dlech, line 59: A comment explaining that this maps D-Bus paths to a BLE handles would be helpful.

bleak/backends/bluezdbus/client.py
```diff
@@ -605,17 +605,17 @@ async def write_gatt_descriptor(self, handle: int, data: bytearray) -> None:
...
-        callback: Callable[[str, Any], Any],
+        callback: Callable[[int, bytearray], None],
         **kwargs
     ) -> None:
         """Activate notifications/indications on a characteristic.
 
-        Callbacks must accept two inputs. The first will be a uuid string
-        object and the second will be a bytearray.
+        Callbacks must accept two inputs. The first will be a integer handle of the characteristic generating the
+        data and the second will be a ``bytearray``.
```
dlech, line 614: Would be nice to say that the bytearray contains the data itself.

bleak/backends/bluezdbus/client.py
```diff
@@ -605,17 +605,17 @@ async def write_gatt_descriptor(self, handle: int, data: bytearray) -> None:
...
 
         .. code-block:: python
 
-            def callback(sender, data):
+            def callback(sender: int, data: bytearray):
                 print(f"{sender}: {data}")
```
dlech, line 619: might want to change this to use logger instead of `print` while we are touching this code

dlech: same comment applies to other backends

bleak/backends/bluezdbus/client.py
dlech: same comment applies to other backends

bleak/backends/dotnet/client.py
```diff
@@ -739,7 +739,7 @@ def dotnet_notification_parser(sender: Any, args: Any):
         reader.Dispose()
 
         return loop.call_soon_threadsafe(
-            func, sender.Uuid.ToString(), bytearray(output)
+            func, int(sender.AttributeHandle), bytearray(output)
```
dlech, line 742: is AttributeHandle not already an int? winrt docs say it returns ushort.

Carglglz:
Yes, I do think this is the way to go. 👍

bleak/backends/bluezdbus/client.py
hbldh: Good idea.

bleak/backends/bluezdbus/client.py
hbldh: Also a good idea.

bleak/backends/bluezdbus/client.py
hbldh: I agree in general, but it then requires importing `logging` in the example code. I will push this potential documentation improvement forward to #266.

bleak/backends/dotnet/client.py
hbldh: It is. I never trust `pythonnet` to give me what I expect though. In the cases where I have not debugged it thoroughly, I tend to add a extra conversion just in case I get a `UInt32` object or similar instead.

Debugged and checked this out now. It was an `int`.

PR #278, opened 2020-09-01
Safer CoreBluetooth presence check

pvieito:
The previous check does not working on macOS Big Sur as it presumes that Core Bluetooth was already linked on the process.

### DIFF
--- a/bleak/__init__.py
+++ b/bleak/__init__.py
@@ -12,6 +12,7 @@ import logging
 import platform
 import subprocess
 import asyncio
+import ctypes
 
 from bleak.__version__ import __version__  # noqa
 from bleak.exc import BleakError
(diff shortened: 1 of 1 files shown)


hbldh:
I do not think that this is the way to go. I do not want to add a `ctypes` loading here.

I have no way of testing this myself, but try changing 

```python
from Foundation import NSClassFromString
```
to
```python
import objc
from Foundation import NSClassFromString
```
Does that fix it? If not, then something like
```python
import objc
try:
    from CoreBluetooth import CBPeripheral
except:
    raise BleakError("Bleak requires the CoreBluetooth Framework")
```

pvieito:
New PR: #280

yujie-tao:
Hi 

I have CoreBluetooth installed on my laptop. No problem to run the `from CoreBluetooth import CBPeripheral` in python script. However, when I try to execute the `from bleak import BleakClient`, I get the "Bleak requires the CoreBluetooth Framework" exception. I am running on Big Sur as well; I had no problem to import BleakClient with the older system. Do anyone know why is that?

Issue #255, opened 2020-07-14
New connections can't be made after some time.


LeoBortoni:
* bleak version: 0.7.0
* Python version: 3.7.7
* Operating System: Windows 10

### Description
So this is what I am trying to do since 2 months ago: My goal is to create application which reads some number N of sensors at the same time via bluetooth, collect its data and publish it on my data bank. It is a simple thing: I created a app which load the list of address of the sensors, and split it on another sublists of simultaneos_connections_list, so then for each sensor in the group of simultaneos_connections_list, I create a thread or async task to read this sensor (Yes, i created a wrapper of the bleak lib that allows me to use threads). This thread, or async task, do this: Create a Bleak client, search and try to connect with the sensor, get its data and publish it on the data bank. After this, is time for the other group of simultaneos_connections create its thread and perform the exact same task, but for others sensor's address. Finished all the groups, the while (True) loop restarts and the applications execute the creation of threads again. Every thing works pretty well for some time (both the async and thread version). The problem occours after 3 or 4 hours running the application. After this time, the bleak lib just cant connect with any other sensor. The bleak finds the bluetooth address and try to connect with the sensor, but it aways stops on the "Get Services..." log and than raise a empty exception, which I assume to be a timeout exception. The number of simultaneos connections N that I have tested are 1,2 and 3. I have tested many differents approaches and i will list it bellow, togehter with my assumptions of why this is happening. For each thread, i try to connect with the sensor 20 times, and each time I call the connect method from the bleak client like that: client.connect(timeout = 3). After try the connection 20 times, the thread throws a exception "Could not find the address". 

Note > For each measurement on my sensors, i create a thread, a bleak client on this thread, try to connect and read it, publish the result on the data bank and then I delet this client.

What is more weird, is that when this application can't connect with the sensor any more, even if i finalize the app and restart it, itself will not work anymore! It can't connect with the sensor, just like if it had been running for 4 hours and then stoped. Even if I restart the bluetooth module on the windows system, it does not connect with any sensor. I do have one test.py program which just connect with the sensor, reads its data and print on the screen, just to test the bleak library, and even this program dosen't work anymore, can't connect. The only thing that allows me to connect with my sensors again is to restart the computer. Then, i am able to run this test.py or the main application with threads and it works.

[...]

Issue #133, opened 2019-12-05
disconnect() doesn't disconnect from device


chunghaw-tseng:
* bleak version: 0.5.1
* Python version: 3.6
* Operating System: Windows 10

### Description

I am currently writing an application to connect to a BLE device, write to some characteristics and then disconnect from the device. From our tests, bleak's disconnect API doesn't always send the disconnect command to the device (seen using a snifer). There are cases that the devices remain connected after calling disconnect but bleak returns that they are already disconnected. Not sure where the problem might be, is there any patch for this problem?

### What I Did

I tried to disconnect, the return is True but looking at the BLE snifer the devices still transmit empty PDUs to maintain the connection. Calling self.client.is_connected() returns False. The connection has not been terminated.

```python
        try:
             response = await self.client.disconnect()  #self.client is the BleakClient 
             print("DISCONNECT RESPONSE", response)
        except BleakError:
              log.error("Device couldn't be disconnected")
        return response  #Returns true
```

hbldh:
Bleak uses, via pythonnet, the C# UWP libraries for connecting to Bluetooth devices. 
I have followed the [Universal Windows Platform (UWP) app samples](https://github.com/microsoft/Windows-universal-samples/blob/master/Samples/BluetoothLE/cs/Scenario2_Client.xaml.cs#L78-L97) method of disconnecting from a BluetoothLEDevice. That being said, it might be that Bleak, pythonnet and/or the C# bridge I have developed delay the disconnecting.

The Dispose method will eventually close the connection. There is a [`Close` method](https://docs.microsoft.com/en-us/uwp/api/windows.devices.bluetooth.bluetoothledevice.close#Windows_Devices_Bluetooth_BluetoothLEDevice_Close) that can be called, but I decided against it to avoid potential errors.

What is the real problem of the device not being freed immediately?

mitchmccarthy:
The peripheral device is intended to drop to a low-power mode after the disconnection, but this issue is causing the connection to persist even though the bleak client says it has disconnected.

We set up a BLE sniffer while we reproduced the issue, and there doesn't seem to be any TERMINATE signals sent over the air.

hbldh:
@mitchmccarthy This SO issue descibes to problems Bleak faces as well: https://stackoverflow.com/questions/39599252/windows-ble-uwp-disconnect

[...]

PR #279, opened 2020-09-02
Release/v0.8.0

### LINKED ISSUE #251: why does bleak set the logging level of the package instead of using whatever the caller sets?
* bleak version: latest
* Python version: n/a
* Operating System: n/a
* BlueZ version (`bluetoothctl -v`) in case of Linux: n/a

### Description

bleak sets the default logging level to DEBUG.  Although this can be overridden, is there some reason that it just doesn't use whatever the root logger is set to?

### LINKED ISSUE #258: Verbose notification logging on bluezdbus client with characteristic notifications
* bleak version: 0.7.1
* Python version: 3.7.3
* Operating System: Linux

### Description
The bleak bluezdbus backend client emits a logging message at both the `logging.DEBUG` and `logging.INFO` level every time a characteristic notification is received. This behavior is different than the `corebluetooth` and `dotnet` backends which do not log on every notification. This is causing my application to emit several hundred logging messages per second at the `logging.INFO` level, rendering the logging extremely verbose and not very useful. I am currently changing the logging level of `bleak.backends.bluezdbus.client` to `logging.WARNING`, which unfortunately also suppresses some messages that I would like to keep.

I believe that the bluezdbus client should behave the same way as the other two backends and not emit a logging message on characteristic notification.

See `BleakClientBlueZDBus._properties_changed_callback()`:
The `logging.DEBUG` message https://github.com/hbldh/bleak/blob/master/bleak/backends/bluezdbus/client.py#L777
The `logging.INFO` message https://github.com/hbldh/bleak/blob/master/bleak/backends/bluezdbus/client.py#L785

### What I Did
Connect to a BLE device under Linux and enable notifications on a characteristic that changes frequently.

### LINKED ISSUE #259: AttributeError raised when calling client.is_connected() after BLE device is no longer communicating
* bleak version: 0.7.1
* Python version: 3.7.3
* Operating System: Raspberry OS (Debian Linux)

### Description

I am seeing an `AttributeError` exception thrown under Linux when attempting to call `client.is_connected()` on a client who's BLE device is no longer communicating (turned off or out of range). I am not seeing this exception on Windows.

### What I Did

Minimal Example:
```python
import asyncio
import logging

from bleak import BleakClient

logging.basicConfig(level=logging.DEBUG)

async def run(address):
    log = logging.getLogger(__name__)

        x = await client.is_connected()
        log.info('Connected: {0}'.format(x))

        while True:
            print(await client.is_connected())
            await asyncio.sleep(0.1)

if __name__ == '__main__':
    address = # REDACTED
    asyncio.run(run(address))
```

Once I started to see "True" repeated on the command line, I turned off the BLE device that was connected, and that results in the following log with an exception raised about 2 seconds after the BLE device has been powered off (MAC address has been redacted):
```python
INFO:__main__:Connected: True
True
True
...
True
True
DEBUG:bleak.backends.bluezdbus.client:DBUS: path: /org/bluez/hci0/dev_REDACTED, domain: org.bluez.Device1, body: {'ServicesResolved': False, 'Connected': False}
DEBUG:bleak.backends.bluezdbus.client:Device REDACTED disconnected.
DEBUG:bleak.backends.bluezdbus.client:Removing rule PropChanged, ID: 1
DEBUG:bleak.backends.bluezdbus.client:Disconnecting from BLE device...
  File "disconnect_test.py", line 21, in <module>
    asyncio.run(run(address))
  File "/usr/lib/python3.7/asyncio/runners.py", line 43, in run
  File "/usr/lib/python3.7/asyncio/base_events.py", line 584, in run_until_complete
    return future.result()
  File "disconnect_test.py", line 16, in run
    print(await client.is_connected())
  File "/home/pi/.local/lib/python3.7/site-packages/bleak/backends/bluezdbus/client.py", line 282, in is_connected
    return await self._bus.callRemote(
AttributeError: 'NoneType' object has no attribute 'callRemote'
```

The same exception is raised if the BLE device is taken out of range or put into an RF shielded box as well. I have on occasion seen this successfully print 'False' once or twice before raising the exception as well when polling `client.is_connected()` at a faster rate, but it is inconsistent.

I have also seen this happen by calling `client.disconnect()` in my application after the BLE device is no longer communicating, but it is not consistent and I am unable to reproduce this with a minimal example at the moment, it looks like the root cause is likely the same as above though.
```python
...
  File "/home/pi/repos/REDACTED/REDACTED/device.py", line 521, in disconnect
    await self.client.disconnect()
  File "/home/pi/.local/lib/python3.7/site-packages/bleak/backends/bluezdbus/client.py", line 265, in disconnect
    is_disconnected = not await self.is_connected()
  File "/home/pi/.local/lib/python3.7/site-packages/bleak/backends/bluezdbus/client.py", line 282, in is_connected
    return await self._bus.callRemote(
AttributeError: 'NoneType' object has no attribute 'callRemote'
```

### LINKED ISSUE #276: set_disconnected_callback have different behavior. works on Linux but crashes on MacOS.
* bleak version: 0.7.1
* Python version: 3.8.5
* Operating System: macOS-10.15.6

### Description

`set_disconnected_callback` have different behavior on Linux and MacOS.

### What I Did

This works on Linux but crashes on MacOS.
```
    bleak_client.set_disconnected_callback(my_disconnect_callback)
    bleak_client.connect()
```
traceback:
```
  File "/usr/local/lib/python3.8/site-packages/bleak/backends/corebluetooth/client.py", line 111, in set_disconnected_callback
    manager = self._device_info.manager().delegate()
AttributeError: 'NoneType' object has no attribute 'manager'
```
(this as `manager` is None until connected.)
Changing the order (i.e. call `set_disconnected_callback` after `connect`) do not seem like a good solution as disconnect can occur before callback is set. Perhaps `set_disconnected_callback` could be deprecated  and replaced with a argument to either `BleakClient` or `BleakClient:connect()`? I suspect it would make it easier to handle this uniformly for all backends?

hbldh:
# Version 0.8.0

## Added

* Added ``find_device_by_address`` method to the ``BleakScanner`` interface, for stopping scanning when a desired address is found.
* Implemented ``find_device_by_address`` in the .NET backend ``BleakScanner`` implementation and switched its ``BleakClient`` implementation to use that method in ``connect``.
* Implemented ``find_device_by_address`` in the BlueZ backend ``BleakScanner`` implementation and switched its ``BleakClient`` implementation to use that method in ``connect``.
* Implemented ``find_device_by_address`` in the Core Bluetooth backend ``BleakScanner`` implementation and switched its ``BleakClient`` implementation to use that method in ``connect``.
* Added text representations of Protocol Errors that are visible in the .NET backend. Added these texts to errors raised.
* Added pairing method in ``BleakClient`` interface.
* Implemented pairing method in .NET backend.
* Implemented pairing method in the BlueZ backend.
* Added stumps and ``NotImplementedError`` on pairing in macOS backend.

## Changed

* **BREAKING CHANGE** All notifications now have the characteristic's integer **handle** instead of its UUID as a
  string as the first argument ``sender`` sent to notification callbacks. This provides the uniqueness of
  sender in notifications as well.
* Version 0.5.0 of BleakUWPBridge, with some modified methods and implementing ``IDisposable``.
* Merged #224. All storing and passing of event loops in bleak is removed.
* Removed Objective C delegate compliance checks. Merged #253.

## Fixed

* .NET backend loop handling bug entered by #224 fixed.
* Removed default ``DEBUG`` level set to bleak logger. Fixes #251.
* More coherency in logger uses over all backends. Fixes #258
* Attempted fix of #255 and #133: cleanups, disposing of objects and creating new ``BleakBridge`` instances each disconnect.
* Fixed some type hints and docstrings.
* Modified the ``connected_peripheral_delegate`` handling in macOS backend to fix #213 and #116.
* Merged #270, fixing a critical bug in ``get_services`` method in Core Bluetooth backend.
* Improved handling of disconnections and ``is_connected`` in BlueZ backend to fix #259.
* Fix for `set_disconnected_callback` on Core Bluetooth. Fixes #276.


hbldh:
This release PR has been active for quite some time. I will approve it and then work from version 0.8.0 and forward.

PR #280, opened 2020-09-02
Safer `Core Bluetooth` presence check

pvieito:
The previous check does not working on macOS Big Sur as it presumes that Core Bluetooth was already loaded on the process.

Superseeds #278

### DIFF
--- a/bleak/__init__.py
+++ b/bleak/__init__.py
@@ -51,9 +51,9 @@ if platform.system() == "Linux":
         BleakClientBlueZDBus as BleakClient,
     )  # noqa
 elif platform.system() == "Darwin":
-    from Foundation import NSClassFromString
-
-    if NSClassFromString("CBPeripheral") is None:
+    try:
+        from CoreBluetooth import CBPeripheral
+    except:
         raise BleakError("Bleak requires the CoreBluetooth Framework")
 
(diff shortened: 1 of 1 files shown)


bleak/__init__.py
```diff
@@ -51,9 +51,9 @@
         BleakClientBlueZDBus as BleakClient,
     )  # noqa
 elif platform.system() == "Darwin":
-    from Foundation import NSClassFromString
-
-    if NSClassFromString("CBPeripheral") is None:
+    try:
+        from CoreBluetooth import CBPeripheral
+    except:
```
dlech, line 56: How about this instead?

```python
except Exception as ex:
    raise BleakError("Bleak requires the CoreBluetooth Framework") from ex
```
Then we don't lose information from the first exception.

pvieito: 👍🏻

hbldh: You have got to be kidding me! I have been coding Python for 12 years now and this is the first time I have seen or heard of `raise x from y`!

🤯

I have many times lamented Python's lack of good exception chaining and it has been there all along...

@dlech thank you for this revelation!

dlech: :grin:

Issue #284, opened 2020-09-09
Finding the MAC address on macOS


bardiabarabadi:
* bleak version: 0.7.1
* Python version: 3.8
* Operating System: macOS Catalina 10.15.6

### Description

I managed to get a list of BLE devices using the same procedure as the scanner.py, i.e. the `discover()` method. Since I am using bleak on macOS, the returned device type is `BLEDeviceCoreBluetooth` which has some attributes including `address`.

The problem is that according to the documentation, the address field when using the macOS backend is actually the CBPeripheral identifier (and not the MAC address). Of course, this is not an issue by itself, but when I want to connect to the device using the `BleakClient` class, I need to have the MAC address (not the identifier).

Apparently, it is not possible to obtain the MAC address from the identifier (it changes every time).

Can you please tell me how can I get the MAC address, or alternatively use the identifier to connect to the BLE device via `BleakClient` class?

Thank you.

dlech:
Please see #140. Apple does not expose the Bluetooth address for security/privacy reasons.

bardiabarabadi:
Thanks for the response @dlech 
Is there any way to connect to the BLE device using the `BleakClient` class on macOS (considering the fact that the Bluetooth address is not available?)

dlech:
If you are using the `discover()` function, then just call the `connect()` method on the device that was discovered.

bardiabarabadi:
Yes, I am using the `discover()` function. The only `connect()` method I found is in bleak/bleak/backends/corebluetooth/client.py in the `BleakClientCoreBluetooth` class. However, `BleakClientCoreBluetooth` requires an address to initialize (that is not available). Is this the `connect()` function you are referring to? If not, can you show me a quick example of how to import and use it?

dlech:
Right. So you can use the `address` attribute of the discovered device to create the client object.

```python
import asyncio
from bleak import discover

async def run():
    devices = await discover()
    for d in devices:
        print(d)
        client = BleakClient(d.address)
        await client.connect()
        # do something with client


loop = asyncio.get_event_loop()
loop.run_until_complete(run())
```

bardiabarabadi:
Thank you for the code. That was actually the first thing I tried before opening this case. However, I rewrote the code to make sure that I am not missing anything and I got the same error (Device with address xxxxxx-xx-xxx-xxxxx was not found)

here is my code and the error I get:
```
import asyncio
from bleak import discover, BleakClient

async def find_muses():
    devices = await discover()
[...]

PR #285, opened 2020-09-10
Connect by BLEDevice as well as string Bluetooth adress

### LINKED ISSUE #191: Connecting via BleakClient with address does an additional discovery.
* bleak version: latest
* Python version: 3.8
* Operating System: macOS 

### Description

I noticed it was taking about 5 seconds to discover devices and then connect to a discovered device (another 5 seconds)... After looking at the code it appears that the client implementation does a second discovery which is painfully slow. Is there any way we can just pass in a populated device or just connect to the uuid directly instead of rediscovering?

https://github.com/hbldh/bleak/blob/08f4fc4cfb9b8c334b31d1e0ee76b6088684ec44/bleak/backends/corebluetooth/client.py#L63

### What I Did

```
async def discoverDevice() -> BLEDevice:
    devices = await discover()
    for d in devices:
        name = d.details.name()
        if name is not None and "TEST" in name:
            return d

    return None


async def run(loop):
    device = await discoverDevice()
    if device is None:
        return int(-1)

    # this does another discovery..
    async with BleakClient(device.address, loop=loop) as client:

loop.run_until_complete(run(loop))
```

hbldh:
* Added the possibility to connect using ``BLEDevice`` instead of a string adress. This allows for skipping the discovery call when connecting.
* Renamed ``BleakClient`` argument ``address`` to ``address_or_device``. 

Fixes #191 
Improves #284

### DIFF
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -12,6 +12,7 @@ from typing import Callable, Any, Union
 
 from twisted.internet.error import ConnectionDone
 
+from bleak.backends.device import BLEDevice
 from bleak.backends.service import BleakGATTServiceCollection
 from bleak.backends.characteristic import BleakGATTCharacteristic
 from bleak.exc import BleakError
@@ -36,20 +37,23 @@ class BleakClientBlueZDBus(BaseBleakClient):
     Implemented by using the `BlueZ DBUS API <https://docs.ubuntu.com/core/en/stacks/bluetooth/bluez/docs/reference/dbus-api>`_.
 
     Args:
-        address (str): The  address of the BLE peripheral to connect to.
+        address_or_device (`BLEDevice` or str): The Bluetooth address of the BLE peripheral to connect to or the `BLEDevice` object representing it.
 
     Keyword Args:
         timeout (float): Timeout for required ``BleakScanner.find_device_by_address`` call. Defaults to 10.0.
 
     """
 
-    def __init__(self, address, **kwargs):
-        super(BleakClientBlueZDBus, self).__init__(address, **kwargs)
+    def __init__(self, address_or_device: Union[BLEDevice, str], **kwargs):
+        super(BleakClientBlueZDBus, self).__init__(address_or_device, **kwargs)
         self.device = kwargs.get("device") if kwargs.get("device") else "hci0"
-        self.address = address
+        self.address = address_or_device
 
         # Backend specific, TXDBus objects and data
-        self._device_path = None
+        if isinstance(address_or_device, BLEDevice):
+            self._device_path = address_or_device.details["path"]
+        else:
+            self._device_path = None
         self._bus = None
         self._reactor = None
         self._rules = {}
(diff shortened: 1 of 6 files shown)


bleak/backends/bluezdbus/client.py
```diff
@@ -36,20 +37,23 @@ class BleakClientBlueZDBus(BaseBleakClient):
...
     Keyword Args:
         timeout (float): Timeout for required ``BleakScanner.find_device_by_address`` call. Defaults to 10.0.
 
     """
 
-    def __init__(self, address, **kwargs):
-        super(BleakClientBlueZDBus, self).__init__(address, **kwargs)
+    def __init__(self, address_or_device: Union[BLEDevice, str], **kwargs):
+        super(BleakClientBlueZDBus, self).__init__(address_or_device, **kwargs)
         self.device = kwargs.get("device") if kwargs.get("device") else "hci0"
```
dhalbert, line 49: There are two different objects named `device` here, one is a BLEDevice, and the other is the BLE adapter name. I was confused by this at first: I didn't even know that you could pass the adapter name as an optional kwarg.

I'd suggest either renaming the argument to `address_or_ble_device`, or, better, I think, renaming the kwarg to `adapter` or something like that. It looks like the kwarg is only present in the bluez backend, so that incompatible change would not be widespread.

dhalbert (changes requested):
Thank you ! I tested this in a simple program, and it worked for me.

I have one comment re argument names.

bleak/backends/bluezdbus/client.py
hbldh: Hmm.. Right you are. That is not a very pleasant clash...

I disliked the `address_or_device` argument name as well and hoped that a better one would come up in code review. I will think about it for a while. Maybe something closer to `peripheral`...

dhalbert: You could split the functionality into a regular constructor and a class method. One would take an address, the other would take a device.

Upward compatible:
`BleakClient(address: str)`
`BleakClient.from_device(device: BLEDevice)`

Breaking:
`BleakClient(device: BLEDevice)`
`BleakClient.from_address(address: str)`

In most of the code I've written (not in bleak), the second case is the most common. I do a filtering scan looking for a device advertising a specific service or having a specific name. I don't know the address in advance.

dlech: > Breaking:
> `BleakClient(device: BLEDevice)`
> `BleakClient.from_address(address: str)`

Other than the fact that this is breaking, I like this. Although we could just leave out the `from_address()` method since there is already `BleakScanner.find_device_by_address()`. ("There should be one-- and preferably only one --obvious way to do it.")

hbldh: I usually like to write classmethods for these kinds of special inits, but here I think it makes the basic usage more complex in a non-productive way. I would like to keep the init of a BleakClient as simple as possible, and rather accept two widely different kind of inputs to avoid confusing people. I get too many mails requesting that I write Bleak code as it is...

I plan on this to be very thouroughly documented in #266.

dhalbert (approved):
This solution to the naming clash is fine by me, and I'm looking forward to using this new functionality.

PR #288, opened 2020-09-14
Format with black

dlech:
Some commits were made without proper formatting. Other changes are due to breaking changes in black v20.

A task is added in CI to ensure that future commits are properly formatted according to black.


dlech:
Black fails to install on Python 3.5 and there is an additional message from pip that says Python 3.5 has reached end of life (yesterday). So I have added a commit to also remove Python 3.5 support.

https://github.com/hbldh/bleak/runs/1113651896?check_suite_focus=true#step:7:6

hbldh:
These are good thing. Please merge them and then rebase the `release/v0.8.0` onto `develop` so that it makes it way into next release.

hbldh:
One thing though, I would like to  or have Black break CI builds on develop branch due to that one has missed running it. I welcome it on release branches, but I dislike it generating errors in development.

dlech:
I don't understand the last comment. Are you saying you don't want the develop branch to run black?

hbldh:
Sorry, misspelled some. Does running black in the CI raise any errors and make builds fail if the committed code needed to be blacked?

dlech:
Probably. If you don't like that, we could add it to a different CI job. However, most projects that I have worked on that use automatic code formatting require that each commit is properly formatted so that later "clean ups" don't pollute the git blame.

Another benefit for me is that if all commits are properly formatted, I can turn on the "format on save" feature in VS code and never have to worry about formatting at all when I write the code since it will automatically run black each time I save a file.

dhalbert:
We have been using GitHub Actions [pre-commit](https://pre-commit.com/) action to run black and pylint when a PR commit is pushed. See our Python library cookiecutter:
https://github.com/adafruit/cookiecutter-adafruit-circuitpython/blob/master/%7B%7B%20cookiecutter.library_name%20%7D%7D/.github/workflows/build.yml
https://github.com/adafruit/cookiecutter-adafruit-circuitpython/blob/master/%7B%7B%20cookiecutter.library_name%20%7D%7D/.pre-commit-config.yaml

hbldh:
I have multiple projects that I cannot use Black on, so I cannot have such auto-save stuff turned on.

I want Bleak to use Black, but I do not want to enforce it on the develop branch. I will forget it occasionally and I really dislike broken CI jobs due to my own negligence. One could argue (and be completely right) that I deserve them...

I prefer pre-commits that handles it in the background to just running Black and breaking the CI job, but it is more work. @dhalbert Thank you for the links! I will look into it eventually.

I can live with this. Merge it.

dlech:
> I have multiple projects that I cannot use Black on, so I cannot have such auto-save stuff turned on.

FWIW, it can be enabled on a per-project basis.



> I can live with this. Merge it.

OK, thanks.

hbldh:
If you use VS Code, yes. However, I am a "visual breakpoint"-addict and use PyCharm. 

It might be possible there, but I haven't done any further exploration except add it as an external command and run it when I need it.

PR #290, opened 2020-09-14
Fix device_info logic in dotnet client

dlech:
The logic was inverted. If a device was given in the BleakClient constructor, then the _device_info attribute is set to the Bluetooth address, otherwise it is set to None. So we should only trying scanning for the device if _device_info is None.


hbldh:
Ah. Very good catch. Will merge.

PR #291, opened 2020-09-14
Disconnected callback improvements.

dlech:
- Add Windows implementation
- Move `set_disconnected_callback()` implementation to base class (since all backends were the same)
- Allow passing disconnected callback as kwarg to `BleakClient()` constructor.

See individual commit messages for more details.


hbldh (approved):

hbldh:
I think it would be good to include this with the 0.8.0 release as well. Try to gather all potential breaking changes in one release in a sense.

PR #293, opened 2020-09-15
Fix getting handle in BlueZ backend

### LINKED ISSUE #292: No "Handle" entry in bluez characteristic dictionary
* bleak version: 0.8.0? (commit aa54b08)
* Python version: 3.8.2
* Operating System: Ubuntu 20.04
* BlueZ version (`bluetoothctl -v`) in case of Linux: 5.53
* txdbus: 1.1.1

My notification callbacks are now getting `None` as the first argument, which should be the integer handle. Tracking this down, it appears to be because the characteristic info collected during discovery via txdbus messages, in `BleakClientBlueZDBus.get_services()`, does not contain handles. That is, at https://github.com/hbldh/bleak/blob/aa54b08a7c5a491bd909917bedb3a582dc9929d5/bleak/backends/bluezdbus/client.py#L427-L432
the value of `char` is a dict without a `"Handle"` key. Here is a typical `char` dict:
```python
{'UUID': '0000fff5-0000-1000-8000-00805f9b34fb', 'Service': '/org/bluez/hci0/dev_50_F1_4A_52_AF_9F/service0023', 'Value': [], 'Flags': ['write']}
```
Am I mistaken in thinking there should there be a `"Handle"` key? It looks like you were expecting one, but guarding against the possibility it was missing.

dlech:
The "Handle" property of the org.bluez.GattCharacteristic1 D-Bus interface is server-only, so it isn't available to Bleak. Instead,
we can scrape the handle from the D-Bus path of the characteristic.

Full object path format for reference:

    [variable prefix]/{hci0,hci1,...}/dev_XX_XX_XX_XX_XX_XX/serviceXX/charYYYY

Fixes #292

### DIFF
--- a/bleak/backends/bluezdbus/characteristic.py
+++ b/bleak/backends/bluezdbus/characteristic.py
@@ -1,4 +1,3 @@
-import re
 from uuid import UUID
 from typing import Union, List
 
@@ -27,8 +26,6 @@ _GattCharacteristicsFlagsEnum = {
     # "authorize"
 }
 
-_handle_regex = re.compile("/char([0-9a-fA-F]*)")
-
 
 class BleakGATTCharacteristicBlueZDBus(BleakGATTCharacteristic):
     """GATT Characteristic implementation for the BlueZ DBus backend"""
@@ -39,17 +36,8 @@ class BleakGATTCharacteristicBlueZDBus(BleakGATTCharacteristic):
         self.__path = object_path
         self.__service_uuid = service_uuid
 
-        # The `Handle` attribute is added in BlueZ Release 5.51. Empirically,
-        # it seems to hold true that the "/charYYYY" that is at the end of the
-        # DBUS path actually is the desired handle. Using regex to extract
-        # that and using as handle, since handle is mostly used for keeping
-        # track of characteristics (internally in bleak anyway).
-        self._handle = self.obj.get("Handle")
-        if not self._handle:
-            _handle_from_path = _handle_regex.search(self.path)
-            if _handle_from_path:
-                self._handle = int(_handle_from_path.groups()[0], 16)
-        self._handle = int(self._handle)
+        # D-Bus object path contains handle as last 4 characters of 'charYYYY'
+        self._handle = int(object_path[-4:], 16)
 
     @property
     def service_uuid(self) -> str:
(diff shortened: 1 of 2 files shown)


dlech:
Hi @dhalbert, could you please test this? I haven't had a chance to test it yet.

bleak/backends/bluezdbus/characteristic.py
```diff
@@ -39,17 +36,8 @@ def __init__(self, obj: dict, object_path: str, service_uuid: str):
         self.__path = object_path
         self.__service_uuid = service_uuid
 
-        # The `Handle` attribute is added in BlueZ Release 5.51. Empirically,
-        # it seems to hold true that the "/charYYYY" that is at the end of the
-        # DBUS path actually is the desired handle. Using regex to extract
-        # that and using as handle, since handle is mostly used for keeping
-        # track of characteristics (internally in bleak anyway).
-        self._handle = self.obj.get("Handle")
```
hbldh, line 47: So, basically it always got `None` here and instead always parsed it from the path instead? At least we know that it is a stable method then...

bleak/backends/bluezdbus/client.py
```diff
@@ -437,7 +437,9 @@ async def get_services(self) -> BleakGATTServiceCollection:
             self.services.add_characteristic(
                 BleakGATTCharacteristicBlueZDBus(char, object_path, _service[0].uuid)
             )
-            self._char_path_to_handle[object_path] = char.get("Handle")
+
+            # D-Bus object path contains handle as last 4 characters of 'charYYYY'
+            self._char_path_to_handle[object_path] = int(object_path[-4:], 16)
```
hbldh, line 442: This is the actual fix for the problem then. But I agree that this is more concise and plain to read than the regex.

hbldh:
I have verified it on my Ubuntu 18.04.3 LTS installation and it works well. Good job. This should be PR:ed to master as soon as possible and included in the 0.8.0 release.

dhalbert (approved):
I tested this on Ubuntu 20.04 with bluez 5.5.3. Works fine as well. Thank you!

bleak/backends/bluezdbus/characteristic.py
dlech: correct

dlech:
Thanks for testing!

PR #306, opened 2020-09-22
Adding timeout to dbus call as well to make sure it doesn't hang forever

cspensky:
I have seen cases where this would get hung up.  It seems like it can be fixed by simply passing the same argument into the DBus request.  In theory the timeout could be 2x the `timeout` value.  If we care about that, we could do `timeout/2` but the intent is moreso that it does not hang forever.


hbldh (approved):
I actually did not know there was a `timeout` argument to `callRemote`. Very good.

hbldh:
Thank you for this fix!

PR #307, opened 2020-09-22
`_uuid` is not defined.  Updating to point to `uuid` property.

cspensky:
Seems like some old code was hanging around.  Updating to simply use the `uuid` property.  Should be a straightforward fix.


hbldh (approved):

hbldh:
Thank you!

PR #311, opened 2020-09-28
Prevent overwriting address in dbus client

soldag:
This will cause errors when passing a `BleDevice` instance as argument. Base class already handles this and sets the address according to the type of `address_or_ble_device`.


hbldh:
Very good; thank you!

PR #318, opened 2020-10-04
github: use --check instead of --diff for black

dlech:
--diff does not cause a test to fail, so it is easy to miss

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -27,16 +27,8 @@ jobs:
                 run: python -m pip install --upgrade pip setuptools wheel
             -   name: Install dependencies
                 run: pip install -r requirements.txt
-            -   name: Install develoment dependencies
+            -   name: Install development dependencies
                 run: pip install -r requirements_dev.txt
-            -   name: Check code formatting with black
-                run: black . --diff
-            -   name: Lint with flake8
-                run: |
-                    # stop the build if there are Python syntax errors or undefined names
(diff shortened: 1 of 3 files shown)


dlech:
There are two problems with this change:

1. It doesn't show the changes needed to fix the failure.
2. It causes any test after it to fail.

hbldh:
Yes, I noticed that the diff caused errors on some platforms and not others...

I agree with using `--check` instead. Code needs to be properly `black`ed before merging then. 

1. It is irrelevant that the changes needed to fix is not shown, since running black and committing will solve it without the user knowing, right?
2. I can accept that. Fix code formatting first, commit that and then handle tests. Not that we have many tests though...

dlech:
>     1. It is irrelevant that the changes needed to fix is not shown, since running black and committing will solve it without the user knowing, right?

If it is only one or two lines that need to be changed, it can be faster to manually make the change compared to installing black.



>     2. I can accept that. Fix code formatting first, commit that and then handle tests. Not that we have many tests though...

I will also look at moving linting to a separate job. Really we only need to lint once instead of for every combination of OS and Python runtime.

hbldh:
1. I understand. Will Black be satisfied if the changes are not exactly as it wants them? It is "my way or the highway" kind of code formatter...
One could catch the exit code of `--check` and if it is `>0` then do a `--diff` call, print output and then return the exit code?

2. Yes. It is a bit wasteful of Action minutes.

dlech:
This is fixed up now.

Issue #319, opened 2020-10-05
'NoneType' object has no attribute 'delegate' exception when connecting


0ge:
* bleak version: 0.8.0
* Python version: Python 3.7.4
* Operating System: macOS 10.15.2 (Catalina)

### Description

I'm trying to write an application that uses tkinter and bleak to discover and interact with a custom BLE device.
However, when I try to connect to the device, it throws an exception:

```
'NoneType' object has no attribute 'delegate'
  File ".../scratch.py", line 29, in run
    await client.connect()
  File ".../miniconda3/envs/py3/lib/python3.8/site-packages/bleak/backends/corebluetooth/client.py", line 86, in connect
    manager = self._device_info.manager().delegate()
```

(There is also a similar exception thrown in the final clause when attempting to disconnect)

I have condensed this to the application shown below.

```python
import asyncio
import re
import traceback
import tkinter as tk
from bleak import BleakClient
from bleak import BleakScanner


async def run():
    if explicit:
        devices = await BleakScanner.discover(loop=loop)
    else:
        devices = await BleakScanner.discover()

    for d in devices:
        print(d)

    matcher = re.compile(r"[A-z]{3}-[0-9]{3}")
    devices = [d for d in devices if matcher.match(d.name)]
    if explicit:
        client = BleakClient(devices[0].address, loop=loop)
    else:
        client = BleakClient(devices[0].address)

    try:
        if explicit:
            await client.connect(loop=loop)
        else:
            await client.connect()
        print("Connected to {0}".format(client))
    except Exception as e:
        print(e)
        traceback.print_tb(e.__traceback__)
    finally:
        await client.disconnect()


def scan_and_connect():
    asyncio.get_event_loop().create_task(run())


async def update_ui():
    while True:
        root.update()
        await asyncio.sleep(1 / 30)


explicit = False  # Explicit or implicit run loops
loop = asyncio.get_event_loop()

root = tk.Tk()
if explicit:
    root.loop = loop
btn = tk.Button(root, text="Scan and connect", command=scan_and_connect)
btn.grid(column=1, row=1)

loop.run_until_complete(update_ui())

# Works as expected (i.e. connects to device)
# loop.run_until_complete(run())
```

### What I Did

I've tried the following combinations:

Bleak | Loop      | Outcome
------|---------|-----------
v0.7.1 | Implicit  | Exception
v0.7.1 | Explicit  | Exception
v0.8.0 | Implicit | Exception
v0.8.0 | Explicit | Exception

[...]

PR #323, opened 2020-10-06
Fix corebluetooth client issue #319

Carglglz:
This fix:
 - passing a `BLEDevice` to `BleakClientCoreBluetooth`
 - ~~raising a `BleakError` if the device returns `None` instead of `CBCentralManager`~~
 - set logging level to debug in `get_services` method
- weak reference of `CBPeripheral` to `CBCentralManager` and `CentralManagerDelegate`, now delegate is passed directly to `BLEDevice`


hbldh:
Content looks good, but you have either run Black with an old version or with incorrect config, since it reformatted a lot of code that broke the CI checks. Make that right and squash all commits to one and we will look into merging this.

Carglglz:
>  but you have either run Black with an old version

Yes, that was the reason. :/ 
All checks passed now 👍

bleak/backends/corebluetooth/client.py
```diff
@@ -80,12 +79,17 @@ async def connect(self, **kwargs) -> bool:
                 raise BleakError(
                     "Device with address {} was not found".format(self.address)
                 )
-
-        logger.debug("Connecting to BLE device @ {}".format(self.address))
-
-        manager = self._device_info.manager().delegate()
-        await manager.connect_(self._device_info)
-        manager.disconnected_callback = self._disconnected_callback_client
+        # self._device_info.manager() should return a CBCentralManager
```
dlech, line 82: I think a simpler change would be:

```diff
diff --git a/bleak/backends/corebluetooth/client.py b/bleak/backends/corebluetooth/client.py
index 954479d..093695b 100644
--- a/bleak/backends/corebluetooth/client.py
+++ b/bleak/backends/corebluetooth/client.py
@@ -73,7 +73,7 @@ class BleakClientCoreBluetooth(BaseBleakClient):
                 self.address, timeout=timeout
             )

-            if device:
+            if device and device.details:
                 self._device_info = device.details
             else:
                 raise BleakError(
```

But the deeper question is why `device.details` would be `None`.

bleak/backends/corebluetooth/client.py
```diff
@@ -80,12 +79,17 @@ async def connect(self, **kwargs) -> bool:
...
+        if hasattr(self._device_info.manager(), "delegate"):
```
dlech, line 83: In https://github.com/hbldh/bleak/issues/319#issuecomment-704403926 it is mention that `self._device_info.manager()` is `None` rather than just missing the `delegate` attribute. So maybe check for `is not None` instead?

Also, I'm wondering if adding this sort of check is just hiding a deeper issue.

Carglglz: > Also, I'm wondering if adding this sort of check is just hiding a deeper issue.


It raises a `BleakError` with the same message as if the device was not detected. Maybe it would be worth to change it.

Also I though of another check when a `CBPeripheral` object is discovered to assert that calling `.manager()` should return a `CBCentralManager` instance instead of `None`?

I agree with the deeper issue. But it is hard to debug without the device that is failing. Or how would you reproduce this?
It may be CoreBluetooth itself that fails silently? Maybe in console logging appear some sort of error message?

dlech: Looking at https://developer.apple.com/documentation/corebluetooth/cbperipheral?language=objc, I don't see that `manager` is a property or method on `CBPeriperal`. Could we be using an undocumented feature? Maybe we need to keep track of the `CBCentralManager` and the `CBPeriperal` objects separately? Or am I missing something?

Carglglz: >  I don't see that manager is a property or method on `CBPeriperal`.  Could we be using an undocumented feature? 

Yes that seemed strange to me too. 🙄 

> Maybe we need to keep track of the CBCentralManager and the CBPeriperal objects separately? Or am I missing something?

This could be a solution, I've just done a test passing directly `CentralManagerDelegate` to BLEDevice and it seems to work nice.

Updating this PR now with these changes 👍

bleak/backends/corebluetooth/scanner.py
```diff
@@ -113,6 +113,7 @@ async def get_discovered_devices(self) -> List[BLEDevice]:
                     details,
                     uuids=uuids,
                     manufacturer_data=manufacturer_data,
+                    delegate=details.manager().delegate(),
```
dlech, line 116: It would probably be safer to use `self._manager.delegate()` since we have found that `details.manager()` is undocumented and may not always exist in the future.

bleak/backends/corebluetooth/client.py
```diff
@@ -47,10 +47,10 @@ def __init__(self, address_or_ble_device: Union[BLEDevice, str], **kwargs):
 
         if isinstance(address_or_ble_device, BLEDevice):
             self._device_info = address_or_ble_device.details
+            self._delegate = address_or_ble_device.metadata.get("delegate")
         else:
             self._device_info = None
-
-        self._device_info = None
+            self._delegate = None
```
dlech, line 53: It could be nice to call this `self._central_manager_delegate` or just `_manager` instead to make it clear which delegate object this is since there is also a `CBPeriperalDelegate` object.

bleak/backends/corebluetooth/scanner.py
Carglglz: Yes I agree with that although `CentralManagerDelegate` is `self._manager.central_manager.delegate()` (checked `self._manager.delegate()` but  `self._manager` is not  the `CBCentralManager`. Pretty confusing name convention in my opinion. 😕

bleak/backends/corebluetooth/client.py
Carglglz: Yes totally agree with this, updating the PR 👍

bleak/backends/corebluetooth/scanner.py
dlech: Doh, I am probably to blame for the bad naming there. We can clean that up later.

dlech:
Merged, thanks!

PR #324, opened 2020-10-07
Cleanups

dlech:
These are some various code cleanups. Mostly suggested by flake8.


dlech:
We should be careful with the "Remove unused imports" commit. It it probably removes some stuff that is used in later pull requests like `Object` in the dotnet `client.py`.

hbldh:
Very good. We wait with this one until #328 is merged.

PR #327, opened 2020-10-08
UUID fixes and cleanups

dlech:
I recently ran into an issue where calling methods such as` BleakGattService.get_characteristic()` with a string for the UUID that contained uppercase letters would fail even though the UUID was otherwise correct.

This fixes that issue and includes some more cleanups in related code.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,11 +16,20 @@ Added
 * Timeout for BlueZ backend connect call to avoid potential infinite hanging. Merged #306.
 * Added Interfaces API docs again.
 
+Changed
+~~~~~~~
+
+* ``BleakCharacteristic.description()`` on .NET now returns the same value as
+  other platforms.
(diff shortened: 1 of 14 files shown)


dlech:
I am also wondering if using the standard Python `uuid.UUID` type would be more robust than trying to always remember to convert UUIDs to lower case strings everywhere. But unfortunately, that would be a breaking change in some cases (e.g. properties that return a UUID).

hbldh:
When I did the handles change, I was thinking about enforcing proper uuid handling only. I left it alone, because I felt uncertain that I would be capable of testing all those changes.

It would still improve and not break anything if `uuid.UUID` were used internally only. At least that is what I think. It would add some conversion overhead through, but remedy the upper- and lowercase situation. It would also avoid all ugly uuid string handling...

hbldh (approved):

Issue #313, opened 2020-09-28
AttributeError: 'NoneType' object has no attribute 'Dispose' when disconnecting devices that have previously disconnected due to range/power off/etc


NbrTDB:
* bleak version: 0.8.0
* Python version: 3.8.5
* Operating System: Windows 10 2004, build 19041.508

### Description

I've found a repeatable scenario wherein Bleak throws an AttributeError on attempting to `await client.disconnect()` a device which had previously disconnected itself and subsequently been reconnected to. I am not certain if this is a genuine issue or just a misuse of Bleak. In any case, catching the AttributeError still does not result in the ability to disconnect the device (it will just throw the same error each time).

In short:

- Spawn a new client and connect to a device
- Cause the device to disconnect on its end (e.g. turn off Bluetooth)
- Discard the client
- Reconnect to the device using a new client
- Attempt to disconnect; first attempt appears to do nothing and device does not disconnect. Any callbacks set on disconnect do not activate and the device does not indicate a disconnection has occurred.
- Second attempt throws AttributeError on the `await client.disconnect()`

### What I Did

I put together a snippet that reproduces the problem for me; it requires a device where the BLE connection can quickly be severed on the device side and said device be ready to reconnect again quickly (other devices should work if the timings are adjusted to suit)

```python
    async def __attributeerror_test(self, mac):
        print("Connecting to mac: {0}".format(mac), flush=True)
        client = BleakClient(mac)
        await client.connect()
        print("Poking services...", flush=True)
        await client.get_services()
        # now disconnect        
        print("Disconnect on device end now...", flush=True)
        for x in range(0,5):
            print("{}...".format(x))
            await asyncio.sleep(1)
        print("Waiting for device to indicate disconnection... ensure it is ready to reconnect", flush=True)
        for x in range(0,15):
            print("{}...".format(x))
            await asyncio.sleep(1)
        # if a disconnection callback had been set, it should have fired by now
        # must make a new client for a new connection
        del client
        print("Reconnecting...", flush=True)        
        client = BleakClient(mac)
        await client.connect()
        print("Poking services...", flush=True)
        await client.get_services()
        print("Disconnecting...", flush=True)
        await client.disconnect()
        print("Waiting for device to indicate disconnection...", flush=True)
        for x in range(0, 10):
            print("{}...".format(x))
            await asyncio.sleep(1)
        print("At this point, nothing has happened and the device still shows as connected... trying again", flush=True)
        print("Disconnecting...", flush=True)
[...]

Issue #312, opened 2020-09-28
Disconnected callbacks firing multiple times when multiple connections are made in a session


NbrTDB:
* bleak version: 0.8.0
* Python version: 3.8.5
* Operating System: Windows 10 2004, build 19041.508

### Description

When connecting to a BLE device, I am creating a new client and setting a disconnected callback using `client.set_disconnected_callback()`, then storing the client in `self.connected_devices { }` with a reference against the MAC address.

This all works great including the firing off the callback when the device is connected (either by `client.disconnect()` or wandering out of range, powered off, etc). The disconnect callback discards the client by deleting it from `self.connected_devices` when fired.

However after reconnecting to the same device in the same session (and again creating a new client, assigning the disconnected callback and stuffing it in the dict of connected devices), on disconnect the callback with run twice. And on the third time, three times, and so on.

I have determined as a workaround, I could try separately tracking all MAC addresses (a `self.client_history` dict, perhaps) and then avoid setting disconnection callbacks on any MAC addresses I find in there, but I wonder if this is going about things the wrong way? Is there some way the client instance is surviving having been discarded? It's as if there is a stale reference to previously-connected devices somewhere.

### What I Did

```python
    async def __connect_device(self, mac):
        # connect to a device based on a mac address 
        print("Connecting to mac: {0}".format(mac),flush=True)
        client = BleakClient(mac)
        try:
            await client.connect()
        except BleakError as err:
            print(err)
        else:
            client.set_disconnected_callback(self.__disconnect_callback)
            await client.get_services()
            self.connected_devices[mac] = client

    async def __disconnect_device(self, mac):
        # connect to a device based on a mac address 
        print("Disconnecting mac: {0}".format(mac),flush=True)
        if mac in self.connected_devices:
            await self.connected_devices[mac].disconnect()

    def __disconnect_callback(self, device_client):
        # fires when a device is disconnected
        print("Disconnected from {0}".format(device_client.address))
        mac = device_client.address
        # remove from self.connected_devices
        if mac in self.connected_devices:
            del self.connected_devices[mac]
```

dlech:
Thanks for the feedback. The disconnect callback on Windows in new in v0.8.0, so we have probably overlooked something. I will look deeper when I have a chance.

NbrTDB:
[...]

PR #328, opened 2020-10-08
Windows connection fixes

dlech:
This fixes some issues with the Windows disconnect callback implementation (e.g. #312 and #313). 

It also contains further improvements for managing the connection lifecycle.

### DIFF
--- a/bleak/backends/dotnet/client.py
+++ b/bleak/backends/dotnet/client.py
@@ -112,6 +112,7 @@ class BleakClientDotNet(BaseBleakClient):
             self._device_info = None
         self._requester = None
         self._bridge = None
+        self._disconnect_events: list[asyncio.Event] = []
 
         self._address_type = (
             kwargs["address_type"]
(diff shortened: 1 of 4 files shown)


bleak/backends/dotnet/client.py
```diff
@@ -168,79 +170,117 @@ async def connect(self, **kwargs) -> bool:
...
 
+        # Called on disconnect event or on failure to connect.
+        def handle_disconnect():
+            if self._connection_status_changed_token:
+                self._requester.remove_ConnectionStatusChanged(
+                    self._connection_status_changed_token
+                )
+                self._connection_status_changed_token = None
+
+            if self._requester:
```
hbldh, line 181: Here we (potentially) dispose of the requester before the session, which is opposite from how it is done in `disconnect`. Will that be a problem do you think?

hbldh (approved):

hbldh:
This is so good. I will add a commit changing event handling in ``discovery.py`` and ``scanner.py`` as well.

I was looking at `GattSession` a while back and didn't see the value of adding it then. I see it now. It was definitely a missing piece in the implementation.

Please add the changes to the `CHANGELOG` as well!

bleak/backends/dotnet/client.py
```diff
@@ -168,79 +170,117 @@ async def connect(self, **kwargs) -> bool:
...
-            ):
-                loop.call_soon_threadsafe(self._disconnected_callback, self)
-
-        self._requester.ConnectionStatusChanged += _ConnectionStatusChanged_Handler
+            loop.call_soon_threadsafe(
+                handle_connection_status_changed, sender.ConnectionStatus
+            )
 
-        # Obtain services, which also leads to connection being established.
-        services = await self.get_services()
```
hbldh, line 186: If `await self.get_services()` is not called in `connect`, it is required of the user to call it before trying to do any other operations. It should be done here to make sure that the `BleakClient` is ready for anything after the connection.

hbldh (changes requested):

bleak/backends/dotnet/client.py
dlech: Fixed. Thanks.

hbldh (approved):

PR #334, opened 2020-10-13
Standardizing arguments passed to callback functions

### LINKED ISSUE #301: Using a standard format for args passed to callback functions
* bleak version: v0.8.0
* Python version: v3.8.5
* Operating System: macOS Catalina v10.15.6
* BlueZ version (`bluetoothctl -v`) in case of Linux: N/A
* Apple Bluetooth Software Version: 7.0.6f7

### Description

Using the ```register_detection_callback()``` function to handle discovery of devices for a service, tested so far on macOS and Linux, raises the possible issue that the callback is sent different arguments depending on what backend of Bleak is being used:

#### CoreBluetooth
```
def callback(p, a, r):
    self._identifiers[p.identifier()] = a
    if self._callback:
        self._callback(p, a, r)
```

#### BluezDbus
```
if self._callback is not None:
    self._callback(message)
```

#### DotNet
```
if self._callback is not None:
      self._callback(sender, e)
```

Would it make sense in the bleak roadmap to use a standardized json or dictionary structure that would be passed to the callback function with all the pertinent data to avoid having to breakdown the ```*args``` depending on what platform is used to improve the platform-agnostic-ity of Bleak? If so, I'm happy to help implement the feature.

### What I Did
N/A

Bernstern:
Implementation for standardizing callback arguments passed to callback functions to make use easier across platforms (resolves #301).

This implements the feature for all platforms and include doc updates and an example.

### DIFF
--- a/bleak/backends/bluezdbus/scanner.py
+++ b/bleak/backends/bluezdbus/scanner.py
@@ -3,11 +3,10 @@ import asyncio
 import pathlib
 from typing import Callable, List
 
-from bleak.backends.scanner import BaseBleakScanner
+from bleak.backends.scanner import BaseBleakScanner, AdvertisementData
 from bleak.backends.device import BLEDevice
 from bleak.backends.bluezdbus import defs, get_reactor
 from bleak.backends.bluezdbus.utils import validate_mac_address
-
 from txdbus import client
 
 logger = logging.getLogger(__name__)
@@ -237,7 +236,7 @@ class BleakScannerBlueZDBus(BaseBleakScanner):
         stop_scanning_event = asyncio.Event()
         scanner = cls(timeout=timeout)
 
-        def stop_if_detected(message):
+        def stop_if_detected(advertisement_data: AdvertisementData):
             if any(
                 device.get("Address", "").lower() == device_identifier
                 for device in scanner._devices.values()
@@ -304,4 +303,38 @@ class BleakScannerBlueZDBus(BaseBleakScanner):
         )
 
         if self._callback is not None:
-            self._callback(message)
+            # First ensure that the message body is a list and has the dict where we can find the information we need
+            if isinstance(message.body, list) and len(message.body) > 2:
+                discovery_data = message.body[1]
+
+                # Get all the information wanted to pack in the advertisement data
+                _address = message.path[message.path.rfind("dev_") + 4 :].replace(
+                    "_", ":"
+                )
+
+                _local_name = discovery_data.get("Name", None)
+
+                _manufacturer_data = discovery_data.get("ManufacturerData", {})
+
+                _service_data = discovery_data.get("ServiceData", {})
+
+                _service_uuids = discovery_data.get("UUIDs", [])
+
+                _rssi = discovery_data.get("RSSI", 0)
+
+                # Pack the advertisement data
+                advertisement_data = AdvertisementData(
+                    address=_address,
+                    local_name=_local_name or "Unknown",
+                    rssi=_rssi,
+                    manufacturer_data=_manufacturer_data,
+                    service_data=_service_data,
+                    service_uuids=_service_uuids,
+                    platform_data=(message,),
+                )
+
+                self._callback(advertisement_data)
+
+            logger.info(
+                "Advertisement Data was an unexpected format, unable to safely trigger callback"
+            )
--- a/bleak/backends/bluezdbus/scanner.py
+++ b/bleak/backends/bluezdbus/scanner.py
@@ -36,7 +36,7 @@ def _filter_on_device(objs):
 
 def _device_info(path, props):
     try:
-        name = props.get("Name", props.get("Alias", path.split("/")[-1]))
+        name = props.get("Alias", "Unknown")
         address = props.get("Address", None)
         if address is None:
             try:
@@ -278,6 +278,43 @@ class BleakScannerBlueZDBus(BaseBleakScanner):
                 if msg_path in self._devices
                 else changed
             )
+
+            if self._callback is not None:
+                device = self._devices[msg_path]
+
+                # Get all the information wanted to pack in the advertisement data
+                _address = device["Address"]
+                _local_name = changed.get("Alias", device["Alias"])
+                # TODO: Look into BlueZ AdvertisementData property to get the local name
+                #  (see https://github.com/bluez/bluez/blob/master/doc/device-api.txt#L260)
+                _manufacturer_data = {
+                    k: bytes(v)
+                    for k, v in changed.get(
+                        "ManufacturerData", device.get("ManufacturerData", {})
+                    ).items()
+                }
+                _service_data = {
+                    k: bytes(v)
+                    for k, v in changed.get(
+                        "ServiceData", device.get("ServiceData", {})
+                    ).items()
+                }
+                _service_uuids = changed.get("UUIDs", device.get("UUIDs", []))
+                _rssi = changed.get("RSSI", device.get("RSSI", 0))
+
+                # Pack the advertisement data
+                advertisement_data = AdvertisementData(
+                    address=_address,
+                    local_name=_local_name,
+                    rssi=_rssi,
+                    manufacturer_data=_manufacturer_data,
+                    service_data=_service_data,
+                    service_uuids=_service_uuids,
+                    platform_data=(device, message)
+                )
+
+                self._callback(advertisement_data)
+
         elif (
             message.member == "InterfacesRemoved"
             and message.body[1][0] == defs.BATTERY_INTERFACE
@@ -301,36 +338,3 @@ class BleakScannerBlueZDBus(BaseBleakScanner):
                 *_device_info(msg_path, self._devices.get(msg_path))
             )
         )
-
-        if self._callback is not None:
-            discovery_data = message.body[1]
-
-            # Get all the information wanted to pack in the advertisement data
-            _address = discovery_data["Address"]
-
-            _local_name = discovery_data.get("Name", "Unknown")
-
-            _manufacturer_data = discovery_data.get("ManufacturerData", {})
-
-            _service_data = discovery_data.get("ServiceData", {})
-
-            _service_uuids = discovery_data.get("UUIDs", [])
-
-            _rssi = discovery_data.get("RSSI", 0)
-
-            # Pack the advertisement data
-            advertisement_data = AdvertisementData(
-                address=_address,
-                local_name=_local_name,
-                rssi=_rssi,
-                manufacturer_data=_manufacturer_data,
-                service_data=_service_data,
-                service_uuids=_service_uuids,
-                platform_data=(message,),
-            )
-
-            self._callback(advertisement_data)
-
-            logger.info(
-                "Advertisement Data was an unexpected format, unable to safely trigger callback"
-            )
--- a/bleak/backends/bluezdbus/scanner.py
+++ b/bleak/backends/bluezdbus/scanner.py
@@ -303,37 +303,33 @@ class BleakScannerBlueZDBus(BaseBleakScanner):
         )
 
         if self._callback is not None:
-            # First ensure that the message body is a list and has the dict where we can find the information we need
-            if isinstance(message.body, list) and len(message.body) > 2:
-                discovery_data = message.body[1]
+            discovery_data = message.body[1]
 
-                # Get all the information wanted to pack in the advertisement data
-                _address = message.path[message.path.rfind("dev_") + 4 :].replace(
-                    "_", ":"
-                )
+            # Get all the information wanted to pack in the advertisement data
+            _address = discovery_data["Address"]
 
-                _local_name = discovery_data.get("Name", None)
+            _local_name = discovery_data.get("Name", "Unknown")
 
-                _manufacturer_data = discovery_data.get("ManufacturerData", {})
+            _manufacturer_data = discovery_data.get("ManufacturerData", {})
 
-                _service_data = discovery_data.get("ServiceData", {})
+            _service_data = discovery_data.get("ServiceData", {})
 
-                _service_uuids = discovery_data.get("UUIDs", [])
+            _service_uuids = discovery_data.get("UUIDs", [])
 
-                _rssi = discovery_data.get("RSSI", 0)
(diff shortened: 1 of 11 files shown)


hbldh:
I like the idea, but I have not got the time to review this right now; it is a lot of changes...

@Bernstern Could you fix the conflicts and squash it down to one commit and I will look this over next week?

bleak/backends/dotnet/scanner.py
```diff
@@ -83,8 +85,50 @@ def AdvertisementWatcher_Received(self, sender, e):
...
+                    "Exception caught while parsing manufacturer data from discovery"
+                )
+                data = {}
+
+            try:
+                callback_data["address"] = e.BluetoothAddress or None
+                callback_data["name"] = e.Advertisement.LocalName or None
+                callback_data["data_channel"] = None
+                callback_data["manufacturer_data"] = data
+                callback_data["service_data"] = None
```
dlech, line 121: I think we should be able to get this with `e.Advertisement.GetSectionsByType(0x16)` plus some additional processing. And 0x20 and 0x21 are probably needed too to be complete.

dlech (changes requested):
Thanks for taking time to work on this. I've made a bunch of comments, but looking at this from a higher level, we are already parsing advertising data in 2 places (implementations of `BLEDevice` and `BaseBleakScanner` abstract classes) and this adds a 3rd. Reusing this existing code would fix many of the bugs I pointed out in this PR.

I think we could create an `AdvertisementData` class based on the dictionary given in this PR and utility functions in each backend to convert the platform-specific data into the common `AdvertisementData` type. Then this common type could be passed around instead of the platform-specific data.

bleak/backends/dotnet/scanner.py
```diff
@@ -1,10 +1,12 @@
+import inspect
```
dlech, line 1: Doesn't look like this is used anywhere.

Bernstern:
I agree, having it as a separate class will be really handy, I'll implement those changes (and remove that debug print :smiley: ) over the next few days.

Bernstern:
I implemented the requested changes to use an AdvertisementData class and updated the documentation and the example I wrote. I didn't quite follow @dlech's suggestion to include the other advertisement sections for the dotnet backend so I'm open to suggestions on how to implement that :smiley:.

bleak/backends/corebluetooth/scanner.py
```diff
@@ -46,10 +45,93 @@ async def start(self):
...
         def callback(p, a, r):
             self._identifiers[p.identifier()] = a
-            if self._callback:
-                self._callback(p, a, r)
+
+            # Get the defaults
+            callback_data = get_reference_callback_format()
+
+            # Get the raw data from the advertisement to be parse
+            service_data_dict_raw = a.get("kCBAdvDataServiceData", {})
```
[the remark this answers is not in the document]
Bernstern: Updated the service data to use the suggested dictionary comprehension and conversion closer.

bleak/backends/bluezdbus/scanner.py
```diff
@@ -304,4 +303,38 @@ def parse_msg(self, message):
         )
 
         if self._callback is not None:
-            self._callback(message)
+            # First ensure that the message body is a list and has the dict where we can find the information we need
+            if isinstance(message.body, list) and len(message.body) > 2:
```
dlech, line 307: When would this be false? Are there other kinds of messages that are received?

If it is not expected to ever be false, we should probably remove the check. Otherwise, it would be nice to invert the check so that the error is handled in the indented if statement and the normal code flow is not indented.

bleak/backends/bluezdbus/scanner.py
```diff
@@ -304,4 +303,38 @@ def parse_msg(self, message):
...
+                discovery_data = message.body[1]
+
+                # Get all the information wanted to pack in the advertisement data
+                _address = message.path[message.path.rfind("dev_") + 4 :].replace(
```
dlech, line 311: Why not use 

    _address = discovery_data["Address"]

instead of scraping the address from the D-Bus path?

bleak/backends/bluezdbus/scanner.py
```diff
@@ -304,4 +303,38 @@ def parse_msg(self, message):
...
+                    "_", ":"
+                )
+
+                _local_name = discovery_data.get("Name", None)
```
dlech, line 315: This could be 

    _local_name = discovery_data.get("Name", "Unknown")

instead of the special case for  `_local_name`  below

bleak/backends/corebluetooth/scanner.py
```diff
@@ -40,11 +39,65 @@ def __init__(self, **kwargs):
     async def start(self):
         self._identifiers = {}
 
+        def safe_list_get(l: list, idx: int, default):
```
dlech, line 42: It looks like this method is unused and can be removed.

bleak/backends/corebluetooth/scanner.py
```diff
@@ -40,11 +39,65 @@ def __init__(self, **kwargs):
...
+            """
+            Returns an index from a list safely similar to .get() with dicts
+            """
+            try:
+                return repr(l[idx]).lower()
+            except IndexError:
+                return default
+
+        def callback_dict_breakdown(data) -> Union[None, dict]:
```
dlech, line 51: This is also not used and can be removed.

bleak/backends/corebluetooth/scanner.py
```diff
@@ -40,11 +39,65 @@ def __init__(self, **kwargs):
...
+
+            # Process manufacturer data into a more friendly format
+            manufacturer_binary_data = a.get("kCBAdvDataManufacturerData")
+            manufacturer_data = {}
+            if manufacturer_binary_data:
+                manufacturer_id = int.from_bytes(
+                    manufacturer_binary_data[0:2], byteorder="little"
+                )
+                manufacturer_value = bytes(manufacturer_binary_data[2:])
+                manufacturer_data = {manufacturer_id: manufacturer_value}
```
dlech, line 86: This could modify the existing dictionary instead of creating a new one

    manufacturer_data[manufacturer_id] = manufacturer_value

bleak/backends/scanner.py
```diff
@@ -111,3 +111,34 @@ async def _find_device_by_address(
             await self.stop()
 
         return device
+
+
+class AdvertisementData:
+    """
+    Wrapper around the advertisement data that each platform returns upon discovery
+    """
```
dlech, line 119: Would be nice to list the allowable kwags in the docstring.

docs/scanning.rst
```diff
@@ -96,6 +96,10 @@ In the manual mode, it is possible to add an own callback that you want to call
 scanner detection, as can be seen above. There are also possibilities of adding scanning filters,
 which differ widely between OS backend implementations, so the instructions merit careful reading.
 
+Note that the when using :code:`BleakScanner.register_detection_callback(callback)`, your callback
+function will be passed a AdvertisementData class whose structure can be found in the base scanner file.
```
dlech, line 100: an :class:`AdvertismentData`

docs/scanning.rst
```diff
@@ -96,6 +96,10 @@ In the manual mode, it is possible to add an own callback that you want to call
...
+`See detection_callback.py` example. Note that platform specific advertisement data can be found in the platform_data field.
```
dlech, line 101: ``platform_data``

examples/detection_callback.py
```diff
@@ -0,0 +1,34 @@
...
+    print(callback_data)
+
+
+async def run():
+    scanner = BleakScanner()
+    scanner.register_detection_callback(simple_callback)
+
+    while True:
+        await scanner.start()
+        await asyncio.sleep(1)
```
dlech, line 29: one second seems a bit short for a demo

dlech (changes requested):
This looks good. Just a few more questions and suggestions.

We can look at reusing this to replace the duplicated code later.

bleak/backends/bluezdbus/scanner.py
Bernstern: I added the check to be safe with accessing the list in the next line, I haven't seen a case yet where it didn't pass so it should be safe to remove.

bleak/backends/scanner.py
Bernstern: Updated to include required and keyword args

Bernstern:
Let me know if there are any other changes you would like to see.

dlech:
I was hoping to have to have some time to test this before merging, but I haven't had time yet.

examples/detection_callback.py
```diff
@@ -0,0 +1,34 @@
...
+import asyncio
+from bleak import BleakScanner
+from bleak.backends.scanner import AdvertisementData
+import logging
+
+logging.basicConfig()
+
+
+def simple_callback(callback_data: AdvertisementData):
+    print(callback_data)
```
dlech, line 20: it would be nice to print all data here (e.g. the suggested `__repr__` method above)

bleak/backends/dotnet/scanner.py
```diff
@@ -89,9 +90,23 @@ def _received_handler(self, sender, event_args):
                 if event_args.BluetoothAddress not in self._devices:
                     self._devices[event_args.BluetoothAddress] = event_args
         if self._callback is not None:
-            self._callback(sender, event_args)
+            # Get a "BLEDevice" from parse_event args
+            temporary_device = self.parse_eventargs(event_args)
```
dlech, line 94: nitpick: `temporary_device` seems a bit verbose, `device` should do just fine

bleak/backends/dotnet/scanner.py
```diff
@@ -89,9 +90,23 @@ def _received_handler(self, sender, event_args):
...
+
+            # Use the BLEDevice to populate all the fields for the advertisement data to return
+            advertisement_data = AdvertisementData(
+                address=temporary_device.address,
+                local_name=temporary_device.name or "Unknown",
+                rssi=temporary_device.rssi,
+                manufacturer_data=temporary_device.metadata["manufacturer_data"],
+                service_data=event_args.Advertisement.GetSectionsByType(0x16),
```
dlech, line 102: The service data needs to be parsed to put it in the same format as other platforms. Here is what I came up with:

```python
            # Decode service data
            service_data = {}
            # 0x16 is service data with 16-bit UUID
            for section in event_args.Advertisement.GetSectionsByType(0x16):
                with BleakDataReader(section.Data) as reader:
                    data = reader.read()
                    service_data[
                        f"0000{data[1]:02x}{data[0]:02x}-0000-1000-8000-00805f9b34fb"
                    ] = data[2:]
            # 0x20 is service data with 32-bit UUID
            for section in event_args.Advertisement.GetSectionsByType(0x20):
                with BleakDataReader(section.Data) as reader:
                    data = reader.read()
                    service_data[
                        f"{data[3]:02x}{data[2]:02x}{data[1]:02x}{data[0]:02x}-0000-1000-8000-00805f9b34fb"
                    ] = data[4:]
            # 0x21 is service data with 128-bit UUID
            for section in event_args.Advertisement.GetSectionsByType(0x21):
                with BleakDataReader(section.Data) as reader:
                    data = reader.read()
                    service_data[str(UUID(bytes_le=data[:16]))] = data[16:]
```

bleak/backends/dotnet/scanner.py
```diff
@@ -89,9 +90,23 @@ def _received_handler(self, sender, event_args):
...
+                service_uuids=temporary_device.metadata["uuids"],
+                platform_data=(sender, event_args),
+            )
+
+            self._callback(advertisement_data)
 
-    def _stopped_handler(self, sender, event_args):
+    def _stopped_handler(self, sender, e):
```
dlech, line 109: `_` would be better than `e` if the argument is not used

bleak/backends/bluezdbus/scanner.py
```diff
@@ -304,4 +303,34 @@ def parse_msg(self, message):
         )
 
         if self._callback is not None:
-            self._callback(message)
+            discovery_data = message.body[1]
+
+            # Get all the information wanted to pack in the advertisement data
+            _address = discovery_data["Address"]
+
+            _local_name = discovery_data.get("Name", "Unknown")
```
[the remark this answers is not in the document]
dlech: Also should be using `"Alias"` instead of `"Name"`

dlech (changes requested):
Finally had some time to get back to this. It is looking really good.

Although I did find some problems on Windows and Linux - see inline comments.

FWIW, you can find my working branch at https://github.com/dlech/bleak/commits/pr/Bernstern/334 which contains the suggested fixes and some other unrelated fixes

dlech:
@hbldh do you have an opinion on where the `AdvertisementData` class should live? Currently it is at `bleak.backends.scanner.AdvertisementData`.

bleak/backends/bluezdbus/scanner.py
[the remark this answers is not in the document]
dlech: There is also a new experimental "AdvertisementData" property in BlueZ that we could use to get the local name instead of assuming the "Name" property came from the advertising data (which it may or may not have).

https://github.com/bluez/bluez/blob/master/doc/device-api.txt#L260

We don't have to implement this now, but we could add a TODO comment to remind us about it.

bleak/backends/bluezdbus/scanner.py
```diff
@@ -304,4 +303,34 @@ def parse_msg(self, message):
         )
 
         if self._callback is not None:
-            self._callback(message)
+            discovery_data = message.body[1]
```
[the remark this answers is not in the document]
Bernstern: Sgtm! I saw similar behavior where a bunch of callback messages were just the RSSI updates or similar, this is a really clean solution

bleak/backends/dotnet/scanner.py
Bernstern: Thank you for your help with this!

Bernstern:
@dlech Thank you very much for your suggestions and your working branch made it very apply your requested changes and test them locally, much appreciated! 

I added all of the changes on the working branch and merged develop back into it since there were merge conflicts but they should be resolved.

dlech:
Awesome. Let's call this good enough and get it merged.

PR #336, opened 2020-10-14
Adding a timeout on OSX so that the connect cannot hang forever.

cspensky:
I have seen connect get itself in a state where will loop indefinitely.  This timeout fix seems to fix it.  `time.time()` seems like a hack, but I cannot think of a better way offhand.  Open to suggestions.

### DIFF
--- a/bleak/backends/corebluetooth/CentralManagerDelegate.py
+++ b/bleak/backends/corebluetooth/CentralManagerDelegate.py
@@ -145,15 +145,17 @@ class CentralManagerDelegate(NSObject):
 
     async def connect_(self, peripheral: CBPeripheral, timeout=10.0) -> bool:
         self._connection_state = CMDConnectionState.PENDING
+        self._connection_state_changed.clear()
         self.central_manager.connectPeripheral_options_(peripheral, None)
 
         try:
             await asyncio.wait_for(
                 self._connection_state_changed.wait(), timeout=timeout
             )
-        except TimeoutError:
+        except asyncio.TimeoutError:
             logger.debug(f"Connection timed out after {timeout} seconds.")
-            return False
+            self.central_manager.cancelPeripheralConnection_(peripheral)
+            raise
 
         self.connected_peripheral = peripheral
 
--- a/bleak/backends/corebluetooth/CentralManagerDelegate.py
+++ b/bleak/backends/corebluetooth/CentralManagerDelegate.py
@@ -77,6 +77,7 @@ class CentralManagerDelegate(NSObject):
 
         self.callbacks = {}
         self.disconnected_callback = None
+        self._connection_state_changed = asyncio.Event()
 
         self.central_manager = CBCentralManager.alloc().initWithDelegate_queue_(
             self, dispatch_queue_create(b"bleak.corebluetooth", DISPATCH_QUEUE_SERIAL)
@@ -146,11 +147,11 @@ class CentralManagerDelegate(NSObject):
         self._connection_state = CMDConnectionState.PENDING
         self.central_manager.connectPeripheral_options_(peripheral, None)
 
-        start = time.time()
-        while self._connection_state == CMDConnectionState.PENDING:
-            await asyncio.sleep(0)
-            if time.time() - start > timeout:
-                raise TimeoutError("Failed to connect in %d seconds" % timeout)
+        try:
+            await asyncio.wait_for(self._connection_state_changed.wait(), timeout=timeout)
+        except TimeoutError:
+            logger.debug(f"Connection timed out after {timeout} seconds.")
+            return False
 
         self.connected_peripheral = peripheral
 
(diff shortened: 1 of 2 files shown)


Carglglz:
```await asyncio.wait_for(client.connect(timeout=3), timeout=20))``` was my solution to this, first timeout is for discovery, second timeout for connection, in the "rare" case where it will loop indefinitely. (Usually happens if switching off or placing the device out of range).

Here the docs : https://docs.python.org/3/library/asyncio-task.html#timeouts

Is it maybe a better approach ?

**Comment by dlech** [comment]
According to the CoreBluetooth docs, connections don't time out, so we need to implement our own. Your suggestion is fine for the way the connection method is currently implemented.

But I think we could fix this up in a more asyncio kind of way though (similar to https://github.com/hbldh/bleak/pull/328/commits/93748618bbeaf35a9823faf810de7753ad046eb9).

* Use an `asyncio.Event` instead of polling for `self._connection_state == CMDConnectionState.PENDING`.

  * `CentralManagerDelegate` will need to maintain a map of peripheral ids to events, then `did_connect_peripheral()` and `did_fail_to_connect_peripheral` need to be modified to call `Event.set()` instead of using `self._connection_state`.
  * `await asyncio.wait_for(event.wait(), timeout=timeout)` can be used to implement the timeout instead of the while loop.

* `self.central_manager.cancelPeripheralConnection_(self.connected_peripheral)` needs to be called in the event of a timeout to cancel the pending connection.

File: bleak/backends/corebluetooth/CentralManagerDelegate.py [hunk 1]
```diff
@@ -146,11 +147,11 @@ async def connect_(self, peripheral: CBPeripheral, timeout=10.0) -> bool:
         self._connection_state = CMDConnectionState.PENDING
         self.central_manager.connectPeripheral_options_(peripheral, None)
 
-        start = time.time()
-        while self._connection_state == CMDConnectionState.PENDING:
-            await asyncio.sleep(0)
-            if time.time() - start > timeout:
-                raise TimeoutError("Failed to connect in %d seconds" % timeout)
+        try:
```
dlech, line 150: I think we need to call `self._connection_state_changed.clear()` here in case connect is called more than once.

bleak/backends/corebluetooth/CentralManagerDelegate.py
```diff
@@ -146,11 +147,11 @@ async def connect_(self, peripheral: CBPeripheral, timeout=10.0) -> bool:
...
+            await asyncio.wait_for(self._connection_state_changed.wait(), timeout=timeout)
+        except TimeoutError:
```
dlech, line 154: This needs to be `asyncio.TimeoutError` rather than the builtin `TimeoutError`.

bleak/backends/corebluetooth/CentralManagerDelegate.py
```diff
@@ -146,11 +147,11 @@ async def connect_(self, peripheral: CBPeripheral, timeout=10.0) -> bool:
...
+            logger.debug(f"Connection timed out after {timeout} seconds.")
```
dlech, line 155: We need to call `self.central_manager.cancelPeripheralConnection_(self.connected_peripheral)` here to cancel the pending connection.

bleak/backends/corebluetooth/CentralManagerDelegate.py
```diff
@@ -146,11 +147,11 @@ async def connect_(self, peripheral: CBPeripheral, timeout=10.0) -> bool:
...
+            return False
```
dlech, line 156: I think it would be better to propagate the exception here by calling `raise` instead of returning `False`. This way callers can know why the connection failed without having to parse the debug log.

dlech (changes requested):
Thanks for updating. I have suggested a few further changes.

bleak/backends/corebluetooth/CentralManagerDelegate.py
cspensky: Great idea.

bleak/backends/corebluetooth/CentralManagerDelegate.py
cspensky: Great catch!

bleak/backends/corebluetooth/CentralManagerDelegate.py
cspensky: This explains why, in rare cases, I was seeing a bunch of "Successful" connections all in a row.  Thanks for the heads up.

bleak/backends/corebluetooth/CentralManagerDelegate.py
cspensky: This makes sense to me.

bleak/backends/corebluetooth/CentralManagerDelegate.py
```diff
@@ -9,6 +9,7 @@
 import asyncio
 import logging
 import platform
+import time
```
dlech, line 12: I missed this before - we no longer need to `import time`.

dlech (approved):
Thanks!

hbldh (approved):

PR #339, opened 2020-10-20
Fix nordic uart UUID

Carglglz:
Nordic RX / TX uuids are the other way around 🔃 

https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/include/bluetooth/services/nus.html


hbldh (approved):

hbldh:
Very good. I will take your word for it. Thanks!

PR #343, opened 2020-10-21
Fixes #329: issue causing advertising data to be lost on successive calls to scanner callback

### LINKED ISSUE #329: Advertised UUIDs are erased on subsequent calls to CentralDelegateManager scanner callback
* bleak version: 0.8.0
* Python version: 3.8.5
* Operating System: macOS 10.15.7
* BlueZ version (`bluetoothctl -v`) in case of Linux: N/A

### Description

When calling `BleakScanner.discover()` on macOS, devices which advertise service UUIDs are returned with `device.metadata["uuids"]` empty. For each of these devices, multiple calls are made to `CentralManagerDelegate.did_discover_peripheral()`. Each call to this function calls the following function, `callback()`, defined inside `BleakScannerCoreBluetooth.start()`:

```python
     def callback(p, a, r):
            self._identifiers[p.identifier()] = a
            if self._callback:
                self._callback(p, a, r)
```

If `callback()` is called multiple times, `self._identifiers[p.identifier()]` will only contain the advertising data `a` passed in on the final call.

### Proposed Solution

I was able to fix the issue for my use case by changing `callback()` to the following:

```python
       def callback(p, a, r):
            if p.identifier() in self._identifiers:
                self._identifiers[p.identifier()] = {**a, **self._identifiers[p.identifier()]}
            else:
                self._identifiers[p.identifier()] = a
            if self._callback:
                self._callback(p, a, r)
```

This change merges the advertising data objects according to the method described in [this stackoverflow post](https://stackoverflow.com/a/26853961). However, I am not familiar enough with this library to know if this could cause other adverse effects. Please let me know if this solution makes sense, or how I should test it.

adamincera:


dlech (approved):
Works for me, thanks!

PR #344, opened 2020-10-21
corebluetooth: use cb_uuid_to_str() to convert kCBAdvDataServiceUUIDs

### LINKED ISSUE #342: corebluetooth returning "uuids" that are service names, like "Heart Rate" or "Device Information"
* bleak version: 0.8.0 and tip of develop
* Python version: 3.8.2
* Operating System: macOS 10.15.7

I was startled to discover that corebluetooth appears to be returning English language strings for standard Bluetooth UUIDS, such as "Heart Rate".

In https://github.com/hbldh/bleak/blob/ccdac2388e882a5e5c446cf36df349fbfedce2e8/bleak/backends/corebluetooth/device.py#L45-L47
`cbuuids` can contain human-readable strings instead of strings that are UUIDs.

I was looking at the advertising data from a standard Heart Rate Monitor device (Scosche), and `cbuuids` contains this:
```python
(
    "Heart Rate",
    "Device Information",
    1725
)
```
bleak lowercases these, on the assumption they are uuids, but these are just readable strings.

So any attempt to determine if the advertised service is the Heart Rate Monitor service (0x180D) has to do a string comparison against "heart rate", instead of "0x180d".

I don't think bleak is doing this conversion, but it's coming directly from `kCBAdvDataServiceUUIDs`. I'm thinking `uuids.py` needs to contain a reverse dictionary to convert these strings back to hex for uniformity across platforms.

Am I missing something?? Has someone else encountered this? As usual Apple's documentation seems to be lacking details.

dlech:
This was missed in 6a0d8c99f5c82b6b1c000316fb1c0dff49a92138.

Also, since this was just using `str()` directly on the `CBUUID` instance, it had an additional bug for well known services since `str()` would return the name of the service instead of a UUID.

Fixes #342.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,17 @@ All notable changes to this project will be documented in this file.
 The format is based on `Keep a Changelog <https://keepachangelog.com/en/1.0.0/>`_,
 and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0.html>`_.
 
+
+`Unreleased`_
+--------------
+
+Fixed
+~~~~~
+
+* Fix well-known services not converted to UUIDs in ``BLEDevice.metadata`` in
+  CoreBluetooth backend. Fixes #342.
+
+
 `0.9.0`_ (2020-10-20)
 ---------------------
 
--- a/bleak/backends/corebluetooth/device.py
+++ b/bleak/backends/corebluetooth/device.py
@@ -3,7 +3,7 @@ from bleak.backends._manufacturers import MANUFACTURERS
 
 from Foundation import NSDictionary
 
-
+from bleak.backends.corebluetooth.utils import cb_uuid_to_str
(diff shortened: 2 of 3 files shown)


dlech:
Hi @dhalbert, could you please test?

dhalbert (approved):
Tested with a Schosche heart rate monitor. This now works just fine. Thank you!

Also tested with some other peripherals that were already working: no regressions.

@hbldh Would you want to do a 0.9.1 release soon, since this is a regression? Perhaps also include #343?

hbldh:
@dhalbert that was idea as well. This should be patched quickly.

dlech:
Thanks for testing!

hbldh:
@dhalbert This is now included in the newly released version 0.9.1 on PyPI.

dhalbert:
Thanks! I'm writing up a guide to our library that uses bleak, and I can update our library `requirements.txt` now.

PR #346, opened 2020-10-21
Add device info bluezdbus client.py

Carglglz:
I may be wrong, but currently there is no way to retrieve device information (name, RSSI...) in Bluezdbus client.
This adds `client._device_info`, similar to CoreBluetooth client.


hbldh:
Is this really useful information? The information there that is interesting, e.g. RSSI, is not updated, so it will be just device name and  suchlikes. In my world, that is information you don't need after connection, but that is just me.

Try to make PRs from pristine branches so that only one commit exists. This I will have to squash and merge and I like to avoid doing that.

Carglglz:
> Try to make PRs from pristine branches so that only one commit exists. This I will have to squash and merge and I like to avoid doing that.

Sorry a I didn't realise of this mess yesterday 😞, I was about to close this and make a proper PR.

> Is this really useful information? The information there that is interesting, e.g. RSSI, is not updated, so it will be just device name and suchlikes. In my world, that is information you don't need after connection, but that is just me.

I do use the Name and suchlike to set further configuration, and since this was already in CoreBluetooth client + being that the change was only 3 lines I though this would be harmless. 

So thanks! 🙏

PR #347, opened 2020-10-22
macOS fixes

dlech:
First commit fixes possible race condition when using `CBCentralManager` in rare cases. (I was also getting a `asyncio.TimeoutError` from `wait_for_powered_on()` when running using a debugger that is fixed by this.)

Second commit fixes a regression from 583c08e4bf1b.

### DIFF
--- a/bleak/backends/corebluetooth/client.py
+++ b/bleak/backends/corebluetooth/client.py
@@ -48,9 +48,7 @@ class BleakClientCoreBluetooth(BaseBleakClient):
 
         if isinstance(address_or_ble_device, BLEDevice):
             self._device_info = address_or_ble_device.details
-            self._central_manager_delegate = address_or_ble_device.metadata.get(
-                "delegate"
-            )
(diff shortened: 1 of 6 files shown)


bleak/backends/corebluetooth/client.py
```diff
@@ -48,9 +48,7 @@ def __init__(self, address_or_ble_device: Union[BLEDevice, str], **kwargs):
 
         if isinstance(address_or_ble_device, BLEDevice):
             self._device_info = address_or_ble_device.details
-            self._central_manager_delegate = address_or_ble_device.metadata.get(
```
hbldh, line 51: I generally always use the `get` method for dict value retrieval, but here the `delegate` key is strictly needed for the functionality is it not?

hbldh (approved):

bleak/backends/corebluetooth/client.py
dlech: Correct. That is why I changed it here.

PR #378, opened 2020-12-07
fix dotnet event callbacks

dlech:
Windows .NET event handler callbacks run on a background thread, so we need to use `call_soon_threadsafe()` to schedule the callbacks in the asyncio event loop. See commit messages for more details.


hbldh (approved):
Very good. I seem to recall that I did the add/remove changes in scanning and it seems that I disregarded the threading here. Good changes.

PR #387, opened 2020-12-08
Removed separate implementation of global discover method. 

dlech:
The "old" `discover()` method was largely duplicating the code from `BleackScanner.discover()`, with the exception of the Mac backend which looked like it might not even work correctly and ignored kwargs.

Rather than trying to maintain two separate implementations that are supposed to dot he same thing, this removes the `discover()` method from the documentation and removes the separate implementation. The name is still kept and made as a synonym for for `BleakScanner.discover` so that we don't completely break existing code that might be using it.


hbldh:
Yes. I meant to do this when I implemented the scanner, but I decided to wait until I was sure it was stable. Then I forgot it. Well done!

PR #390, opened 2020-12-10
Detection callback improvments

dlech:
After using the changes from #334 a bit more, I found myself wanting a `BLEDevice` in addition to the `AdvertismentData`. This way if out find a device you like, you can pass the `BLEDevice` to `BleakClient()` directly without having to find it again in the list returned by `get_discovered_devices()`.

Technically, some things like the device identifier (address or UUID on Mac) and RSSI aren't part of the advertising data and were already in `BLEDevice` so I removed them from `AdvertisingData`. This also opened the opportunity to clean up some other things along the way.

@Bernstern, does this work for you?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,7 @@ Added
 * Added AdvertisementData class used with detection callbacks across all supported platforms. Merge #334
 * Added BleakError raised during import on unsupported platforms.
 * Added rssi parameter to BLEDevice constructor.
+* Added detection_callback kwarg to BleakScanner constructor.
 
 Changed
 ~~~~~~~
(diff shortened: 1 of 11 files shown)


Bernstern:
This looks great to me @dlech! In our project we have had to "reconstruct" a `BLEDevice` object in the callback function to get around using `get_discovered_devices()` again. I think this is a really clever approach. 

I can dig into it tomorrow and test it locally, at least on Linux and OSX, this looks to clean things up very nicely!

dlech:
Good to know it is solving a real problem.

Bernstern:
After playing around with it on Linux and osx I am really happy, having the `BLEDevice` object returned cleaned up our logic a lot. Didn't take many changes at all to accommodate.

All in all this looks great to me and has been working as expected on both platforms.

dlech:
Thanks for testing!

hbldh:
I have also tested in on my Ubuntu system tonight and feel satisfied with it. @dlech will you do the honours of making a 0.10.0 release of these much needed fixes and improvements that have been made the last couple of days?

PR #400, opened 2020-12-21
Fix rssi values in BLE scanner

Carglglz:
In latest bleak version 0.10.0 the RSSI values of devices returned by `get_discovered_devices` are all zero.
This comes from `BLEDevice` class which has a `rssi=0` keyword argument default to zero. 
This PR updates this value with the real `RSSI` value.

@dlech I don't know if this is the best way to solve this, so feel free to change anything before merging 👍 

PD: Also for some reason bleak version in develop branch is still `0.9.1`, I guess this should be updated as well.

### DIFF
--- a/bleak/__version__.py
+++ b/bleak/__version__.py
@@ -1,3 +1,3 @@
 # -*- coding: utf-8 -*-
 
-__version__ = "0.9.1"
+__version__ = "0.10.0"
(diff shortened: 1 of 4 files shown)


dlech:
I think this is OK for now. 

I noticed before that this seemed to be missing but I couldn't think of a better way to do it either. There are some other functions for getting RSSI for a `CBPeriperal`, but one is deprecated and the other is async and requires actually communicating with the device.

And looking at the bigger picture, macOS is the only backend that has has a special OS function to get discovered devices. In the other backends, Bleak just manually collects the discovered devices. So in the future, it probably makes sense to just implement a common collector that can be used by all backends.

PR #401, opened 2020-12-22
bluez: resolve services once again after reconnecting

mniestroj:
So far disconnect() method has cleared all services collection, but it
didn't reset client._services_resolved flag. This means that services
were not resolved after reconnecting to the same client.

Just clear client._services_resolved, which will allow to resolve
services properly with each subsequent reconnect.


dlech:
Great, thanks!

PR #410, opened 2021-01-04
Adding error checking to properly catch and handle edge cases

cspensky:
Just a few error conditions that came up in my testing that I cleaned up.

### DIFF
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -513,6 +513,9 @@ class BleakClientBlueZDBus(BaseBleakClient):
             (bytearray) The read data.
 
         """
+        if self._bus is None:
+            raise BleakError("D-Bus object is None")
+
         if not isinstance(char_specifier, BleakGATTCharacteristicBlueZDBus):
             characteristic = self.services.get_characteristic(char_specifier)
         else:
--- a/bleak/backends/bluezdbus/scanner.py
+++ b/bleak/backends/bluezdbus/scanner.py
@@ -242,7 +242,10 @@ class BleakScannerBlueZDBus(BaseBleakScanner):
             # discovered devices list
 
             msg_path = message.body[0]
-            del self._devices[msg_path]
+            try:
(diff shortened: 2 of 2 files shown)


bleak/backends/bluezdbus/client.py
```diff
@@ -513,6 +513,9 @@ async def read_gatt_char(
             (bytearray) The read data.
 
         """
+        if self._bus is None:
+            raise BleakError("D-Bus object is None")
```
dlech, line 517: How about saying that the client is not connected instead? Isn't that what triggers this problem?

dlech (commented):
Thanks for this. You should also try out the `dbus-next-2` branch as it will probably replace the BlueZ backend soon.

I already fixed the last two issues in that branch.

bleak/backends/bluezdbus/scanner.py
```diff
@@ -242,7 +242,10 @@ def _parse_msg(self, message: Message):
             # discovered devices list
 
             msg_path = message.body[0]
-            del self._devices[msg_path]
+            try:
```
dlech, line 245: FWIW, here is how I fixed it: https://github.com/hbldh/bleak/pull/389/commits/5bf7266825c7b4714b90e59200739bfc8d014018

dlech: But nothing wrong with this way either :smile:

bleak/backends/bluezdbus/client.py
cspensky: Good point.  Let me look into this one more and clean up the error message.

bleak/backends/bluezdbus/scanner.py
cspensky: 👍 I like it.  These were a bit old, I just wanted to document them since they were all short and sweet.

cspensky:
Closing this.  I'll submit another pull request with some cleaned up error messages now that our branches are synced back up.

PR #420, opened 2021-01-11
adv_service_data

jsotogaviard:

   guidelines in that document.
    - OS X / macOS >= 10.11

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,11 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 `Unreleased`_
 -------------
 
+Added
+~~~~~
+
+* Added ``AdvertisementServiceData`` in BLEDevice.
+
 Fixed
 ~~~~~
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,7 +13,7 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 Added
 ~~~~~
 
-* Added ``AdvertisementServiceData`` in BLEDevice.
+* Added ``AdvertisementServiceData`` in BLEDevice in macOS devices
 
 Fixed
 ~~~~~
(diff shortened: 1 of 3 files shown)


bleak/backends/corebluetooth/scanner.py
```diff
@@ -132,6 +132,11 @@ async def get_discovered_devices(self) -> List[BLEDevice]:
                 for u in advertisementData.get("kCBAdvDataServiceUUIDs", [])
             ]
 
+            service_data = {}
+            adv_service_data = advertisementData.get("kCBAdvDataServiceData", [])
+            for u in adv_service_data:
+                service_data[cb_uuid_to_str(u)]= bytes(adv_service_data[u])
```
dlech, line 138: please add a space before the equals to fix the format test failure

dlech (changes requested):
Thanks! I'm not sure how we missed this before!

CHANGELOG.rst
```diff
@@ -10,6 +10,11 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 `Unreleased`_
 -------------
 
+Added
+~~~~~
+
+* Added ``AdvertisementServiceData`` in BLEDevice.
```
dlech, line 16: Please add something to say that this is just for macOS here.

bleak/backends/corebluetooth/scanner.py
jsotogaviard: done

CHANGELOG.rst
jsotogaviard: done

dlech (approved):

dlech:
Thanks for updating. Is it safe to assume you have tested this with a device that provides service data in the advertising data?

jsotogaviard:
I have tested with xiaomi mi thermometers and atc fw : https://github.com/atc1441/ATC_MiThermometer

hbldh:
@jsotogaviard If you could use your device as an example to try to implement this for the backend missing this data (the BlueZ/Linux backend I think), that would be great!

Regardless, thank you for the contribution!

jsotogaviard:
I only Have Mac/windows computers ...

dlech:
>  If you could use your device as an example to try to implement this for the backend missing this data (the BlueZ/Linux backend I think), that would be great!

This was actually implemented in #334 and I tested all backends then. But somehow we managed to delete this from macOS in the meantime.

Issue #405, opened 2020-12-30
ProtocolError (Error: 0x0F: Connection Rejected due to Unacceptable BD_ADDR)


BioMycoBit:
* bleak version: 0.10.0
* Python version: 3.9
* Operating System: Windows 10
* BlueZ version (`bluetoothctl -v`) in case of Linux: PyBluez 0.30

### client.write_gatt_char(on_uuid, [0x1], True)

Attempting to control Philips Hue lights via Bluetooth Low Energy.  I am able to successfully connect to the client and read the services and characteristics but **receive BD_ADDR errors for many of the gatt characteristics**.

### Example Error
[Characteristic] 97fe6561-0003-4f62-86e9-b71ee2da3d22: (Handle: 20) (read,write-without-response,write) | Name: Unknown, Value: _**b'Could not get GATT characteristics for 97fe6561-0003-4f62-86e9-b71ee2da3d22: ProtocolError (Error: 0x0F: Connection Rejected due to Unacceptable BD_ADDR)'**_

### Exploration

Very few references online in regards to Unacceptable BD_ADDR.  

BLE Connection failure : Connection is refused due to unacceptable BD_ADDR
https://stackoverflow.com/questions/64985001/ble-connection-failure-connection-is-refused-due-to-unacceptable-bd-addr

This example suggested that the end user was utilizing Bluetooth Classic and not BLE.  Didn't think that'd be possible with Bleak...

### Code
```p
import asyncio
import logging
from bleak import BleakClient

async def run(address, debug=False):
    log = logging.getLogger(__name__)
    if debug:
        import sys

        log.setLevel(logging.DEBUG)
        h = logging.StreamHandler(sys.stdout)
        h.setLevel(logging.DEBUG)
        log.addHandler(h)

    async with BleakClient(address) as client:
        x = await client.is_connected()
        log.info(f"{address} Connected: {x}")

        name_uuid = '00002a00-' + '0000' +'-1000-8000-00805f9b34fb'
        service_uuid = '932c32bd-' + '0000' + '-47a2-835a-a8d455b859dd'
        on_uuid = '932c32bd-' + '0002' + '-47a2-835a-a8d455b859dd'

        name = await client.read_gatt_char(name_uuid)
        service = client.services.get_service(service_uuid)
        characteristic = service.get_characteristic(on_uuid)

        log.info(f"Name: {name}")
        log.info(f"Service: {service}")
        log.info(f"Characteristic: {characteristic}")

        # on
[...]

PR #421, opened 2021-01-13
Pairing improvement for .NET backend

hbldh:
Added protection levels when trying to pair with devices in Windows backend. It works with ProtectionLevel = 2 (Encrypted) at least, which solves some problems, e.g. #405.

Also wrote a sample script for connecting to and controlling Philips Hue lamps, albeit only tested in Windows.

### DIFF
--- a/bleak/backends/dotnet/client.py
+++ b/bleak/backends/dotnet/client.py
@@ -327,8 +327,15 @@ class BleakClientDotNet(BaseBleakClient):
             )
 
             if protection_level:
-                raise NotImplementedError(
-                    "Cannot set minimally required protection level yet..."
+                pairing_result = await wrap_IAsyncOperation(
+                    IAsyncOperation[DevicePairingResult](
+                        custom_pairing.PairAsync.Overloads[DevicePairingKinds, DevicePairingProtectionLevel]
+                        (
+                            ceremony,
+                            protection_level
+                        )
+                    ),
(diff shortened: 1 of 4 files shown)


bleak/backends/dotnet/client.py
```diff
@@ -333,13 +345,7 @@ def handler(sender, args):
...
                 )
 
-            try:
-                custom_pairing.PairingRequested -= handler
-            except Exception:
-                # TODO: Find a way to remove WinRT events...
-                pass
-            finally:
-                del handler
+            custom_pairing.remove_PairingRequested(pairing_requested_token)
```
dlech, line 348: Should this be in a `finally:` to ensure that it is always removed even if an exception is raised between the time the event is added and here?

hbldh: Good idea. Do we have try and finally for other deregistrations of handlers? Or are they all safely removed on disconnect?

dlech: I think most are handled in disconnect rather than short term like this.

dlech: and I think we have something to remove the event handlers if the connection fails, but it wouldn't hurt to double-check

hbldh: Enclosed the removal in a finally clause just now.

dlech (approved):

PR #422, opened 2021-01-14
python-for-android backend

xloem:
This is my past week or so of work on a [python-for-android](https://python-for-android.readthedocs.io/en/latest/) backend.  Python-for-android is primarily used by [kivy](https://kivy.org/), a python gui toolkit.

This code is working for me, but I haven't tested features my test device doesn't use, such as pairing.  I think this code might be valuable enough to merge in, maybe with a comment to report to me any bugs encountered, with logcat output, or somesuch.

If this isn't a good addition to this project, the python-for-android project might consider it as a pull request, too, as part of their android interface library.

I'd appreciate criticism on my implementation choices.  Most of my energy was spent just making it work.


hbldh:
Good work! This is interesting, but it needs some additional handling first. It needs some documentation and some clear descriptions on how to use. 

Other questions:
- Is it only for `python-for-android`? 
- Does it (or at least, can it) work with [Beeware's Android solution](https://docs.beeware.org/en/latest/index.html) as well?
- Does it only work with Kivy? Are there other ways of running it on Android?

xloem:
- It uses `pyjnius` which I've mostly seen used for `python-for-android`. I recently (after coding this up) discovered the proprietary android app `pydroid 3` which does on-mobile python development and pyjnius/android also works.  I haven't tested this PR with pydroid 3.
- I haven't implemented beeware, as I'm less familiar with it and it uses different interfacing libraries for the android parts.  Due to https://github.com/beeware/rubicon-java/issues/25 the solution would presently require making a custom beeware template that includes the auxiliary java classes, and requiring apps build with that template.  My thought was to polish up p4a first, and then add beeware some other time.
- I believe it is possible to use python-for-android without kivy or pydroid3.  I haven't done this myself: you'd have to build or connect with a UI, for the user to see the application.

Is it pretty obvious what docs need updating here?

hbldh:
Thanks for the answers; I would very much like to make Bleak compatible with Beeware solutions eventually, so I wanted to hear your thoughts on that!

The docs that need updating are [here](https://github.com/hbldh/bleak/tree/develop/docs/backends) and then adding api-references [here](https://github.com/hbldh/bleak/blob/develop/docs/api.rst). In combination with the example and README that you have already written, could you please include some links or references on where to start reading about how to get python-for-android onto the phone?

Also, `pyjnius` needs to be added to `requirements.txt` and to `setup.py` so that they will be installed if it is an android system, right?

xloem:
@hbldh I threw something together for docs.  The p4a backend uses the `pyjnius` and `android` packages, but python-for-android already installs both of these itself, and they are not needed on other platforms.  python-for-android reads as 'linux' in sys.platform so it would be an extra manual check for environment variables in setup.py ...  still something valuable?

hbldh:
@xloem Sorry for taking time to merge this. I am interested in adding this, but there are several other things needing triage and subsequent release. I am planning on making a 0.11.0 release with a lot of potentially problematic code, and I want to wait with this until after that release to merge this to develop. Just so you know.

jeffsf:
I have only glanced at the code at this point, so I can't comment on its technical correctness. 

Having ~~worked~~ struggled with BLE on the Android stack, one thing that is clear is that it is not robust and not "asynchronous" in the way one might hope. My own explorations have aligned well those reported elsewhere that the application layer needs to serially send commands and check for completion before sending another command. At least for that application, I needed to manage matching ACKs with requests explicitly to prevent BLE failures.

Here are some references I found (I was unable to confirm if it is one per CUUID, device, or stack before abandoning the third-party code base I was working with.)

> Android seems to have a limit of one pending request
> If we're lucky, it is per device
> 
> (2016) https://www.toothpasteandbubblegum.com/blog/queued-data-bluetooth-le-android
> (2017) https://stackoverflow.com/questions/18011816/is-the-native-android-ble-implementation-synchronous-in-nature
> (2016) https://github.com/NordicPlayground/puck-central-android
> (2020) https://punchthrough.com/android-ble-guide/
> * Reaffirms that everything needs to be done serially
> * Confirms that failures need to be handled externally
> 
> Multiple references at and linked from the recently updated https://www.stkent.com/2017/09/18/ble-on-android.html_
> 
> 
>

hbldh:
@jeffsf Thank you for your comments. I will take this under consideration; I have not tested this Android solution myself yet and need to confirm that it does what it needs to be included in Bleak.

rocketstrong600:
i have been testing his commits and they seem to be working well i was able to write a app that talks over nordic ble uart

dlech:
Is there a way to get this up and running in 15 minutes or less so we can test it easily? I've looked a few times but I don't see a way to just run a simple command line python file on Android.

xloem:
I'm so sorry, I don't remember, you'd have to ask in python4android communities.  But I think it's pretty reliable to delete the buildozer build folder, and I think there are some specific subfolders inside it you can delete instead to speed the rebuild up.  I'm not sure what they are, I'm afraid.

xloem:
Thanks so much for your review and work.

I don't actually remember this code very much at all, so when I respond to the review I'm mostly re-reviewing it myself.

Do you guys need me to find a test device again and address all the concerns you identified?  Many of them look simple and valid but I should test if I implement the changes.

xloem:
had a marathon today of addressing most of these things.  haven't figured out the recipe building thing yet at this time.

dlech:
Merged! Thanks again for all of the amazing work!

dlech:
I also forgot to say that you should feel free to add yourself to the AUTHORS file if you like.

PR #430, opened 2021-01-22
Fixing set_scanning_filter

cspensky:
`set_scanning_filter` does not work in the current implementation.  This fix implements all of the proper varients for BlueZ 5.48.  There are more on the master branch for BlueZ, but I only implemented those in the docstring.


dlech:
Nice, thanks!

Issue #429, opened 2021-01-22
Services & Characteristics dictionary is retaining the old values.


chinmayjoshi1992:
System Using:

Windows 10
python 3.7
bleak ver : 0.10.0

I've a device which has 10-15 services.

When I send a get_services() request it only gives me services which I get in 5 seconds after that connection losts.
Is there a way to get all the services in one go?
```
2021-01-22 15:55:44,225 bleak.backends.dotnet.scanner DEBUG: Received 80:E1:26:xx:xx:xx: 
2021-01-22 15:55:44,233 bleak.backends.dotnet.scanner DEBUG: Received  80:E1:26:xx:xx:xx:  Unknown.
2021-01-22 15:55:44,233 bleak.backends.dotnet.client DEBUG: Connecting to BLE device @  80:E1:26:xx:xx:xx: 
2021-01-22 15:55:44,510 bleak.backends.dotnet.client DEBUG: _ConnectionStatusChanged_Handler: 1
2021-01-22 15:55:44,510 bleak.backends.dotnet.client DEBUG: Get Services...
2021-01-22 15:55:44,800 bleak.backends.dotnet.client INFO: Services resolved for BleakClientDotNet ( 80:E1:26:xx:xx:xx: )
Services:
xxxxxxxx-xxxx-xxxx-xxx-00805f9b34fb: Generic Attribute Profile
xxxxxxxx-xxxx-xxxx-xxx-0242ac120002: Unknown
00005005-0000-1000-8000-00805f9b34fb: Vendor specific
2021-01-22 15:55:44,800 bleak.backends.dotnet.client DEBUG: Disconnecting from BLE device...
2021-01-22 15:55:47,848 bleak.backends.dotnet.client DEBUG: _ConnectionStatusChanged_Handler: 0
```

hbldh:
Please include the code you run to get that result. I cannot help you otherwise.

chinmayjoshi1992:
Actually, I'm running the service_explorer.py example file.

So the scenario is I have one embedded device running BLE. I frequently add-Remove characteristics from the device. Earlier the script was showing the old parameters. So I ran the same script on different machines and IT was showing the correct characteristics.

Is the lib store the characteristics somewhere? If so how can I clear the same? 

Also there is one chara for which it is showing Error 0x0b.

hbldh:
The `service_explorer.py` script is designed to fetch all services, print them and then disconnect. Thus, that the connection is lost is by design.

[...]

PR #431, opened 2021-01-25
Adding option to avoid using cached services in .NET backend.

hbldh:
## Added
* Keyword arguments to ``get_services`` method on ``BleakClient``.
* Keyword argument ``use_cached`` on .NET backend, to enable uncached reading of services, characteristics and descriptors in Windows.

Setting ``use_cached`` to ``True`` by default, since reading them from device noticably increased the connection time. Devices with mutable services are more rare than the opposite I think.

This might solve #429.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,9 @@ Added
 * Added ``AdvertisementServiceData`` in BLEDevice in macOS devices
 * Protection levels (encryption) in Windows backend pairing. Solves #405.
 * Philips Hue lamp example script. Relates to #405.
+* Keyword arguments to ``get_services`` method on ``BleakClient``.
+* Keyword argument ``use_cached`` on .NET backend, to enable uncached reading
+  of services, characteristics and descriptors in Windows.
 
 Fixed
 ~~~~~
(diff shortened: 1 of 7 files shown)


chinmayjoshi1992:
bleak.exc.BleakDotNetTaskError: System.Exception: Catastrophic failure (Exception from HRESULT: 0x8000FFFF (E_UNEXPECTED))

facing this issue

dlech (approved):
Looks OK to me. 

Do you know what the behavior is on macOS? I know on Linux/BlueZ you have to actually remove the device and discover it again to no use the cache. So it might be worth mentioning that in the docs somewhere. It causes problems for people with microcontrollers that can change the services on the device.

hbldh:
I will have to look into #429 before merging this...

No, I have very little knowledge about how it performs in the other backends, primarily because I do not modify my devices services. I implemented this because it looked like low hanging fruit that might solve an issue...

On macOS I have no idea how to handle this and on Linux only what you have found out and shared here and in other issues.
It might go in the FAQ/troubleshooting section that you set up.

hbldh:
Added some documentation on the topic.

dlech:
> Added some documentation on the topic.

I like it. 👍

hbldh:
I might have used your words from another issue describing what needs to be done on BlueZ...😄 🙏🏻

PR #434, opened 2021-01-27
Having async methods as notification handlers

hbldh:
I have gotten a steady stream of questions in my mailbox about having async methods for notification callbacks. This time I had recently seen the [`asyncio.run_coroutine_threadsafe`](https://docs.python.org/3.7/library/asyncio-task.html#asyncio.run_coroutine_threadsafe) method and saw that it might just do the trick in a pretty simple fashion. This is a PR to get feedback on this idea and also testing and sounding boards for macOS.

- Added handling of async callbacks for Windows.
  - [X] Implemented 
  - [X] Tested.
- Added handling of async callbacks for BlueZ.
  - [X] Implemented 
  - [ ] Tested.
- Added handling of async callbacks for Core Bluetooth.
  - [X] Implemented 
  - [ ] Tested.

### DIFF
--- a/examples/sensortag.py
+++ b/examples/sensortag.py
@@ -137,7 +137,8 @@ async def run(address, debug=False):
         battery_level = await client.read_gatt_char(BATTERY_LEVEL_UUID)
         print("Battery Level: {0}%".format(int(battery_level[0])))
 
-        def keypress_handler(sender, data):
+        async def keypress_handler(sender, data):
+            asyncio.get_event_loop().call_soon_threadsafe(lambda x: print(x), "Async!")
             print("{0}: {1}".format(sender, data))
 
         write_value = bytearray([0xA0])
--- a/examples/sensortag.py
+++ b/examples/sensortag.py
@@ -138,7 +138,6 @@ async def run(address, debug=False):
         print("Battery Level: {0}%".format(int(battery_level[0])))
 
         async def keypress_handler(sender, data):
-            asyncio.get_event_loop().call_soon_threadsafe(lambda x: print(x), "Async!")
             print("{0}: {1}".format(sender, data))
 
         write_value = bytearray([0xA0])
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,6 +20,7 @@ Added
 * Keyword argument ``use_cached`` on .NET backend, to enable uncached reading
   of services, characteristics and descriptors in Windows.
 * Documentation on troubleshooting OS level caches for services.
+* New example added: Async callbacks with a queue and external consumer
 
 Fixed
 ~~~~~
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -698,6 +698,15 @@ class BleakClientBlueZDBus(BaseBleakClient):
                 notification to bytearray.
 
         """
+
+        if inspect.iscoroutinefunction(callback):
+
+            def bleak_callback(s, d):
+                asyncio.create_task(callback(s, d))
+
+        else:
+            bleak_callback = callback
+
         _wrap = kwargs.get("notification_wrapper", True)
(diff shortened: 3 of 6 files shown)


examples/sensortag.py
```diff
@@ -137,7 +137,8 @@ async def run(address, debug=False):
         battery_level = await client.read_gatt_char(BATTERY_LEVEL_UUID)
         print("Battery Level: {0}%".format(int(battery_level[0])))
 
-        def keypress_handler(sender, data):
+        async def keypress_handler(sender, data):
+            asyncio.get_event_loop().call_soon_threadsafe(lambda x: print(x), "Async!")
```
dlech, line 141: This line seems odd. We are already in an async function, so why starting another async task from here? And why the need for threadsafe? We are already on the event loop thread.

dlech (commented):
I'm not a fan of all of the code duplication here. And adding extra code in contexts where exceptions can't be thrown/caught isn't the greatest. 

Here is a different idea... instead of handling it at the point of the callback from native code, handle it when the callback is set in `start_notify()`.

I think we can just add something like this:

```python

    async def start_notify(...):
        if inspect.iscoroutinefunction(callback):
            async_callback = callback
            callback = lambda s, d: asyncio.create_task(async_callback(s, d))
        ...
```

Then everything else just works as before.

dlech:
>     * Added handling of async callbacks for BlueZ.  Not tested yet.

I guess we better start thinking about merging the `dbus-next-2` branch before we get too many conflicts.

examples/sensortag.py
hbldh: Yes, I agree, it is odd. I got wrapped up in the async world. What I really wanted to do was call some async method, don't remeber why I did this instead... It would have sufficed with just printing something...

hbldh:
I agree with not adding code where exceptions cannot be caught, but all added stuff that is "new" (except handing it over to the event loop via `run_coroutine_threadsafe`) is outside of the actual callback, in the wrapping decorator code. The only difference is how the callback function is called and that check is called when adding the notification is it not?

I will test the solution you propose instead. It will definitely be cleaner and easier to maintain.

dlech:
> The only difference is how the callback function is called and that check is called when adding the notification is it not?

Yes, it is not too bad, really.

examples/sensortag.py
hbldh: Made the calback async only. It is enough.

hbldh:
The solution you propose is much cleaner, and also possible to aply across all backends. I made it an embedded function instead since my IDE says `PEP 8: E731 do not assign a lambda expression, use a def`, even though I preferred a lambda.

I once had a problem in Windows passing on the notification data that the data buffer was changed between scheduling callback handler and actually running it, so the data that the callback handler received was not the data that was intended. I wanted to add a `copy` on the data in case this is still present, but waiting to see if this is actually a problem. Just saying it if it happens to surface as an issue.

I need to get on the documentation task soon. There is quite a lot that should be put on record...

examples/sensortag.py
```diff
@@ -137,7 +137,7 @@ async def run(address, debug=False):
         battery_level = await client.read_gatt_char(BATTERY_LEVEL_UUID)
         print("Battery Level: {0}%".format(int(battery_level[0])))
 
-        def keypress_handler(sender, data):
+        async def keypress_handler(sender, data):
```
dlech, line 140: It would be nice if we had an example that actually awaited something. Can't think of anything of-hand though.

dlech (approved):

examples/sensortag.py
hbldh: I was thinking I maybe needed to write an example that unloaded data to a async queue. Was planning to wait for the docs work, but it might just fit here better.

hbldh: Added said example.

Issue #362, opened 2020-11-19
Multiple services/characteristics with the same UUID should not be an error


AGlass0fMilk:
* bleak version: 0.9.1
* Python version: 3.9.0
* Operating System: Ubuntu 20.04
* BlueZ version (`bluetoothctl -v`) in case of Linux: 5.53

### Description

There is no universal requirement in the BT specification that a GATT Server cannot have multiple instances of the same service in its profile, or multiple instances of the same characteristic in a service. This restriction _may_ be specified by a GATT Service specification, but it is not the default nature of GATT services.

Currently, bleak stores services in the client object inside a dictionary using the UUID as the key. This innately disallows multiple services with the same UUID. In fact, it raises an exception here:

https://github.com/hbldh/bleak/blob/d7e9558825330d1f09fb7ad4a64096d33586434e/bleak/backends/service.py#L107-L117

I am implementing a [DFU Service for Mbed OS](https://github.com/ARMmbed/mbed-os-experimental-ble-services/pull/13) and I was planning to use bleak as the platform for a desktop-based update tool. Through our discussion of the DFU Service specification, we want to allow multiple DFU Services to exist on the GATT Server so that different embedded devices within the same product may be updated over BLE.

This limitation of bleak will make it impossible to support multiple DFU services on the same GATT server. Another example of a service (one that is BT SIG specified, in fact) that allows multiple instances in this way is the common [Battery Service](https://github.com/hbldh/bleak/files/5568479/BAS_SPEC_V10.pdf):

![ble-battery-svc](https://user-images.githubusercontent.com/350500/99698745-21aa4a00-2a5f-11eb-802e-40a70e3686f0.png)

The suggested fix would be to:
1. Change from storing any attributes (services/characteristics/descriptors, etc) in a map using their UUID as a key. The key should be the _handle_ of the attribute.
1. Implement new functions `get_services_by_uuid()`, `get_characteristics_by_uuid()`, and so on to return _Lists_ of `BleakGattService`, `BleakGattCharacteristic`, ...
1. Deprecate the `get_service()`, `get_descriptor()`, etc functions and in the mean time have them return the _first occurrence_ of a service/descriptor by UUID (ie: return `get_services_by_uuid(uuid)[0])

hbldh:
Characteristics and descriptors already use handles, but for services I was not sure I could get handles in a good fashion in all OS backends, so I let that remain as uuids. I assumed that no one would ever want to split characteristics over two different services with the same uuid...

[...]

PR #449, opened 2021-02-19
Changed Service identification to handles instead of UUID

### LINKED ISSUE #445: bleak.exc.BleakError: This service is already present in this BleakGATTServiceCollection!
* bleak version: 0.10.0
* Python version: 3.7.9
* Operating System: Windows 10

### Description

I am trying to connect to my esp32 through BLE, the esp is running micropython with the standard micropython BLE library. I could already connect the device with my phone using Serial bluetooth interface app (https://play.google.com/store/apps/details?id=de.kai_morich.serial_bluetooth_terminal&hl=nl&gl=US) and send data back and forth.

Now I am trying to connect it to my pc using Bleak. I got some errors at my first attempts saying that the write and read characteristics UUID were wrong. Now I tried the service_explorer example (https://github.com/hbldh/bleak/blob/develop/examples/service_explorer.py) to find these UUID's. But the error: This service is already present in this BleakGATTServiceCollections comes up, this error also appeared in my first attempts to connect the devices. I searched for this error but didn't find any mentions of it.

### What I Did

This is basically the service_explorer example with the esp32 address filled in.

```
async def run(address, debug=False):
    log = logging.getLogger(__name__)
    if debug:
        import sys

        log.setLevel(logging.DEBUG)
        h = logging.StreamHandler(sys.stdout)
        h.setLevel(logging.DEBUG)
        log.addHandler(h)

        x = await client.is_connected()
        log.info("Connected: {0}".format(x))
        for service in client.services:
            log.info("[Service] {0}: {1}".format(service.uuid, service.description))
                if "read" in char.properties:
                    try:
                        value = bytes(await client.read_gatt_char(char.uuid))
                    except Exception as e:
                        value = str(e).encode()
                else:
                    value = None
                log.info(
                    "\t[Characteristic] {0}: (Handle: {1}) ({2}) | Name: {3}, Value: {4} ".format(
                        char.uuid,
                        char.handle,
                        ",".join(char.properties),
                        char.description,
                        value,
                    )
                )
                for descriptor in char.descriptors:
                    value = await client.read_gatt_descriptor(descriptor.handle)
                    log.info(
                        "\t\t[Descriptor] {0}: (Handle: {1}) | Value: {2} ".format(
                            descriptor.uuid, descriptor.handle, bytes(value)
                        )
                    )


if __name__ == "__main__":
    address = (
        "BC:DD:C2:C7:93:8E"
        if platform.system() != "Darwin"
        else "B9EA5233-37EF-4DD6-87A8-2A875E821C46"
    )
    print(address)
    loop.set_debug(True)
    loop.run_until_complete(run(address, True))
```

### Output

```
PS C:> python .\service_explorer.py
BC:DD:C2:C7:93:8E
  File ".\service_explorer.py", line 58, in <module>
    loop.run_until_complete(run(address, True))
  File "C:\AppData\Local\Programs\Python\Python37\lib\asyncio\base_events.py", line 587, in run_until_complete
    return future.result()
  File ".\service_explorer.py", line 18, in run
  File "C:\AppData\Local\Programs\Python\Python37\lib\site-packages\bleak\backends\client.py", line 59, in __aenter__
    await self.connect()
  File "C:\AppData\Local\Programs\Python\Python37\lib\site-packages\bleak\backends\dotnet\client.py", line 238, in connect
    await self.get_services()
  File "C:\AppData\Local\Programs\Python\Python37\lib\site-packages\bleak\backends\dotnet\client.py", line 440, in get_services
    self.services.add_service(BleakGATTServiceDotNet(service))
  File "C:\AppData\Local\Programs\Python\Python37\lib\site-packages\bleak\backends\service.py", line 116, in add_service
    "This service is already present in this BleakGATTServiceCollection!"
bleak.exc.BleakError: This service is already present in this BleakGATTServiceCollection!

```
Thank you in advance!

hbldh:
Identifying Services by UUID is wrong, since a peripheral can have multiple services with the same UUID. The proper identifier should be the handle instead. From previously having thought the service handles was not present in the APIs that Bleak use, I have now done a depper investigation and foudn that I was wrong, they were there and they were quite easily obtained as well.

This should once and for all prevent the `This service is already present in this BleakGATTServiceCollection!` error from occuring for users.

- [X] Windows backend
- [X] BlueZ backend
- [x] CoreBluetooth backend (tested in OS X 10.11 only)

Fixes #445 and #362.

This PR also removes all `__str__` methods from subclasses of `BleakGATT*` in favour of the ones in the abstract base classes.

### DIFF
--- a/bleak/backends/bluezdbus/characteristic.py
+++ b/bleak/backends/bluezdbus/characteristic.py
@@ -29,11 +29,14 @@ _GattCharacteristicsFlagsEnum = {
 class BleakGATTCharacteristicBlueZDBus(BleakGATTCharacteristic):
     """GATT Characteristic implementation for the BlueZ DBus backend"""
 
-    def __init__(self, obj: dict, object_path: str, service_uuid: str):
+    def __init__(
+        self, obj: dict, object_path: str, service_uuid: str, service_handle: int
+    ):
         super(BleakGATTCharacteristicBlueZDBus, self).__init__(obj)
         self.__descriptors = []
         self.__path = object_path
         self.__service_uuid = service_uuid
+        self.__service_handle = service_handle
 
         # D-Bus object path contains handle as last 4 characters of 'charYYYY'
         self._handle = int(object_path[-4:], 16)
@@ -42,6 +45,10 @@ class BleakGATTCharacteristicBlueZDBus(BleakGATTCharacteristic):
     def service_uuid(self) -> str:
         """The uuid of the Service containing this characteristic"""
         return self.__service_uuid
+
+    @property
+    def service_handle(self) -> int:
+        return self.__service_handle
 
     @property
     def handle(self) -> int:
--- a/bleak/backends/bluezdbus/utils.py
+++ b/bleak/backends/bluezdbus/utils.py
@@ -2,11 +2,13 @@
 import asyncio
 import re
 
+from bleak import BleakError
 from bleak.uuids import uuidstr_to_str
 
 from bleak.backends.bluezdbus import defs
 
 _mac_address_regex = re.compile("^([0-9A-Fa-f]{2}[:-]){5}([0-9A-Fa-f]{2})$")
+_service_handle_regex = re.compile('service([0-9A-Fa-f]*)')
 
 
 def validate_mac_address(address):
@@ -46,3 +48,11 @@ def format_GATT_object(object_path, interfaces):
     return "\n{0}\n\t{1}\n\t{2}\n\t{3}".format(
         _type, object_path, _uuid, uuidstr_to_str(_uuid)
     )
+
+
+def extract_service_handle_from_path(path):
+    try:
+        return int(_service_handle_regex.search(path).groups()[-1], 16)
+    except Exception as e:
+        raise BleakError(f"Could not parse service handle from path: {path} ({e})")
+
--- a/bleak/backends/bluezdbus/utils.py
+++ b/bleak/backends/bluezdbus/utils.py
@@ -8,7 +8,6 @@ from bleak.uuids import uuidstr_to_str
 from bleak.backends.bluezdbus import defs
 
 _mac_address_regex = re.compile("^([0-9A-Fa-f]{2}[:-]){5}([0-9A-Fa-f]{2})$")
-_service_handle_regex = re.compile("service([0-9A-Fa-f]*)")
 
 
 def validate_mac_address(address):
@@ -52,6 +51,6 @@ def format_GATT_object(object_path, interfaces):
 
 def extract_service_handle_from_path(path):
     try:
-        return int(_service_handle_regex.search(path).groups()[-1], 16)
+        return int(path[-4:], 16)
     except Exception as e:
-        raise BleakError(f"Could not parse service handle from path: {path} ({e})")
+        raise BleakError(f"Could not parse service handle from path: {path}") from e
--- a/bleak/backends/bluezdbus/utils.py
+++ b/bleak/backends/bluezdbus/utils.py
@@ -8,7 +8,7 @@ from bleak.uuids import uuidstr_to_str
 from bleak.backends.bluezdbus import defs
 
 _mac_address_regex = re.compile("^([0-9A-Fa-f]{2}[:-]){5}([0-9A-Fa-f]{2})$")
-_service_handle_regex = re.compile('service([0-9A-Fa-f]*)')
+_service_handle_regex = re.compile("service([0-9A-Fa-f]*)")
 
 
 def validate_mac_address(address):
@@ -55,4 +55,3 @@ def extract_service_handle_from_path(path):
         return int(_service_handle_regex.search(path).groups()[-1], 16)
     except Exception as e:
         raise BleakError(f"Could not parse service handle from path: {path} ({e})")
-
--- a/bleak/backends/characteristic.py
+++ b/bleak/backends/characteristic.py
@@ -34,12 +34,18 @@ class BleakGATTCharacteristic(abc.ABC):
         self.obj = obj
 
     def __str__(self):
-        return "{0}: {1}".format(self.uuid, self.description)
+        return f"{self.uuid} (Handle: {self.handle}): {self.description}"
 
     @property
     @abc.abstractmethod
     def service_uuid(self) -> str:
         """The UUID of the Service containing this characteristic"""
+        raise NotImplementedError()
+
+    @property
+    @abc.abstractmethod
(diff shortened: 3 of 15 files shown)


bleak/backends/bluezdbus/characteristic.py
```diff
@@ -29,11 +29,14 @@
...
 
-    def __init__(self, obj: dict, object_path: str, service_uuid: str):
+    def __init__(
+        self, obj: dict, object_path: str, service_uuid: str, service_handle: int
+    ):
         super(BleakGATTCharacteristicBlueZDBus, self).__init__(obj)
         self.__descriptors = []
         self.__path = object_path
         self.__service_uuid = service_uuid
+        self.__service_handle = service_handle
```
hbldh, line 39: The service handle could probably be obtained via regexp solution from the characteristic's path, but since we already send in the uuid, I sent in the handle as well.

examples/service_explorer.py
```diff
@@ -40,19 +40,20 @@ async def run(address, debug=False):
...
+                            await client.read_gatt_descriptor(descriptor.handle)
+                        )
+                    except Exception as e:
+                        value = str(e).encode()
                     log.info(
-                        "\t\t[Descriptor] {0}: (Handle: {1}) | Value: {2} ".format(
-                            descriptor.uuid, descriptor.handle, bytes(value)
+                        f"\t\t[Descriptor] {descriptor}) | Value: {value} ".format(
+                            descriptor.uuid,
+                            descriptor.handle,
```
dlech, line 56: maybe move `log.info` into the try block and `log.error` on exception?

dlech (commented):
This seems straightforward. Just a few minor comments inline.

bleak/backends/bluezdbus/utils.py
```diff
@@ -2,11 +2,13 @@
 import asyncio
 import re
 
+from bleak import BleakError
 from bleak.uuids import uuidstr_to_str
 
 from bleak.backends.bluezdbus import defs
 
 _mac_address_regex = re.compile("^([0-9A-Fa-f]{2}[:-]){5}([0-9A-Fa-f]{2})$")
+_service_handle_regex = re.compile("service([0-9A-Fa-f]*)")
```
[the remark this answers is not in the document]
hbldh: Yes. Will change this.

bleak/backends/bluezdbus/utils.py
```diff
@@ -46,3 +48,10 @@ def format_GATT_object(object_path, interfaces):
     return "\n{0}\n\t{1}\n\t{2}\n\t{3}".format(
         _type, object_path, _uuid, uuidstr_to_str(_uuid)
     )
+
+
+def extract_service_handle_from_path(path):
+    try:
+        return int(_service_handle_regex.search(path).groups()[-1], 16)
+    except Exception as e:
+        raise BleakError(f"Could not parse service handle from path: {path} ({e})")
```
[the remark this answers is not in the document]
hbldh: Yes. Forgot about this solution. I write few libraries except this and pass on exceptions instead of directly handling them, so I tend to forget to do this...

bleak/backends/characteristic.py
```diff
@@ -34,14 +34,20 @@ def __init__(self, obj: Any):
...
     def service_uuid(self) -> str:
         """The UUID of the Service containing this characteristic"""
         raise NotImplementedError()
 
+    @property
+    @abc.abstractmethod
+    def service_handle(self) -> int:
+        """The integer handle of the Service containing this characteristic"""
+        raise NotImplementedError()
+
```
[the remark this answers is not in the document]
hbldh: It would have been better, yes. It will have to be done in a later PR though.

bleak/backends/corebluetooth/service.py
```diff
@@ -15,6 +15,12 @@ class BleakGATTServiceCoreBluetooth(BleakGATTService):
     def __init__(self, obj: CBService):
         super().__init__(obj)
         self.__characteristics = []
+        self.__handle = int(self.obj.startHandle())
```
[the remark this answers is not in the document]
hbldh: Yes. Important.

bleak/backends/service.py
```diff
@@ -91,34 +97,50 @@ def __iter__(self) -> Iterator[BleakGATTService]:
 
     @property
     def services(self) -> dict:
-        """Returns dictionary of UUID strings to BleakGATTService"""
+        """Returns dictionary of handles mapping to BleakGATTService"""
```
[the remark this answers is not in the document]
hbldh: Agree.

examples/service_explorer.py
hbldh: It got a bit messier, but the logging gets better.

examples/service_explorer.py
```diff
@@ -26,35 +26,39 @@ async def run(address, debug=False):
...
-                            descriptor.uuid, descriptor.handle, bytes(value)
+                    try:
+                        value = bytes(
+                            await client.read_gatt_descriptor(descriptor.handle)
                         )
-                    )
+                        log.info(f"\t\t[Descriptor] {descriptor}) | Value: {value}")
+                    except Exception as e:
+                        value = str(e).encode()
+                        log.error(f"\t\t[Descriptor] {descriptor}) | Value: {value}")
```
dlech, line 61: Shouldn't we be able to use `e` directly in the format string?

    log.error(f"\t\t[Descriptor] {descriptor}) | Exception: {e}")

bleak/backends/service.py
```diff
@@ -79,46 +85,64 @@ def __init__(self):
 
     def __getitem__(
         self, item: Union[str, int, UUID]
-    ) -> Union[BleakGATTService, BleakGATTCharacteristic, BleakGATTDescriptor]:
+    ) -> Union[BleakGATTService, BleakGATTCharacteristic, BleakGATTDescriptor, None]:
```
dlech, line 88: FWIW, I think I have been using this patter in Bleak.

    Optional[Union[BleakGATTService, BleakGATTCharacteristic, BleakGATTDescriptor]]

examples/service_explorer.py
hbldh: Interesting. I wonder why I I thought this was a good idea? I think I just wanted it to be a bytes object like values in general. There is no need for that though.

bleak/backends/service.py
hbldh: Did not know about `Optional` and it being the preferred solution. Thank you!

AGlass0fMilk:
Fantastic work!!! :+1:

Issue #180, opened 2020-05-05
Change from pythonnet to WinRT


cHemingway:
* bleak version: 0.6.1
* Python version: 3.8
* Operating System: Windows 10 1909
* BlueZ version (`bluetoothctl -v`) in case of Linux: 

### Description
Bleak fails to install on Windows 10 with Python 3.8, as Python 3.8 is not supported by pythonnet, which appears to not officially support python 3.7 or 3.8. 

Microsoft have now released [WinRT](https://pypi.org/project/winrt/), which is an official python bridge to the WindowsRT APIs, and so it might be possible to use this instead of PythonNet for access to Bluetooth on windows. See https://github.com/beeware/toga/issues/884 for another project that is discussing making this change. 

Hopefully, this might solve other open issues tagged with .NET too.

cHemingway:
Had a quick go, some immediate issues.

- A lot of base types are missing as `System` namespace doesn't seem to be exposed, so `from System import Array, Byte, UInt64` does not work as `from winrt.system import Array,Byte,UInt64`. The WinRT bridge might convert some of these types automatically, so we don't need to use them directly.

- Enum definitions are different as WinRT capitalises them, so `BluetoothLEAdvertisementType.ScanResponse` becomes `BluetoothLEAdvertisementType.SCAN_RESPONSE`

- `utils.py` uses `System.Threading.Task` which is not exposed. Would need to be replaced with `Windows.System.Threading`

hbldh:
This is indeed interesting, especially if the BeeWare project deems it to be. 

Still, it does not seem to solve the Python 3.8 problem in Windows though according to the Toga , so changing is not really worthwhile to do right now.

claybrooks:
For what it's worth, I was able to get pythonnet working with python3.8.  `pip install pythonnet` did not work, I had to build from source from their github.

hbldh:
Finally got some time to wxplored this path in the `feature/winrt` branch. I have converted the entire `pythonnet` solution to `winrt` and it works, with some caveats.
Apparently `winrt` does not convert `GUID` objects properly, so no `uuid` attributes can be read (raises a `NotImplementedError`).  One can get around this by some ugly regex-solutions for the service uuids, but characteristics has to be queried by handle (or by the characteristic object itself).

[...]

PR #450, opened 2021-02-19
Brought the WinRT backend up to date with develop

hbldh:
Implements #180.

### DIFF
--- a/bleak/__init__.py
+++ b/bleak/__init__.py
@@ -77,8 +77,19 @@ elif platform.system() == "Windows":
             "Requires at least Windows 10 version 0.16299 (Fall Creators Update)."
         )
 
-    from bleak.backends.dotnet.scanner import BleakScannerDotNet as BleakScanner  # noqa
-    from bleak.backends.dotnet.client import BleakClientDotNet as BleakClient  # noqa
+    py_major, py_minor, _ = platform.python_version_tuple()
+    if int(py_major) >= 3 and int(py_minor) >= 9:
+        from bleak.backends.winrt.scanner import (
+            BleakScannerWinRT as BleakScanner,
+        )  # noqa
+        from bleak.backends.winrt.client import BleakClientWinRT as BleakClient  # noqa
+    else:
+        from bleak.backends.dotnet.scanner import (
+            BleakScannerDotNet as BleakScanner,
+        )  # noqa
+        from bleak.backends.dotnet.client import (
+            BleakClientDotNet as BleakClient,
+        )  # noqa
 else:
     raise BleakError(f"Unsupported platform: {platform.system()}")
 
--- a/bleak/__init__.py
+++ b/bleak/__init__.py
@@ -82,15 +82,17 @@ elif platform.system() == "Windows":
     try:
         from bleak.backends.winrt.scanner import (
             BleakScannerWinRT as BleakScanner,
-        )  # noqa
-        from bleak.backends.winrt.client import BleakClientWinRT as BleakClient  # noqa
+        )  # noqa: F401
+        from bleak.backends.winrt.client import (
+            BleakClientWinRT as BleakClient,
+        )  # noqa: F401
     except ImportError:
         from bleak.backends.dotnet.scanner import (
             BleakScannerDotNet as BleakScanner,
-        )  # noqa
+        )  # noqa: F401
         from bleak.backends.dotnet.client import (
             BleakClientDotNet as BleakClient,
-        )  # noqa
+        )  # noqa: F401
 
 else:
     raise BleakError(f"Unsupported platform: {platform.system()}")
--- a/bleak/__init__.py
+++ b/bleak/__init__.py
@@ -20,7 +20,7 @@ _on_rtd = os.environ.get("READTHEDOCS") == "True"
 
 _logger = logging.getLogger(__name__)
 _logger.addHandler(logging.NullHandler())
-if bool(os.environ.get("BLEAK_LOGGING", True)):
+if bool(os.environ.get("BLEAK_LOGGING", False)):
     FORMAT = "%(asctime)-15s %(name)-8s %(levelname)s: %(message)s"
     handler = logging.StreamHandler(sys.stdout)
     handler.setLevel(logging.DEBUG)
--- a/bleak/__init__.py
+++ b/bleak/__init__.py
@@ -77,19 +77,21 @@ elif platform.system() == "Windows":
             "Requires at least Windows 10 version 0.16299 (Fall Creators Update)."
         )
 
-    py_major, py_minor, _ = platform.python_version_tuple()
-    if int(py_major) >= 3 and int(py_minor) >= 9:
+    # If the winrt package is installed, assume that the user has opted to use that backend
+    # instead of the pythonnet/BleakBridge implementation.
+    try:
         from bleak.backends.winrt.scanner import (
             BleakScannerWinRT as BleakScanner,
         )  # noqa
         from bleak.backends.winrt.client import BleakClientWinRT as BleakClient  # noqa
-    else:
+    except ImportError:
         from bleak.backends.dotnet.scanner import (
             BleakScannerDotNet as BleakScanner,
         )  # noqa
         from bleak.backends.dotnet.client import (
             BleakClientDotNet as BleakClient,
         )  # noqa
+
 else:
     raise BleakError(f"Unsupported platform: {platform.system()}")
 
--- a/bleak/__init__.py
+++ b/bleak/__init__.py
@@ -20,7 +20,7 @@ _on_rtd = os.environ.get("READTHEDOCS") == "True"
 
 _logger = logging.getLogger(__name__)
 _logger.addHandler(logging.NullHandler())
-if bool(os.environ.get("BLEAK_LOGGING", False)):
+if bool(os.environ.get("BLEAK_LOGGING", True)):
     FORMAT = "%(asctime)-15s %(name)-8s %(levelname)s: %(message)s"
     handler = logging.StreamHandler(sys.stdout)
     handler.setLevel(logging.DEBUG)
--- a/bleak/__init__.py
+++ b/bleak/__init__.py
@@ -121,4 +121,3 @@ def cli():
 
 if __name__ == "__main__":
     cli()
-# dfs
--- a/bleak/__init__.py
+++ b/bleak/__init__.py
@@ -121,3 +121,4 @@ def cli():
 
 if __name__ == "__main__":
     cli()
+# dfs
--- a/bleak/__init__.py
+++ b/bleak/__init__.py
@@ -78,7 +78,7 @@ elif platform.system() == "Windows":
         )
 
     py_major, py_minor, _ = platform.python_version_tuple()
-    if int(py_major) >= 3 and int(py_minor) >= 7:
+    if int(py_major) >= 3 and int(py_minor) >= 9:
         from bleak.backends.winrt.scanner import (
             BleakScannerWinRT as BleakScanner,
         )  # noqa
--- a/bleak/__init__.py
+++ b/bleak/__init__.py
@@ -20,7 +20,7 @@ _on_rtd = os.environ.get("READTHEDOCS") == "True"
 
 _logger = logging.getLogger(__name__)
 _logger.addHandler(logging.NullHandler())
-if bool(os.environ.get("BLEAK_LOGGING", False)):
+if bool(os.environ.get("BLEAK_LOGGING", True)):
     FORMAT = "%(asctime)-15s %(name)-8s %(levelname)s: %(message)s"
     handler = logging.StreamHandler(sys.stdout)
     handler.setLevel(logging.DEBUG)
@@ -78,7 +78,7 @@ elif platform.system() == "Windows":
         )
 
     py_major, py_minor, _ = platform.python_version_tuple()
-    if int(py_major) >= 3 and int(py_minor) >= 9:
+    if int(py_major) >= 3 and int(py_minor) >= 7:
         from bleak.backends.winrt.scanner import (
             BleakScannerWinRT as BleakScanner,
         )  # noqa
--- a/bleak/backends/dotnet/client.py
+++ b/bleak/backends/dotnet/client.py
@@ -9,7 +9,7 @@ import logging
 import asyncio
 import uuid
 from functools import wraps
-from typing import Callable, Any, Union
+from typing import Callable, Any, List, Union
 
 from bleak.backends.device import BLEDevice
 from bleak.backends.dotnet.scanner import BleakScannerDotNet
@@ -37,6 +37,7 @@ from BleakBridge import Bridge  # noqa: F401
 from System import UInt64, Object
 from System.Runtime.InteropServices.WindowsRuntime import EventRegistrationToken
 from Windows.Foundation import IAsyncOperation, TypedEventHandler
+from Windows.Security.Cryptography import CryptographicBuffer
 from Windows.Devices.Enumeration import (
     DevicePairingResult,
     DevicePairingResultStatus,
@@ -113,8 +114,8 @@ class BleakClientDotNet(BaseBleakClient):
         else:
             self._device_info = None
         self._requester = None
-        self._connect_events: list[asyncio.Event] = []
-        self._disconnect_events: list[asyncio.Event] = []
+        self._connect_events: List[asyncio.Event] = []
+        self._disconnect_events: List[asyncio.Event] = []
         self._connection_status_changed_token: EventRegistrationToken = None
         self._session: GattSession = None
 
--- a/bleak/backends/dotnet/client.py
+++ b/bleak/backends/dotnet/client.py
@@ -415,7 +415,7 @@ class BleakClientDotNet(BaseBleakClient):
 
         return not self._requester.DeviceInformation.Pairing.IsPaired
 
-    #     #GATT services methods
+    # GATT services methods
 
     async def get_services(self, **kwargs) -> BleakGATTServiceCollection:
         """Get all services registered for this GATT server.
--- a/bleak/backends/dotnet/client.py
+++ b/bleak/backends/dotnet/client.py
@@ -37,7 +37,6 @@ from BleakBridge import Bridge  # noqa: F401
 from System import UInt64, Object
 from System.Runtime.InteropServices.WindowsRuntime import EventRegistrationToken
 from Windows.Foundation import IAsyncOperation, TypedEventHandler
-from Windows.Security.Cryptography import CryptographicBuffer
 from Windows.Devices.Enumeration import (
     DevicePairingResult,
     DevicePairingResultStatus,
--- a/bleak/backends/dotnet/client.py
+++ b/bleak/backends/dotnet/client.py
@@ -310,7 +310,7 @@ class BleakClientDotNet(BaseBleakClient):
 
         Keyword Args:
             protection_level:
-                    DevicePairingProtectionLevel
+                    Windows.Devices.Enumeration.DevicePairingProtectionLevel
                         1: None - Pair the device using no levels of protection.
                         2: Encryption - Pair the device using encryption.
                         3: EncryptionAndAuthentication - Pair the device using
@@ -560,7 +560,6 @@ class BleakClientDotNet(BaseBleakClient):
     async def read_gatt_char(
         self,
         char_specifier: Union[BleakGATTCharacteristic, int, str, uuid.UUID],
-        use_cached=False,
         **kwargs,
     ) -> bytearray:
         """Perform read operation on the specified GATT characteristic.
@@ -569,6 +568,8 @@ class BleakClientDotNet(BaseBleakClient):
             char_specifier (BleakGATTCharacteristic, int, str or UUID): The characteristic to read from,
                 specified by either integer handle, UUID or directly by the
                 BleakGATTCharacteristic object representing it.
+
+        Keyword Args:
             use_cached (bool): ``False`` forces Windows to read the value from the
                 device again and not use its own cached value. Defaults to ``False``.
 
@@ -576,6 +577,9 @@ class BleakClientDotNet(BaseBleakClient):
             (bytearray) The read data.
 
         """
+
+        use_cached = kwargs.get("use_cached", False)
+
         if not isinstance(char_specifier, BleakGATTCharacteristic):
             characteristic = self.services.get_characteristic(char_specifier)
         else:
@@ -620,13 +624,13 @@ class BleakClientDotNet(BaseBleakClient):
                 )
         return value
 
-    async def read_gatt_descriptor(
-        self, handle: int, use_cached=False, **kwargs
-    ) -> bytearray:
+    async def read_gatt_descriptor(self, handle: int, **kwargs) -> bytearray:
         """Perform read operation on the specified GATT descriptor.
 
         Args:
             handle (int): The handle of the descriptor to read from.
+
+        Keyword Args:
             use_cached (bool): ``False`` forces Windows to read the value from the
                 device again and not use its own cached value. Defaults to ``False``.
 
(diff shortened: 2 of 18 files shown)


bleak/__init__.py
```diff
@@ -77,8 +77,13 @@
             "Requires at least Windows 10 version 0.16299 (Fall Creators Update)."
         )
 
-    from bleak.backends.dotnet.scanner import BleakScannerDotNet as BleakScanner  # noqa
-    from bleak.backends.dotnet.client import BleakClientDotNet as BleakClient  # noqa
+    py_major, py_minor, _ = platform.python_version_tuple()
+    if int(py_major) >= 3 and int(py_minor) >= 9:
```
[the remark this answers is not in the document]
dlech: Maybe we could make the backend selectable with an environment variable?

dlech (commented):
I haven't looked at the new files yet, but I have made a few initial comments.

bleak/__init__.py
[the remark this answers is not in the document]
hbldh: I would like it to be selectable as you say, but given that `pythonnet` will break installation on 3.9+ I have no idea on how to make them coexist in that fashion. Only 3.7 and 3.8 would be able to install them both and it felt better to make pythonnet an extras which one could install.  This was best I could think of... 

I wrote some thoughts about it in #180 as well.

bleak/backends/dotnet/client.py
```diff
@@ -412,7 +412,7 @@ async def unpair(self) -> bool:
 
         return not self._requester.DeviceInformation.Pairing.IsPaired
 
-    # GATT services methods
+    #     #GATT services methods
```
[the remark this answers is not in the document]
hbldh: Yes, will fix.

bleak/backends/winrt/client.py
```diff
@@ -0,0 +1,836 @@
...
+                        data,
+                        descriptor.uuid,
+                        _communication_statues.get(write_result.status, ""),
+                    )
+                )
+
+    async def start_notify(
+        self,
+        char_specifier: Union[BleakGATTCharacteristic, int, str, uuid.UUID],
+        callback: Callable[[str, Any], Any],
```
dlech, line 715: `Callable[[str, bytearray], None]`?

Isn't sender `int` rather than `str`, so `Callable[[int, bytearray], None]`?

bleak/backends/winrt/client.py
```diff
@@ -0,0 +1,836 @@
...
+                notification/indication on, specified by either integer handle, UUID or
+                directly by the BleakGATTCharacteristic object representing it.
+
+        """
+        if not isinstance(char_specifier, BleakGATTCharacteristic):
+            characteristic = self.services.get_characteristic(char_specifier)
+        else:
+            characteristic = char_specifier
+        if not characteristic:
+            raise BleakError("Characteristic {} not found!".format(char_specifier))
```
dlech, line 806: `f""`? Here and many other places

bleak/__init__.py
```diff
@@ -20,7 +20,7 @@
 
 _logger = logging.getLogger(__name__)
 _logger.addHandler(logging.NullHandler())
-if bool(os.environ.get("BLEAK_LOGGING", False)):
+if bool(os.environ.get("BLEAK_LOGGING", True)):
```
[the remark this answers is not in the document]
hbldh: Ooops. This was bad. Will fix.

bleak/backends/winrt/characteristic.py
```diff
@@ -0,0 +1,128 @@
...
+)
+
+
+_GattCharacteristicsPropertiesEnum = {
+    GattCharacteristicProperties.NONE: (
+        "None",
+        "The characteristic doesn’t have any properties that apply",
+    ),
+    GattCharacteristicProperties.BROADCAST: (
+        "Broadcast".lower(),
```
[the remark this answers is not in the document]
hbldh: Since I copied it from the .NET documentation page and went over it with a column caret...

setup.py
```diff
@@ -26,7 +26,8 @@
     'pyobjc-framework-CoreBluetooth;platform_system=="Darwin"',
     'pyobjc-framework-libdispatch;platform_system=="Darwin"',
     # Windows reqs
-    'pythonnet;platform_system=="Windows"',
+    'pythonnet;platform_system=="Windows" and python_version < "3.7.0"',
```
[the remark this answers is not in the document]
hbldh: No, I know. I made some changes in setup.py to get a user to install the WinRT branch and guarantee that WinRT was used regardless of Python version. Will change this back before making the draft into a real PR.

bleak/backends/winrt/client.py
```diff
@@ -0,0 +1,836 @@
...
+                                    _communication_statues.get(
+                                        descriptors_result.status, ""
+                                    ),
+                                )
+                            )
+                    for descriptor in list(descriptors_result.descriptors):
+                        self.services.add_descriptor(
+                            BleakGATTDescriptorWinRT(
+                                descriptor,
+                                "",
```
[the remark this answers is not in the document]
hbldh: Strange bug.

bleak/backends/winrt/client.py
hbldh: I am planing on doing a separate PR with f-strings later on after applying [flynt](https://github.com/ikamensh/flynt) on the codebase.

dlech: nice

bleak/backends/winrt/client.py
```diff
@@ -0,0 +1,835 @@
...
+    def __init__(self, address_or_ble_device: Union[BLEDevice, str], **kwargs):
+        super(BleakClientWinRT, self).__init__(address_or_ble_device, **kwargs)
+
+        # Backend specific. WinRT objects.
+        if isinstance(address_or_ble_device, BLEDevice):
+            self._device_info = address_or_ble_device.details.BluetoothAddress
+        else:
+            self._device_info = None
+        self._requester = None
+        self._connect_events: list[asyncio.Event] = []
```
dlech, line 100: should be `List[]` rather than `list[]`

bleak/backends/winrt/client.py
```diff
@@ -0,0 +1,835 @@
...
+
+        read_result = await characteristic.obj.read_value_async(
+            BluetoothCacheMode.CACHED if use_cached else BluetoothCacheMode.UNCACHED
+        )
+
+        if read_result.status == GattCommunicationStatus.SUCCESS:
+            reader = DataReader.from_buffer(read_result.value)
+            # TODO: Figure out how to use read_bytes instead...
+            value = bytearray(
+                [reader.read_byte() for _ in range(reader.unconsumed_buffer_length)]
```
dlech, line 534: I made an interesting find. No reader required.

```python
from winrt.windows.security.cryptography import CryptographicBuffer
...
value = bytearray(CryptographicBuffer.copy_to_byte_array(read_result.value))
```

The `copy_to_byte_array` method returns a list of integers.

bleak/backends/winrt/client.py
```diff
@@ -0,0 +1,835 @@
...
+
+        writer = DataWriter()
+        writer.write_bytes(list(data))
+        response = (
+            GattWriteOption.WRITE_WITH_RESPONSE
+            if response
+            else GattWriteOption.WRITE_WITHOUT_RESPONSE
+        )
+        write_result = await characteristic.obj.write_value_with_result_async(
+            writer.detach_buffer(), response
```
dlech, line 642:
```python
CryptographicBuffer.create_from_byte_array(list(data))
```
instead of writer

bleak/backends/winrt/client.py
hbldh: A `CryptographicBuffer`?! Why yes, it makes sense, but I would never have though of looking for what I needed there...

It does introduce an element of surprise to the potential code reader though. I would have wondered why the developers had chosen to do this when it had nothing to do with cryptography.

It is almost three times as fast as well. Good find!

bleak/backends/winrt/client.py
hbldh: The speed of this is not as critical, but it is almost twice as fast as the DataWriter solution. Nice!

hbldh:
The `Windows.Security` was not available through `pythonnet` I am afraid, so the old `dotnet` backend will have to make do with what it has. I guess I need to do some BleakUWPBridge magic to access it, and I think it is not worth the effort.

bleak/backends/winrt/client.py
dlech: I came across this class before because it is the only place in .NET where there is a byte array to hex string conversion function.

hbldh:
This PR is now ready for review. I was planning on getting it over to develop quite soon so that it might be tested by users as well, before any 0.12.0 is under way.

bleak/backends/winrt/client.py
```diff
@@ -0,0 +1,824 @@
...
+            )
+        else:
+            return False
+
+    async def pair(self, protection_level=None, **kwargs) -> bool:
+        """Attempts to pair with the device.
+
+        Keyword Args:
+            protection_level:
+                    DevicePairingProtectionLevel
```
dlech, line 288: Is this a `winrt` type? Maybe use the full namespace here to make it clear. Or use strings instead so that the code can work cross-platform.

bleak/backends/winrt/client.py
```diff
@@ -0,0 +1,824 @@
...
+            logger.info("Services resolved for %s", str(self))
+            self._services_resolved = True
+            return self.services
+
+    # I/O methods
+
+    async def read_gatt_char(
+        self,
+        char_specifier: Union[BleakGATTCharacteristic, int, str, uuid.UUID],
+        use_cached=False,
```
dlech, line 504: type hint for `use_cached`? Or make it part of `**kwargs` like `get_services()`?

bleak/backends/winrt/client.py
```diff
@@ -0,0 +1,824 @@
...
+                raise BleakError(
+                    "Could not read characteristic value for {0}: {1}".format(
+                        characteristic.uuid,
+                        _communication_statues.get(read_result.status, ""),
+                    )
+                )
+        return value
+
+    async def read_gatt_descriptor(
+        self, handle: int, use_cached=False, **kwargs
```
dlech, line 558: same comment about `use_cached`

bleak/backends/winrt/client.py
```diff
@@ -0,0 +1,824 @@
...
+
+        fcn = _notification_wrapper(bleak_callback, asyncio.get_event_loop())
+        fcn_token = characteristic_obj.add_value_changed(fcn)
+        self._notification_callbacks[characteristic.handle] = fcn, fcn_token
+        status = await characteristic_obj.write_client_characteristic_configuration_descriptor_async(
+            cccd
+        )
+
+        if status != GattCommunicationStatus.SUCCESS:
+            # This usually happens when a device reports that it support indicate,
```
dlech, line 766: _supports_

bleak/backends/winrt/client.py
```diff
@@ -0,0 +1,824 @@
...
+            characteristic_obj.characteristic_properties
+            & GattCharacteristicProperties.NOTIFY
+        ):
+            cccd = GattClientCharacteristicConfigurationDescriptorValue.NOTIFY
+        else:
+            cccd = GattClientCharacteristicConfigurationDescriptorValue.NONE
+
+        fcn = _notification_wrapper(bleak_callback, asyncio.get_event_loop())
+        fcn_token = characteristic_obj.add_value_changed(fcn)
+        self._notification_callbacks[characteristic.handle] = fcn, fcn_token
```
dlech, line 760: Do we actually need to include `fcn` here? I don't see where it is used later anywhere.

Also `event_handler_token` might be a better name than `fcn_token` to indicate that this is a .NET event handler we are talking about.

bleak/backends/winrt/scanner.py
```diff
@@ -0,0 +1,194 @@
...
+        self._received_token = None
+        self._stopped_token = None
+
+    def _received_handler(
+        self,
+        sender: winbleadv.BluetoothLEAdvertisementWatcher,
+        event_args: winbleadv.BluetoothLEAdvertisementReceivedEventArgs,
+    ):
+        """Callback for AdvertisementWatcher.Received"""
+        # TODO: Cannot check for if sender == self.watcher in winrt?
```
dlech, line 65: How about `sender.equals(self.watcher)`? Although when would it never be not equal?

bleak/backends/winrt/scanner.py
```diff
@@ -0,0 +1,194 @@
...
+            return
+
+        # Get a "BLEDevice" from parse_event args
+        device = self.parse_eventargs(event_args)
+
+        # Decode service data
+        service_data = {}
+        # 0x16 is service data with 16-bit UUID
+        for section in event_args.advertisement.get_sections_by_type(0x16):
+            # TODO: Figure out how to use read_bytes instead...
```
dlech, line 87: use crypto API here instead of reader?

bleak/backends/winrt/scanner.py
```diff
@@ -0,0 +1,194 @@
...
+            self.watcher.remove_stopped(self._stopped_token)
+        except Exception as e:
+            logger.debug("Could not remove event handlers: {0}...".format(e))
+
+        self._stopped_token = None
+        self._received_token = None
+
+        self.watcher = None
+
+    async def set_scanning_filter(self, **kwargs):
```
dlech, line 158: this is no longer async. 

also, should copy implementation from dotnet backend or raise not NotImplementedError

bleak/backends/winrt/scanner.py
```diff
@@ -0,0 +1,194 @@
...
+                and event_args.bluetooth_address in self._scan_responses
+            ):
+                new_device.name = self._scan_responses[
+                    event_args.bluetooth_address
+                ].advertisement.local_name
+            found.append(new_device)
+
+        return found
+
+    def parse_eventargs(self, event_args):
```
dlech, line 176: maybe should be `_parse_event_args`?

bleak/backends/winrt/service.py
```diff
@@ -0,0 +1,46 @@
...
+from winrt.windows.devices.bluetooth.genericattributeprofile import GattDeviceService
+
+
+class BleakGATTServiceWinRT(BleakGATTService):
+    """GATT Characteristic implementation for the .NET backend, implemented with WinRT"""
+
+    def __init__(self, obj: GattDeviceService):
+        super().__init__(obj)
+        self.__characteristics = [
+            # BleakGATTCharacteristicDotNet(c) for c in obj.GetAllCharacteristics()
```
dlech, line 16: should this comment be deleted or is it a TODO?

setup.py
```diff
@@ -91,6 +92,11 @@ def run(self):
     install_requires=REQUIRED,
     test_suite="tests",
     tests_require=TEST_REQUIRED,
+    extras_require={
```
dlech, line 95: is this still needed?

dlech (commented):
I haven't tried running it yet, but I found a few things to nit-pick anyway. :smile:

bleak/__init__.py
```diff
@@ -77,8 +77,19 @@
             "Requires at least Windows 10 version 0.16299 (Fall Creators Update)."
         )
 
-    from bleak.backends.dotnet.scanner import BleakScannerDotNet as BleakScanner  # noqa
-    from bleak.backends.dotnet.client import BleakClientDotNet as BleakClient  # noqa
+    py_major, py_minor, _ = platform.python_version_tuple()
+    if int(py_major) >= 3 and int(py_minor) >= 9:
```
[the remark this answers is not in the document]
hbldh: Yes, forgot about this. Added this comment:

```
# If the winrt package is installed, assume that the user has opted to use that backend
# instead of the pythonnet/BleakBridge implementation.
```

bleak/backends/dotnet/client.py
```diff
@@ -37,6 +37,7 @@
 from System import UInt64, Object
 from System.Runtime.InteropServices.WindowsRuntime import EventRegistrationToken
 from Windows.Foundation import IAsyncOperation, TypedEventHandler
+from Windows.Security.Cryptography import CryptographicBuffer
```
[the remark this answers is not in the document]
hbldh: 👍🏻

bleak/backends/winrt/client.py
```diff
@@ -0,0 +1,824 @@
...
+
+        # Start a GATT Session to connect
+        event = asyncio.Event()
+        self._connect_events.append(event)
+        try:
+            self._session = await GattSession.from_device_id_async(
+                self._requester.bluetooth_device_id
+            )
+            # This keeps the device connected until we dispose the session or
+            # until we set MaintainConnection = False.
```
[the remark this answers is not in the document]
hbldh: 👍🏻

bleak/backends/winrt/client.py
```diff
@@ -0,0 +1,824 @@
...
+    getattr(DeviceUnpairingResultStatus, v): v
+    for v in dir(DeviceUnpairingResultStatus)
+    if "_" not in v and isinstance(getattr(DeviceUnpairingResultStatus, v), int)
+}
+
+
+class BleakClientWinRT(BaseBleakClient):
+    """The native Windows Bleak Client.
+
+    Implemented using `winrt <https://pythonnet.github.io/>`_, a package that enables Python developers to access
```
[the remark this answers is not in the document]
hbldh: 👍🏻

bleak/backends/winrt/client.py
```diff
@@ -0,0 +1,824 @@
...
+
+    async def disconnect(self) -> bool:
+        """Disconnect from the specified GATT server.
+
+        Returns:
+            Boolean representing if device is disconnected.
+
+        """
+        logger.debug("Disconnecting from BLE device...")
+        # Remove notifications. Remove them first in the BleakBridge and then clear
```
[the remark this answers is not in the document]
hbldh: 👍🏻

bleak/backends/winrt/scanner.py
hbldh: How did I miss this? Of course.

bleak/backends/winrt/scanner.py
hbldh: 👍🏻

bleak/backends/winrt/scanner.py
hbldh: Yes. Made this change in the `dotnet` backend as well.

bleak/backends/winrt/service.py
hbldh: Hm. Not sure. Will delete it, since it is not used. Made this change in the `dotnet` backend as well.

The getting of characteristics is async and cannot be done here in the constructor anyway, I think.

setup.py
hbldh: Yes, I thought that it would allow users to install winrt backend at the same time as bleak if they desired:

```
pip install bleak[winrt]
```
This will play nicely with the "if winwrt is installed, use that", will it not?

bleak/backends/winrt/client.py
```diff
@@ -0,0 +1,824 @@
...
+        # remaining notifications in Python as well.
+        for handle, (fcn, fcn_token) in list(self._notification_callbacks.items()):
+            char = self.services.get_characteristic(handle)
+            char.obj.remove_value_changed(fcn_token)
```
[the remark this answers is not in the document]
hbldh: Hm, yes. I think I will leave that for now and create a separate issue for it later.

bleak/backends/winrt/client.py
```diff
@@ -0,0 +1,824 @@
...
+            self._disconnect_events.append(event)
+            try:
+                self._requester.close()
+                await asyncio.wait_for(event.wait(), timeout=10)
+            finally:
+                self._disconnect_events.remove(event)
+
+        return True
+
+    async def is_connected(self) -> bool:
```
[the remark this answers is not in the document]
hbldh: :thumbs:

bleak/backends/winrt/client.py
```diff
@@ -0,0 +1,824 @@
...
+        # Return the Service Collection.
+        if self._services_resolved:
+            return self.services
+        else:
+            logger.debug("Get Services...")
+            services_result = await self._requester.get_gatt_services_async(
+                BluetoothCacheMode.CACHED if use_cached else BluetoothCacheMode.UNCACHED
+            )
+
+            if services_result.status != GattCommunicationStatus.SUCCESS:
```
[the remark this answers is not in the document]
hbldh: Yes, it could be made into a `assert_reply` similar to what you did in the BlueZ backend. Will create a separate issue for it once this PR is merged.

bleak/backends/winrt/client.py
```diff
@@ -0,0 +1,824 @@
...
+
+        """
+        if self._requester:
+            return (
+                self._requester.connection_status == BluetoothConnectionStatus.CONNECTED
+            )
+        else:
+            return False
+
+    async def pair(self, protection_level=None, **kwargs) -> bool:
```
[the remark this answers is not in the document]
hbldh: 👍🏻

bleak/backends/winrt/client.py
hbldh: Yes. Added complete namespace both here and in `dotnet` backend.

bleak/backends/winrt/client.py
hbldh: Made it part of kwargs. It is still set to `False` by default and does not use not use the `self._use_cached` value. This is a bit problematic, but I *think* the users could be fooled if it is defaulted as `True` (as `self._use_cached` is by default)?

bleak/backends/winrt/client.py
hbldh: 👍🏻

bleak/backends/winrt/client.py
hbldh: 👍🏻

bleak/backends/winrt/scanner.py
hbldh: Yes, that might work; did not think of that. In winrt you get a different memory address for `sender`, so I just let it slide. I *think* this will be equal all the time anyway...

hbldh: No, it did not work. I will leave it as it is. Maybe it works due to the wrong reasons in the `dotnet` backend...

bleak/backends/winrt/client.py
hbldh: No, it is not needed. I had it there so I could call `del` on it previously. This to try to remedy all possible memory issues...
Will try to remove this.

bleak/backends/winrt/client.py
dlech: Since we have lots of custom hardware users, False is definitely the best default.

setup.py
dlech: Yes, that makes sense.

dlech: For that matter, we could also add `pythonnet` here since it is possible to install it from source in Python 3.9.

bleak/__init__.py
```diff
@@ -77,8 +77,21 @@
...
         )
 
-    from bleak.backends.dotnet.scanner import BleakScannerDotNet as BleakScanner  # noqa
-    from bleak.backends.dotnet.client import BleakClientDotNet as BleakClient  # noqa
+    # If the winrt package is installed, assume that the user has opted to use that backend
+    # instead of the pythonnet/BleakBridge implementation.
+    try:
+        from bleak.backends.winrt.scanner import (
+            BleakScannerWinRT as BleakScanner,
+        )  # noqa
```
dlech, line 85: I think it is nice to give the specific warning/error that is being supressed

    # noqa: F401

setup.py
hbldh: 👍🏻

bleak/__init__.py
hbldh: 👍🏻

sander76:
Looking at this with much interest. Is it your intention to eventually replace the pythonnet implementation for this WinRT one ?

sander76:
@hbldh Thanks for the response. I've been watching Winrt for some time now, and it took quite a while for it to support python3.8. Finally, after some begging, it came, but it doesn't give you much confidence that it will be supported on the longer term, even though it might be the better alternative..

Issue #175, opened 2020-04-07
win10 local_name is 'Unknown'


vjani68:
* bleak version: 0.6.1
* Python version: 3.6
* Operating System: Win10
* BlueZ version (`bluetoothctl -v`) in case of Linux: 

### Description

I use scanner get_discovered_devices but I see device name's 'Unknown' always

### What I Did

I saw scanner.py in the 135. line:            
 if (
                not new_device.name
                and event_args.BluetoothAddress in self._scan_responses
            ):
                new_device.name = self._scan_responses[
                    event_args.BluetoothAddress
                ].Advertisement.LocalName
            found.append(new_device)
this new_device.name never nothing because in the 160. line:
return BLEDevice(
            bdaddr, local_name, event_args, uuids=uuids, manufacturer_data=data
BLEDevice init local_name = 'Unknown

dlech:
I haven't had time to dig into this, but I noticed recently (using Qt bluetooth fwiw) that on Windows, the advertising data and scan response are separated. In other words, there are multiple advertisements that need to be put together in order to get the full response. See [ BluetoothLEAdvertisementType](https://docs.microsoft.com/en-us/uwp/api/windows.devices.bluetooth.advertisement.bluetoothleadvertisementtype).

For example, on the devices I have the list of services and manufacturer specific data are in the ` ConnectableUndirected ` (`ADV_IND`) packet and the local name is in the ` ScanResponse ` (`SCAN_RSP`) packet.

I think bleak needs to be fixed to put the information these two packets (and probably other advertisement types for other devices too) together in a single "response" so that windows behavior matches other platforms.

vjani68:
I am using WX. If I use discover than ok. 
the problem is that this line never run
**new_device.name = self._scan_responses[
event_args.BluetoothAddress
].Advertisement.LocalName
found.append(new_device)**

**if (
not new_device.name
and event_args.BluetoothAddress in self._scan_responses
):**
If i change this
if (
new_device.name
and event_args.BluetoothAddress in self._scan_responses
):
then ok

hbldh:
I see the problem. I will try to fix so that it does not say `Unknown` before testing if it is empty or None.

pjbosco:
This requires an update to the constructor of BLEDevice. If the 'name' parameter is None, the BLEDevice.name gets set to 'Unknown'. 

So long as BLEDevice never accepts None as a name, the code in the following branch in the dotnet scanner.py will never execute:
```
            if (
[...]

PR #458, opened 2021-02-22
Fix for the 'Unknown' name in scan results on Windows 10. 

pjbosco:
Updated `BLEDevice` constructor to remove setting self.name to `Unknown` if not name. 
Updated `BLEDevice.__str__` conditional to check not self.name instead of equality to `Unknown` .
Updated second return statement in `BLEDevice.__str__` to use `self.name` or `Unknown`.

Updated changelog

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -31,6 +31,7 @@ Fixed
 * Fixed ``BaseBleakClient.services_resolved`` not reset on disconnect on BlueZ
   backend. Merged #401.
 * Fixed RSSI missing in discovered devices on macOS backend. Merged #400.
+* Fixed scan result shows 'Unknown' name of the ``BLEDevice``. Fixes #371.
 
 
 `0.10.0`_ (2020-12-11)
--- a/bleak/backends/device.py
+++ b/bleak/backends/device.py
@@ -27,7 +27,7 @@ class BLEDevice(object):
         #: The Bluetooth address of the device on this machine.
         self.address = address
         #: The advertised name of the device.
-        self.name = name if name else "Unknown"
+        self.name = name
         #: The OS native details required for connecting to the device.
         self.details = details
         #: RSSI, if available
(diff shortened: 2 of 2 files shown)


hbldh:
@pjbosco Looking good. However, this does not include the changes you proposed in #175; shouldn't that be added to this PR as well?

pjbosco:
I found that changing the dotnet scanner.py was unnecessary given the changes to BLEDevice.__str__(). Essentially, the changes to scanner.py would have set BLEDevice.name itself to "Unknown" rather than simply allowing it to be None or empty string and dealing with the presentation in __str__(). I think the better approach is for BLEDevice.name to reflect physical reality; if we didn't find a name, we don't make the assumption that it is "Unknown".

dlech:
I think this is a step in the right direction.

Perhaps getting a bit off topic, but on BlueZ, we can never get a name of `Unknown` since BlueZ will give the Bluetooth address as the name if there is no name. So to make things more consistent cross-platform, perhaps we should do the same here?

hbldh:
It does? In that case, we should try to change the current `Unknown` default to that then.

dlech:
> It does?

Yes, I noticed this when working on the `dbus-next-2` branch.

pjbosco:
> It does? In that case, we should try to change the current `Unknown` default to that then.

Keep in mind, for the dotnet scanner to get and set the actual name from the ScanResponse `not BLEDevice.name` must resolve to `True`. If we set BLEDevice.name to anything that breaks that resolution in BLEDevice constructor, dotnet scanner will never set the correct name. 

See https://github.com/pjbosco/bleak/blob/cb8fd75821bb87013ce15ffd86bceccc093d67f2/bleak/backends/dotnet/scanner.py#L212

Perhaps if the ScanResponse does not have a local name, set name to the bluetooth address.
For example the following:
```
            if (
                not new_device.name
                and event_args.BluetoothAddress in self._scan_responses
            ):
                new_device.name = self._scan_responses[
                    event_args.BluetoothAddress
                ].Advertisement.LocalName
```

Would become:
```
            if (
                not new_device.name
                and event_args.BluetoothAddress in self._scan_responses
            ):
                new_device.name = self._scan_responses[
                    event_args.BluetoothAddress
                ].Advertisement.LocalName or event_args.BluetoothAddress
```

I think this would solve the broken windows behavior while bringing the behavior in line with BlueZ.

PR #483, opened 2021-03-16
updating docstring for dotnet.client.BleakClientDotNEt connect method

sander76:
I guess I prefer behaviour change, but for now I aligned the docstring better with the current behaviour.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,7 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 Added
 ~~~~~
 
+* Updated ``dotnet.client.BleakClientDotNet`` connect method docstring.
 * Added ``AdvertisementServiceData`` in BLEDevice in macOS devices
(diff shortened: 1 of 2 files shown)


dlech:
Thanks!

> I guess I prefer behaviour change

Is there an open issue to discuss this?

hbldh:
Yes, I am curious as well. @sander76 open an issue or discussion and propose the behaviour change you would like.

sander76:
@hbldh @dlech Well, I don't know whether it's worth a separate issue (and don't take this as criticism as I think this lib rocks). In this particular case I would prefer raising an exception (like the BleakError) when connection fails. I see this in other places as well, returning either None, an object or a True or False value. I prefer wrapping functions like these in a `try except` block. This way I don't need to check for None and type checking is easier too (return the `type` instead of the `Optional[type]`. If you like I can still put this in an issue.

dlech:
I think at this point, all of the backends will raise `BleakError` (or some other error) if the connection fails. Returning `True` is still done for backwards compatibility.

PR #495, opened 2021-03-28
Replace get_discovered_devices with discovered_devices

### LINKED ISSUE #489: get_discovered_devices() is not actually async
Technically speaking, `BleakScanner.get_disovered_devices()` does not need to be async since none of the implementations await anything.

However, changing it would be a major breaking change. We could probably lessen the impact by doing something similar to https://github.com/hbldh/bleak/commit/bc6da1e81a677ad9065936f52781bc15693732a8 to make it not quite so breaking if we decide it is worth making the change (for some time, it would return an object that is both a list and a future that gives a warning about the change).

dlech:
This adds a new `BleakScanner.discovered_devices` property to replace the `BleakScanner.get_discovered_devices()` async method.

A deprecation warning is added to `BleakScanner.get_discovered_devices()` and the documentation for that method.

Fixes #489

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,16 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 `Unreleased`_
 -------------
 
+Added
+~~~~~
+
+* Added ``BleakScanner.discovered_devices`` property.
+
+Changed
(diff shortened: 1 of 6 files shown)


hbldh (approved):
I see no problems with this. It is a good change.

PR #496, opened 2021-03-28
Fix linux_autodoc_mock_import in docs/conf.py

dlech:
This expects the module name rather than the package name, so we have to use an underscore instead of a dash.


hbldh (approved):
This also needs to be adjusted for `winrt` when we finally build with Python 3.9. I think I need to write a `.readthedocs.yaml` and lock down Python version to avoid potential future breakage.

PR #505, opened 2021-04-03
Correct and improve Philips Hue example

eb3095:
The Hue example is missing examples of how to use color. I pulled the logic from another repo on the conversion to the correct format and added a reference line for it. I also added my own repo since I am using this and may expand in the future.

The color characteristic is wrong, its 0005, not 0004, 0004 appears to be color temperature (soft-white / blue-white). I haven't played with this or I would of expanded on it, so I just left the line for anyone looking for it.


hbldh:
Thank you for this! I only have a bulb with no colour, so I could not test the other characteristics. I will modify the example script you sent to be able to handle both colour and non-colour Hue lamps.

PR #520, opened 2021-04-20
change typing on data parameter in `write_gatt_char` and `write_gatt_descriptor`

sander76:
The write_gatt_char and write_gatt_descriptor methods allow for both a bytes and bytearray object. Typing did only allow for a bytearray.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,7 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 Added
 ~~~~~
 
+* Change typing of data parameter to ``Union[bytes, bytearray]``
 * WinRT backend added
 * Added ``BleakScanner.discovered_devices`` property.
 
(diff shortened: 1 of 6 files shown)


dlech:
Have you tested this on all operating systems? I'm pretty sure that `dbus-next` (BlueZ backend) will complain if anything other than `bytearray` is used currently.

Looking at the bigger picture, ideally these methods should accept anything bytes-like, not just `bytes` and `bytearray`. But we would probably need to add a cast in the code, at least for the `dbus-next` backend.

From what I have read, it seems like just `bytes` is used to mean anything bytes-like. There is also [typing.ByteString](https://docs.python.org/3/library/typing.html#typing.ByteString) but it is one of those places that mentions just using `bytes` instead.

Related: https://github.com/python/typing/issues/593

sander76:
@dlech No I haven't. Being lazy I thought it just might work (it also mentioned both bytes and bytearray). Unfortunately I am not in the position to test this on anything other than windows10.

dlech:
No problem. I looked at the BlueZ backend and we are already casting to `bytes`, so it should be fine. I still think it would be nice to also include `memoryview` in the union (or use `typing.ByteString`, or just use `bytes`).

PR #521, opened 2021-04-21
Fix read and set of D-Bus properties in BleakClientBlueZDBus.pair()

bojanpotocnik:
When checking if device is already paired in `BleakClientBlueZDBus`
`pair()` method, result of call reading the Paired property
reply.body[0] is object type `dbus_next.signature.Variant` and always
evaluated to `True`, even if the actual value represented was `False`.
Test:
```python
>>> reply.body[0]
<dbus_next.signature.Variant ('b', False)>
>>> bool(reply.body[0])
True
>>> reply.body[0].value
False
>>> bool(reply.body[0].value)
False
```
Moreover, when setting the Trusted property, the following exception
was raised:
```plain
    dbus_next.errors.SignatureBodyMismatchError: DBus VARIANT type "v"
    must be Python type "Variant", got <class 'bool'>
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -32,6 +32,7 @@ Fixed
 * Fixed ``linux_autodoc_mock_import`` in ``docs/conf.py``.
 * Minor fix for disconnection event handling in BlueZ backend. Fixes #491.
 * Corrections for the Philips Hue lamp example. Merged #505
+* Fixed BleakClientBlueZDBus.pair() method always returning True. Fixes #503.
 
 
 `0.11.0`_ (2021-03-17)
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -499,7 +499,10 @@ class BleakClientBlueZDBus(BaseBleakClient):
             )
         )
         assert_reply(reply)
-        if reply.body[0]:
+        if reply.body[0].value:
+            logger.debug(
+                f"BLE device @ {self.address} already paired with {self._adapter}"
+            )
             return True
 
         # Set device as trusted.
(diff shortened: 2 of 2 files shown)


dlech:
Thanks! Could you please add a changelog entry for this?

bojanpotocnik:
Done. In the meantime, while I was browsing through issues for some other pairing-related stuff, I discovered that https://github.com/hbldh/bleak/discussions/503 also found out this, so I added the reference.

dlech:
Awesome, thanks again!

asafs932:
> When checking if device is already paired in `BleakClientBlueZDBus`
> `pair()` method, result of call reading the Paired property
> reply.body[0] is object type `dbus_next.signature.Variant` and always
> evaluated to `True`, even if the actual value represented was `False`.
> Test:
> 
> ```python
> >>> reply.body[0]
> <dbus_next.signature.Variant ('b', False)>
> >>> bool(reply.body[0])
> True
> >>> reply.body[0].value
> False
> >>> bool(reply.body[0].value)
> False
> ```
> 
> Moreover, when setting the Trusted property, the following exception
> was raised:
> 
> ```
>     dbus_next.errors.SignatureBodyMismatchError: DBus VARIANT type "v"
>     must be Python type "Variant", got <class 'bool'>
> ```

i tried to do the pairing sequence with my device, Hence i commented out the lines:
"
        # if reply.body[0]:
        #     return True
" 
but i get like you said this error:
"error DBus VARIANT type "v" must be Python type "Variant", got <class 'bool'>."

my code:
```
            client = BleakClient(device_address)
            try:
                await client.connect()
                await client.pair(123456)
                is_connected = True
                logger.info("connection success")
            except Exception as e:
                logger.info("connection fails with error {}".format(e))
```

How is it possible to overcome this pairing issue? my configured passkey is "123456" and i already verified it works using the nrf app.
more info: the central device is a LINUX computer and the peripheral device which i try to connect is an esp32 device.

bojanpotocnik:
@asafs932 bleak currently only supports Just Works pairing method. For passkey pairing, try https://github.com/hbldh/bleak/pull/523 and ask (me) there if you encounter any problem. Example is also provided in scope of that PR - you need to use callback to provide the pin instead of providing it directly.

Issue #513, opened 2021-04-15
Dynamically select default adapter


lopsided98:
* bleak version: 0.11.0
* Python version: 3.8.8
* Operating System: Linux
* BlueZ version (`bluetoothctl -v`) in case of Linux: 5.57

### Description

If not specified, the adapter name on Linux currently defaults to `hci0`. Not all Linux systems have an `hci0` adapter (on one of mine, the only adapter is `hci1`). It would be useful to have a way to dynamically select the default adapter name from those actually available on the system, rather than hardcoding it. In 0.10.0, this was actually possible by passing `adapter=None` to `BleakScanner`, but I suspect this feature might have been unintentional. The [relevant code](https://github.com/hbldh/bleak/blob/v0.10.0/bleak/backends/bluezdbus/scanner.py#L16) was removed in the switch from `txdbus` to `dbus-next`.

PR #524, opened 2021-04-22
Dynamically select default BlueZ adapter if not provided

bojanpotocnik:
If specific adapter is not provided as a parameter, default adapter is
selected automatically as a first adapter out of all powered adapter in
the system, instead of using hardcoded "hci0".

Additional logic to only choose powered adapters is added for two
convenience reasons:
- If adapter is powered down but selected for operation, Bleak will not
  power it on. Consequently, `org.bluez.Error.NotReady` error will occur
  anyway when trying to use such adapter.
- If user has multiple adapters present on the system but currently uses
  only some while others are powered off, it most probably wants to use
  one of the active adapters.

Solves #513


bleak/backends/bluezdbus/utils.py
```diff
@@ -51,3 +53,53 @@ def extract_service_handle_from_path(path):
...
+    """
+    bus_provided = bus and bus.connected
+    if not bus_provided:
+        bus = await MessageBus(bus_type=BusType.SYSTEM).connect()
+
+    reply = await bus.call(
+        Message(
+            destination=defs.BLUEZ_SERVICE,
+            path="/",
+            member="GetManagedObjects",
```
dlech, line 80: This is a very expensive operation. Instead of opening a new D-Bus connection and calling this method, could we make use of the existing copy of this information in the BleakScanner?

bojanpotocnik: I have added 2 commits (those will be squashed later) which solve this. What do you think? In the afternoon I will also modify `client.py`, which has 50 lines of duplicate code (this and adding signal handlers).

bleak/backends/bluezdbus/scanner.py
```diff
@@ -77,8 +78,31 @@ def __init__(self, **kwargs):
     async def start(self):
         self._bus = await MessageBus(bus_type=BusType.SYSTEM).connect()
 
+        # Find the HCI device to use for scanning and get cached device properties
+        reply = await self._bus.call(
```
dlech, line 82: Moving this before attaching the signal listeners will cause a race condition where signals can be missed, so unfortunately this is not going to work.

dlech:
We have made some changes recently that should help unblock this PR.  There is now a [BlueZManager](https://github.com/hbldh/bleak/blob/726c579bbdd263b812a0736811b138459a30e4b3/bleak/backends/bluezdbus/manager.py#L210) with a global instance that we can use to get the current BlueZ D-Bus properties at any time (well almost any time - it must be in an async function). We should be able to add the proposed `get_default_adapter` as a method to this class and use it in `BleackScanner.start()` and `BleakClient.connect()` as needed.

dlech:
I made the required changes to get this merged. Thanks for the contribution!

PR #525, opened 2021-04-23
Adding mtu_size client

es617:
Adding a simple method to the macOS client to get the current ATT MTU size.
The ATT MTU size includes the ATT header, while CoreBluetooth returns the maximum ATT payload size, hence the +3 in the getMtuSize method.

### DIFF
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -1046,3 +1046,33 @@ class BleakClientBlueZDBus(BaseBleakClient):
                     disconnecting_event = self._disconnecting_event
                     if disconnecting_event:
                         task.add_done_callback(lambda _: disconnecting_event.set())
+
+
+    async def get_mtu_size(self, char_specifier: Union[BleakGATTCharacteristicBlueZDBus, int, str, UUID]):
+        if not self.is_connected:
+            raise BleakError("Not connected")
+
+        if not isinstance(char_specifier, BleakGATTCharacteristicBlueZDBus):
+            characteristic = self.services.get_characteristic(char_specifier)
+        else:
+            characteristic = char_specifier
+
+        member = "AcquireWrite"
+
+        if "notify" in characteristic.properties:
+            member="AcquireNotify"
+
+        reply = await self._bus.call(
+            Message(
+                destination=defs.BLUEZ_SERVICE,
+                path=characteristic.path,
+                interface=defs.GATT_CHARACTERISTIC_INTERFACE,
+                member=member,
+                signature="a{sv}",
+                body=[{}],
+            )
+        )
+
+        assert_reply(reply)
+        return reply.body[1]
+
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -98,6 +98,9 @@ class BleakClientBlueZDBus(BaseBleakClient):
             bluez_version[0] == 5 and bluez_version[1] >= 48
         )
 
+        # set default MTU size
+        self._mtu_size = 23
+
     # Connectivity methods
 
     async def connect(self, **kwargs) -> bool:
@@ -334,6 +337,11 @@ class BleakClientBlueZDBus(BaseBleakClient):
             # Get all services. This means making the actual connection.
             await self.get_services()
 
+            # Get the MTU size, assuming the server has at least one
+            # notify or write-without-response characteristic
+            # If not, we use default ATT MTU size of 23
+            self._mtu_size = await self.get_mtu_size(self.services)
+
             return True
         except BaseException:
             await self._cleanup_all()
@@ -569,6 +577,11 @@ class BleakClientBlueZDBus(BaseBleakClient):
         return self._DeprecatedIsConnectedReturn(
             False if self._bus is None else self._properties.get("Connected", False)
         )
+
+    @property
+    def mtu_size(self) -> bool:
+        """Get ATT MTU size for active connection"""
+        return self._mtu_size
 
     # GATT services methods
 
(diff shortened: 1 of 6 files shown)


dlech:
Could you also add this function for the other backends?

es617:
@dlech I added support for the other three backends. A few notes:

1) The windows implementation seems pretty straightforward. The only question is whether the value returned is the ATT MTU size or the max ATT payload size. Reading the [specs](https://docs.microsoft.com/en-us/uwp/api/windows.devices.bluetooth.genericattributeprofile.gattsession.maxpdusize?view=winrt-19041#Windows_Devices_Bluetooth_GenericAttributeProfile_GattSession_MaxPduSize), it should be the ATT MTU size (including the header). Unfortunately, I haven't been able to test on Windows yet.

1b) If I get to test on Windows, how do I test the two different win backends?

2) The BlueZ implementation is not the greatest, but it's the only option available as far as I can see. It will only work if the server has a characteristic with "notify" or "write wo response" property. 

3) I tried to keep the API the same among the different backends; in reality, only the BlueZ backend requires a "characteristic" to be passed. The other backends will just ignore that argument. Not sure what is the recommended approach for BLEAK in these situations. Please advise!

dlech:
Thanks for the effort!

I think for Windows, right now the backend depends on the Python runtime version, so if you don't want to install different versions of Python, you probably just have to edit Bleak to be sure to load the one you want.

On BlueZ, I agree this is probably the best we can do. Instead of making the user have to provide a characteristic, would it be possible to go through all of the cached characteristics and chose the "best" one? I think the "best" one would be one where we can call `AquireWrite` (write without response property set) and fall back to `AquireNotify` if no characteristic is available for `AquireWrite`. (Devices could start sending notifications plus it will fail if notifications are already enabled, so it would be best to avoid `AquireNotify`.)

Also, we should close the file descriptor returned by these D-Bus methods, otherwise it will probably block other operations.

It would also be good to cache the returned MTU value on the BlueZ backend so that we only have to call the complex lookup once per connection.

dlech:
Another idea: 

Make this a property instead of an async method. On Windows and Mac, this property will *just work*. On Linux, this property will return an error unless the user first calls a special method to fetch and cache the MTU value. This way, the user has full control over which characteristic to use (and possibly a choice of using AquireWrite or AquireNotify). Then if BlueZ ever adds a proper MTU property, we won't be stuck with an unnecessary async method in Bleak.

es617:
Thanks for the feedback!

I still need to run some tests, but here for early review. 
- Now using a property for all the clients
- BlueZ has default mtu size set to 23. After service discovery we try to get the actual MTU size. If that fails, we stick with 23. Not sure whether to return an error or lie with a smaller (conservative) MTU size. In any case, this behavior should be documented.

To be tested:
- what happens when we call get_mtu_size prior to bonding and we have encrypted characteristics. 
- Windows clients

es617:
Pushed another round of changes.
Unfortunately, the "get mtu size" method can't be called right after connection. If the device is not bonded calling AcquireNotify or AcquireWrite on an encrypted characteristic causes the device to disconnect.

The only option we have is to let the user call the method manually (now called discover_mtu_size, to distinguish from "get"), passing the UUID of a characteristic and the preferred property to use for the discovery.

bleak/backends/bluezdbus/client.py
```diff
@@ -1046,3 +1058,41 @@ def _parse_msg(self, message: Message):
...
+            characteristic = self.services.get_characteristic(char_specifier)
+        else:
+            characteristic = char_specifier
+
+        if char_property is "notify":
+            member = "AcquireNotify"
+        elif char_property is "write-without-response":
+            member = "AcquireWrite"
+        else:
+            raise BleakError(
```
dlech, line 1080: This should probably be `ValueError` rather than `BleakError`.

dlech (commented):
This looks good to me. Just one minor comment.

It would be nice to have an example program too, but we can save that for a separate PR.

dlech:
I also just noticed that we are setting `negotiate_unix_fd` based on the BlueZ version, so I'm wondering if trying to close the file descriptor will cause a problem when `negotiate_unix_fd=False`.

dlech:
I would be happy with BlueZ just returning 23 for now along with logging a warning that it is not properly implemented.

Then we can look at fixing dbus-next and/or adding an mtu property to BlueZ to eventually get this working right.

es617:
thanks! I just pushed the simplified implementation for bluez. I'll test on windows and then this should be good to go

bleak/backends/bluezdbus/client.py
```diff
@@ -570,6 +570,14 @@ def is_connected(self) -> bool:
             False if self._bus is None else self._properties.get("Connected", False)
         )
 
+    @property
+    def mtu_size(self) -> bool:
```
dlech, line 574: return type is `int`, not `bool`

bleak/backends/corebluetooth/client.py
```diff
@@ -125,6 +125,12 @@ def is_connected(self) -> bool:
             False if manager is None else manager.isConnected
         )
 
+    @property
+    def mtu_size(self) -> bool:
```
dlech, line 129: return type should be `int`

bleak/backends/dotnet/client.py
```diff
@@ -305,6 +305,11 @@ def is_connected(self) -> bool:
             else self._requester.ConnectionStatus == BluetoothConnectionStatus.Connected
         )
 
+    @property
+    def mtu_size(self) -> bool:
```
dlech, line 309: return type should be `int`

bleak/backends/winrt/client.py
```diff
@@ -280,6 +280,11 @@ def is_connected(self) -> bool:
             == BluetoothConnectionStatus.CONNECTED
         )
 
+    @property
+    def mtu_size(self) -> bool:
```
dlech, line 284: return type should be `int`

bleak/backends/winrt/client.py
```diff
@@ -280,6 +280,11 @@ def is_connected(self) -> bool:
...
+        """Get ATT MTU size for active connection"""
+        return self._session.MaxPduSize
```
dlech, line 286: the winrt API uses lower_snake_case, so `max_pdu_size` here

dlech (changes requested):
After making the suggested change (and merging the fix from #537), the winrt backend works.

dlech:
Hmm.... I'm playing with this on Windows more and I just got the wrong MTU (517 instead of 23 as expected for the device I'm using). So I guess that `connect()` returns before the OS has negotiated the MTU. Not sure what to do about that other than document it. 

Windows has a [MaxPduSizeChanged ](https://docs.microsoft.com/en-us/uwp/api/windows.devices.bluetooth.genericattributeprofile.gattsession.maxpdusizechanged) event but I'm not sure that I would trust it to always be called when a device is connected and to only be called once. If it did work this way, we could wait for the event when connecting.

dlech:
After some more experimentation, it seems that the device isn't connecting properly or is becoming disconnected before I read the mtu, so I think we can ignore my previous comment.

es617:
thanks for the feedback and for testing on Windows! 
Could you check if the max_pdu_size (or MaxPduSize) include the ATT payload (like returning 23, instead of 20)? It's not clear from the specs. Just to be sure we are consistent between the different backends.

dlech:
I have confirmed that MaxPduSize does return the actual MTU.

dlech (approved):

es617:
thanks for the approve! I don't see a merge button ("Only those with write access to this repository can merge pull requests."), should someone else merge this?

dlech:
merged, thanks!

PR #533, opened 2021-05-05
During init() support raising a "BLE is unsupported"

epoz:
For some older devices the bluetooth stack does not support BLE.

The error message "Bluetooth device is turned off" can then be confusing, if on that device bluetooth is actually switched on (and working) but the actual problem is that BLE is not supported.


dlech (approved):
Looks good to me, thanks!

epoz:
> Could you run `black` on the code or manually delete the trailing white space on the blank lines?

The patch was submitted using the web-editor of Github directly, I never did a local checkout. Lemme see if I can do a manual edit...

dlech:
Thanks!

Issue #437, opened 2021-02-02
Scanning BLE devices returns empty uuids in metadata (?)


rdagher:
* bleak version: 0.10.0
* Python version: Python 3.8.5
* Operating System: MacOS Catalina 10.15.7
* BlueZ version (`bluetoothctl -v`) in case of Linux: N/A

### Description
Hi,

I wanted to test scanning devices and print the details of the scanned [device](https://github.com/hbldh/bleak/blob/develop/bleak/backends/device.py) (name, metadata, etc.). 

The devices are detected (name, address), nevertheless the metadata field `metadata = {'uuids': [], ...}` does not contain the service uuids which is expected to be a list of one element, in my case `'uuids': ['680c21d9-c946-4c1f-9c11-baa1c21329e7']`.

The reason is I want to filter the scan output, application-side, on the device's service uuid to only report devices of interest.

For the test, I have two BLE devices running (namely DWDBB5 and DWD48C) - scan using a mobile app on Android ([nrfConnect](https://play.google.com/store/apps/details?id=no.nordicsemi.android.mcp&hl=fr&gl=US)):
![nrf-scan-screenshot](https://user-images.githubusercontent.com/11162310/106587419-5c813180-6528-11eb-9cc9-8964811e35cf.jpg)

### What I Did
Attached the test script containing two simple scanning functions.
[ble_scan.py.txt](https://github.com/hbldh/bleak/files/5910483/ble_scan.py.txt)
[ble_scan_log.txt](https://github.com/hbldh/bleak/files/5910485/ble_scan_log.txt)

Here's a snippet of what I get as output:

```python
import asyncio
from bleak import BleakScanner
from bleak.backends.device import BLEDevice

from ppretty import ppretty



def dev_dump(dev:BLEDevice):
    #print(ppretty(dev, show_protected=True, show_static=True, show_properties=True))
    print(f'device {dev}:')
    print(f'\taddr = {dev.address}')
    print(f'\tname = {dev.name}')
    print(f'\trssi = {dev.rssi}')
    print(f'\tdetails = {dev.details}')
    print(f'\tmetadata = {dev.metadata}')

    print(f'>> Services (Mac) = {dev.details.services}')
    print(ppretty(dev.details.services, show_protected=True, show_static=True, show_properties=True))

async def basic_scan():
    devices = await BleakScanner.discover()
    for d in devices:
        dev_dump(d)


async def adv_scan(scan_duration:int, service_uuid:str):
    # scan for devices
[...]

PR #534, opened 2021-05-12
Fix missing delegate in BLEDevice metadata

### LINKED ISSUE #448: Device returned by detection callback is not properly initialized on corebluetooth
* bleak version:0.10.0
* Python version:3.7.2
* Operating System:MacOS Catalina 10.15.2

### Description

The device returned by the detection callback is missing "delegate" which is needed by the ctor of BleakClientCoreBluetooth

### What I Did

I work around the issue by not using the callback and repeatedly calling get_discovered_devices instead.

Sample (modified from bleak/examples/detection_callback.py):
```python
import asyncio
from bleak import BleakClient
from bleak.backends.device import BLEDevice
from bleak.backends.scanner import AdvertisementData
import logging

logging.basicConfig()

dev = None

def simple_callback(device: BLEDevice, advertisement_data: AdvertisementData):
    global dev
    dev = device
    print(device.address, "RSSI:", device.rssi, advertisement_data)


async def run():
    scanner = BleakScanner()
    scanner.register_detection_callback(simple_callback)

    while True:
        await scanner.start()
        await asyncio.sleep(5.0)
        await scanner.stop()

        global dev
        if dev is not None:
            async with BleakClient(dev) as client:
                print("Connected") 


loop.run_until_complete(run())
```

Command line:
```
python3 detect_sample.py
```
Output:
```
  File "detect_sample.py", line 35, in <module>
    loop.run_until_complete(run())
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/asyncio/base_events.py", line 584, in run_until_complete
    return future.result()
  File "detect_sample.py", line 30, in run
    async with BleakClient(dev) as client:
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/bleak/backends/corebluetooth/client.py", line 51, in __init__
    self._central_manager_delegate = address_or_ble_device.metadata["delegate"]
KeyError: 'delegate'
```

dlech:
This fixes #448.

Also include a code cleanup in the same file while we are touching this.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,8 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 Fixed
 ~~~~~
 
+* Fixed ``KeyError`` when trying to connect to ``BLEDevice`` from advertising
+  data callback on macOS. Fixes #448.
 * Handling of undetected devices in ``connect_by_bledevice.py`` example. Fixes #487.
 * Added ``Optional`` typehint for ``BleakScanner.find_device_by_address``.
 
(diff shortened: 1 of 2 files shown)


dlech:
Probably also fixes https://github.com/hbldh/bleak/issues/437

PR #541, opened 2021-05-18
Mac fixes

dlech:
These are just a few small fixes I noticed while reading the code.

### DIFF
--- a/bleak/backends/corebluetooth/PeripheralDelegate.py
+++ b/bleak/backends/corebluetooth/PeripheralDelegate.py
@@ -80,7 +80,7 @@ class PeripheralDelegate(NSObject):
 
     async def discoverServices(self, use_cached=True) -> [CBService]:
         event = self._services_discovered_event
-        if event.is_set() and (use_cached is True):
+        if event.is_set() and use_cached:
             return self.peripheral.services()
 
         event.clear()
@@ -93,7 +93,7 @@ class PeripheralDelegate(NSObject):
     async def discoverCharacteristics_(
         self, service: CBService, use_cached=True
     ) -> [CBCharacteristic]:
-        if service.characteristics() is not None and use_cached is True:
+        if service.characteristics() is not None and use_cached:
             return service.characteristics()
 
         sUUID = service.UUID().UUIDString()
@@ -106,7 +106,7 @@ class PeripheralDelegate(NSObject):
     async def discoverDescriptors_(
         self, characteristic: CBCharacteristic, use_cached=True
     ) -> [CBDescriptor]:
-        if characteristic.descriptors() is not None and use_cached is True:
+        if characteristic.descriptors() is not None and use_cached:
             return characteristic.descriptors()
 
         cUUID = characteristic.UUID().UUIDString()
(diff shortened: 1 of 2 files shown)


bleak/backends/corebluetooth/PeripheralDelegate.py
```diff
@@ -196,7 +197,7 @@ async def startNotify_cb_(
         event = self._characteristic_notify_change_events.get_cleared(c_handle)
         self.peripheral.setNotifyValue_forCharacteristic_(True, characteristic)
         # wait for peripheral_didUpdateNotificationStateForCharacteristic_error_ to set event
-        # await event.wait()
+        await event.wait()
```
hbldh, line 200: Why was this event waiting commented out? It feels like it must have been done due to some reason once...

hbldh (approved):

bleak/backends/corebluetooth/PeripheralDelegate.py
dlech: As mentioned in the commit message, it was commented out in f3bead7 and was unrelated to that change, which was about fixing RSSI.

dlech: @Carglglz, you remember why this was done?

dlech: Bleak objects cannot be used from threads other than the one where they were created (not needing to do this is one of the benefits of using asyncio). There is `call_soon_threadsafe()` to schedule something on an event loop from a different thread if needed.

Could you open a new issue for the hang when stopping notifications with a reproducible code example?

Carglglz: > Also how does peripheral_didUpdateNotificationStateForCharacteristic_error_ to set event sets the event flag to True if event is not passed to self.peripheral.setNotifyValue_forCharacteristic_ ?

Ok, I remember now how it is done

```
def did_update_notification_for_characteristic(
        self, peripheral: CBPeripheral, characteristic: CBCharacteristic, error: NSError
    ):
        cUUID = characteristic.UUID().UUIDString()
        c_handle = characteristic.handle()
        if error is not None:
            raise BleakError(
                "Failed to update the notification status for characteristic {}: {}".format(
                    cUUID, error
                )
            )
        logger.debug("Character Notify Update")

        event = self._characteristic_notify_change_events.get(c_handle)
        if event:
            event.set()
        else:
            logger.warning(
                "Unexpected event didUpdateNotificationStateForCharacteristic"
            )

def peripheral_didUpdateNotificationStateForCharacteristic_error_(
        self, peripheral: CBPeripheral, characteristic: CBCharacteristic, error: NSError
    ):
        logger.debug("peripheral_didUpdateNotificationStateForCharacteristic_error_")
        self._event_loop.call_soon_threadsafe(
            self.did_update_notification_for_characteristic,
            peripheral,
            characteristic,
            error,
        )
```

I see now that ``self._event_loop.call_soon_threadsafe`` could  be the reason why it was blocking. I was handling notifications in another thread with a new event loop. I will test this changes , see If it is fixed or what to do to avoid commenting out ``await event.wait()`` and works with threads too.

Carglglz: > Could you open a new issue for the hang when stopping notifications with a reproducible code example?
Yes, although the reproducible code example may be a little too complicated. 

See https://github.com/Carglglz/bleico, 

https://github.com/Carglglz/bleico/blob/1f0907a8425f2db4ffb10666b87ec7c7fc63ba22/bleico/systrayicon.py#L964

Here I'm trying to start and stop notifications in a GUI application (systray icon to be precise)... I may be a better way to do this but at that moment I didn't find any.

PR #543, opened 2021-05-19
Add hack for getting MTU on BlueZ backend.

dlech:
BlueZ doesn't have a good way to get the negotiated MTU, but we can get it by abusing existing D-Bus methods.

Since the workaround is async and could cause unwanted side effects for some devices, it opt-in as seen in the example code.

### DIFF
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -81,6 +81,9 @@ class BleakClientBlueZDBus(BaseBleakClient):
         # used to ensure device gets disconnected if event loop crashes
         self._disconnect_monitor_event: Optional[asyncio.Event] = None
 
+        # used to override mtu_size property
+        self._mtu_size: Optional[int] = None
+
         # get BlueZ version
         p = subprocess.Popen(["bluetoothctl", "--version"], stdout=subprocess.PIPE)
         out, _ = p.communicate()
@@ -137,8 +140,7 @@ class BleakClientBlueZDBus(BaseBleakClient):
 
         # Create system bus
         self._bus = await MessageBus(
-            bus_type=BusType.SYSTEM,
-            negotiate_unix_fd=self._write_without_response_workaround_needed,
+            bus_type=BusType.SYSTEM, negotiate_unix_fd=True
         ).connect()
 
         try:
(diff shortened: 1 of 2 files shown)


dlech:
FYI, @es617.

bleak/backends/bluezdbus/client.py
```diff
@@ -570,13 +572,56 @@ def is_connected(self) -> bool:
...
+        but before accessing the ``mtu_size`` property.
+        """
+        # This will try to get the "best" characteristic for getting the MTU.
+        # We would rather not start notifications if we don't have to.
+        try:
+            method = "AcquireWrite"
+            char = next(
+                c
+                for c in self.services.characteristics.values()
+                if "write" in c.properties
```
es617, line 589: If I understand correctly, this might work only with write-without-response
"For client it only works with characteristic that has WriteAcquired property which relies on write-without-response Flag." https://github.com/Vudentz/BlueZ/blob/master/doc/gatt-api.txt

bleak/backends/bluezdbus/client.py
```diff
@@ -570,13 +572,56 @@ def is_connected(self) -> bool:
...
         )
 
+    async def _acquire_mtu(self) -> None:
+        """Acquires the MTU for this device by calling the "AcquireWrite" or
+        "AcquireNotify" method of the first characteristic that has such a method.
+
+        This method only needs to be called once, after connecting to the device
+        but before accessing the ``mtu_size`` property.
+        """
+        # This will try to get the "best" characteristic for getting the MTU.
```
es617, line 582: I was seeing some issues when the characteristic was encrypted and the device was not bonded. Usually, an unsufficient-auth error causes a disconnect. Have you tried that? I don't think there is a workaround, but maybe worth mentioning in the doc, or example?

bleak/backends/bluezdbus/client.py
dlech: You are right. I missed this since the device I was testing with supports both..

bleak/backends/bluezdbus/client.py
dlech: I don't have any devices that use encryption so I haven't seen that. But I can add something to the doc comment.

es617:
looks good to me!

dlech:
Thanks for the review!

PR #544, opened 2021-05-20
Populate rssi field for WinRT backend

devanlai:
Update the WinRT branch to report RSSI using the standard `BLEDevice.rssi` field added in #390.


dlech:
Great, thank you!

PR #554, opened 2021-06-01
Fix propagation of CoreBluetooth errors

rtoijala:
Errors from CBPeripeheralDelegate calls were being raised in the asyncio
event loop, but not in the correct task. In order to reach the user's
code, the error has to be passed through the event to the user call that
is awaiting the event.

Without these changes, e.g. ATT errors sent by the peripheral caused
error messages to be printed, but since the event that the user's call
was waiting on was never set, the user's code was blocked.

The implementation was adapted from https://github.com/hbldh/bleak/pull/209

I have not been able to test with anything older than macOS Big Sur (11.4).

flake8, tox and pytest report loads of unrelated errors, but I did not see any
new errors from these changes.

   guidelines in that document.
    - OS X / macOS >= 10.11

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -40,6 +40,7 @@ Fixed
 * Fixed BleakClientBlueZDBus.pair() method always returning True. Fixes #503.
 * Fixed waiting for notification start/stop to complete in CoreBluetooth backend.
 * Fixed write without response on BlueZ < 5.51.
+* Fixed error propagation for CoreBluetooth events
 
 
 `0.11.0`_ (2021-03-17)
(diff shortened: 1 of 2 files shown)


dlech (commented):
I was noticing this was a problem recently too, so I'm glad you have taken the time to fix it!

Instead of creating our own class based on `asyncio.Event`, how about using `asyncio.Future` instead?

dlech:
This looks great. It looks like there are some formatting issues. Since we need to fix that anyway, it would be nice to change the messages to `f""` strings instead of using `.format()`.

dlech (approved):
I've tested this now. Seems to be working. Thanks again!

PR #557, opened 2021-06-03
Fixed deadlock in CoreBluetooth backend when device disconnects

### LINKED ISSUE #535: Program disconnects before connection established. Locks up Bleak
* bleak version: 0.11.0
* Python version: 3.9
* Operating System: MacOS Big Sir

### Description

Trying to connect and disconnect from an array of devices. trying to connect and then disconnect to make sure devices under test are working properly. I am iterating through the devices and cycling the connect/disconnect N number of times. 
(devices have firmware that sends device to sleep. sometimes wont wake up)

Program would freeze after 20-30 connect / disconnect cycles. 

### What I Did
i added print statements to client.py and scanner.py to monitor in the terminal where it was freezing.
This was to help me visualize. It appears that sometimes the disconnect callback gets called and locks up the program. its call back to disconnect before it has established a connection.

It is preferable if an issue contains a [Miminal Workable Example](https://stackoverflow.com/help/minimal-reproducible-example).
This will otherwise be one of the first questions you will get as a response.

It is also preferable if that example is not dependent on a specific peripheral device, but can be run and
reproduced with other BLE peripherals as well.

```
Paste the command(s) you ran and the output.
If there was a crash, please include the traceback here as well.
```

```python
    def _disconnected_callback_client(self):
        """
        Callback for device disconnection. Bleak callback sends one argument as client. This is wrapper function
        that gets called from the CentralManager and call actual disconnected_callback by sending client as argument
        """
        logger.debug("Received disconnection callback...")
        print("received disconnect callback")
        if self._disconnected_callback is not None:
            self._disconnected_callback(self)

    async def disconnect(self) -> bool:
        """Disconnect from the peripheral device"""
        print("hit disconnect function")
        if manager is None:
            return False
        await manager.disconnect()
        # Ensure that `get_services` retrieves services again, rather than using the cached object
        self._services = None
        return True
```

my output normally looks like this: 
```
await manager.connect
await self.get_services()
returning
connection successful with device: F53E1F1D-2DCE-44F1-ACEC-737FF821AE07
hit disconnect function
received disconnect callback
connection with device 47 was successful
```

When the program freezes, it looks like this:
```
await manager.connect
await self.get_services()
received disconnect callback
```

dlech:
This fixes #535. In the CoreBluetooth backend, when a device disconnects, none of the peripheral delegate callbacks will be called. So if any futures are pending, they will be waiting forever. This adds a disconnect hook that will raise an exception in all pending futures to avoid this deadlock.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -42,6 +42,8 @@ Fixed
 * Fixed write without response on BlueZ < 5.51.
 * Fixed error propagation for CoreBluetooth events
 * Fixed failed import on CI server when BlueZ is not installed.
+* Fixed deadlock in CoreBluetooth backend when device disconnects while
+  callbacks are pending. Fixes #535.
 
 
 `0.11.0`_ (2021-03-17)
--- a/bleak/backends/corebluetooth/CentralManagerDelegate.py
+++ b/bleak/backends/corebluetooth/CentralManagerDelegate.py
@@ -7,11 +7,12 @@ Created on June, 25 2019 by kevincar <kevincarrolldavis@gmail.com>
 """
 
 import asyncio
+import itertools
 import logging
 import platform
 import threading
 from enum import Enum
-from typing import List
+from typing import List, Optional
 
 import objc
 from CoreBluetooth import (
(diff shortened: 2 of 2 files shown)


dlech:
I tested this by adding

```python
        manager.central_manager.cancelPeripheralConnection_(
            manager.connected_peripheral
        )
```

before `await self.get_services()` in `BleakClientCoreBluetooth.connect()`.

dlech:
Superseded by ##564

PR #562, opened 2021-06-04
winrt: fix AttributeError in _stopped_handler()

### LINKED ISSUE #556: AttributeError in WinRT scanner _stopped_handler() callback
* bleak version: v0.12.0b1 (sha: ce63ed4)
* Python version: 3.9.5
* Operating System: Windows 10 20H2 (OS Build 19042.985)

### Description

I don't have a minimal example that reproduces the bug, but it appears that when scanning with the WinRT backend using `bleak.discover()`, I very rarely get exceptions in the scanner watcher stop callback:

```
Exception in callback BleakScannerWinRT._stopped_handler(<_winrt_Windo...0012FDA26F590>, <_winrt_Windo...0012FDA26F5B0>)

handle: <Handle BleakScannerWinRT._stopped_handler(<_winrt_Windo...0012FDA26F590>, <_winrt_Windo...0012FDA26F5B0>)>
  File "asyncio\events.py", line 80, in _run
  File "bleak\backends\winrt\scanner.py", line 118, in _stopped_handler
AttributeError: 'NoneType' object has no attribute 'status'
```

The `_stopped_handler()` callback that triggers the exception:
```python3
def _stopped_handler(self, sender, e):
    logger.debug(
        "{0} devices found. Watcher status: {1}.".format(
            len(self._devices), self.watcher.status
        )
    )
```

The `stop()` method that I think has a race condition:
```python3
async def stop(self):
    self.watcher.stop()

    try:
        self.watcher.remove_received(self._received_token)
        self.watcher.remove_stopped(self._stopped_token)
    except Exception as e:
        logger.debug("Could not remove event handlers: {0}...".format(e))

    self._stopped_token = None
    self._received_token = None

    self.watcher = None
```

I'm speculating that there might be a race condition where:
1. The scanner calls `stop()` to stop the watcher as part of its cleanup after the call to `bleak.discover()` returns.
2. The watcher stopped callback fires after we call `self.watcher.stop()` but before we remove the stopped callback, since it seems to run in a different thread.
3. The callback schedules our `_stopped_handler()` callback to run in our main thread/event loop via `call_soon_threadsafe()`.
4. Our main thread resumes and removes the callbacks successfully and sets `self.watcher = None`
5. `stop()` returns and eventually our `_stopped_handler()` callback is scheduled to run
6. `_stopped_handler()` attempts to access `self.watcher`, which is now None, triggering the exception.

From a quick attempt at inspecting the sender, it looks like one workaround would be for the `_stopped_handler()` callback to assume `sender` is always the watcher object and extract the status from the sender directly, rather than reading it from `self.watcher`. But I'm not intimately familiar with the WinRT callbacks dispatch, so I'm not sure if that always works.

devanlai:
Fixes #556 by first checking that `self.watcher is not None` before attempting to access the watcher status.

This edge case can come up if the watcher's stop event handler has a chance to fire when we manually stop the watcher as part of the scanner's `stop()` method.

### DIFF
--- a/bleak/backends/winrt/scanner.py
+++ b/bleak/backends/winrt/scanner.py
@@ -113,11 +113,13 @@ class BleakScannerWinRT(BaseBleakScanner):
         self._callback(device, advertisement_data)
 
     def _stopped_handler(self, sender, e):
-        logger.debug(
-            "{0} devices found. Watcher status: {1}.".format(
-                len(self._devices), self.watcher.status
+        # self.watcher can be None if we stopped the watcher in stop()
+        if self.watcher is not None:
+            logger.debug(
+                "{0} devices found. Watcher status: {1}.".format(
+                    len(self._devices), self.watcher.status
+                )
             )
-        )
 
     async def start(self):
         self.watcher = BluetoothLEAdvertisementWatcher()


devanlai:
Closing this in favor of #571.

PR #564, opened 2021-06-05
Mac cleanup and fixes

### LINKED ISSUE #535: Program disconnects before connection established. Locks up Bleak

### LINKED ISSUE #538: Crash on mac after timeout connecting to device
* bleak version: 0.11
* Python version: 3.8.5
* Operating System: MacOS 11.3.1

### Description

Sometimes (not always) when I get a timeout connecting to a device I get an `asyncio.exceptions.TimeoutError` exception. Subsequently the program crashes. 

Even if I catch the exception the Python program will still crash on exit.

Here are the relevant bits of the crash report:

```
Crashed Thread:        5  Dispatch queue: bleak.corebluetooth

Exception Type:        EXC_BAD_INSTRUCTION (SIGILL)
Exception Codes:       0x0000000000000001, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Termination Signal:    Illegal instruction: 4
Termination Reason:    Namespace SIGNAL, Code 0x4
Terminating Process:   exc handler [20790]

Application Specific Information:
BUG IN CLIENT OF LIBPTHREAD: pthread_exit() called from a thread not created by pthread_create()

Thread 0:: Dispatch queue: com.apple.main-thread
0   org.python.python             	0x00000001031f7a64 PyTuple_ClearFreeList + 84
1   org.python.python             	0x00000001032dc4ce collect + 3454
2   org.python.python             	0x00000001032db737 _PyGC_CollectNoFail + 55
3   org.python.python             	0x00000001032a2db1 PyImport_Cleanup + 1521
4   org.python.python             	0x00000001032b8f98 Py_FinalizeEx + 152
5   org.python.python             	0x00000001032da8db Py_RunMain + 1659
6   org.python.python             	0x00000001032dae3f pymain_main + 223
7   org.python.python             	0x00000001032db03b Py_BytesMain + 43
8   libdyld.dylib                 	0x00007fff20619f3d start + 1

Thread 5 Crashed:: Dispatch queue: bleak.corebluetooth
0   libsystem_pthread.dylib       	0x00007fff205fc7cd pthread_exit + 56
1   org.python.python             	0x00000001032ca905 PyThread_exit_thread + 21
2   org.python.python             	0x000000010326dd90 PyEval_RestoreThread + 128
3   org.python.python             	0x00000001032bc68a PyGILState_Ensure + 122
4   _objc.cpython-38-darwin.so    	0x0000000103a5ff98 method_stub + 40
5   _objc.cpython-38-darwin.so    	0x0000000103a83294 ffi_closure_unix64_inner + 708
6   _objc.cpython-38-darwin.so    	0x0000000103a827b6 ffi_closure_unix64 + 70
7   com.apple.CoreBluetooth       	0x00007fff30b31822 -[CBCentralManager handlePeripheralDisconnectionCompleted:] + 223
8   com.apple.CoreBluetooth       	0x00007fff30b31822 -[CBCentralManager handlePeripheralDisconnectionCompleted:] + 223
9   com.apple.CoreBluetooth       	0x00007fff30b31822 -[CBCentralManager handlePeripheralDisconnectionCompleted:] + 223
```
that last one is repeated for the 512 stack frames dumped in the crash report.

What _may_ be relevant is that I'm connecting to devices that are powered for only 200ms after sending an advertisement and then power down for 2 seconds (unless contacted by a client within those 200ms). If I understand the BLE specs correctly this is supported behaviour.

dlech:
This is a bit much all at once, but one thing led to another... and another...

Commit messages have more details, but the basic gist is:
- Improved type hints (includes adding typings for CoreBluetooth and other Obj-C APIs)
- Fixed a bunch of bugs
  - Fixes #535 
  - Fixes #538
  - And more
- Removed some unused and overly complex code


dlech:
I'm going to go ahead and merge this one. If there are regressions, they are more likely to be found by people using the develop branch.

PR #565, opened 2021-06-05
Discovery filter

jackjansen:
This change adds a new method `BleakScanner.find_device_by_filter(filterfunc)` that finds the first device that matches the filter function.

There's a new example `scanner_byname.py` that is similar to `scanner.py` but you specify the name of the BLE device you are looking for in stead of the address.

(I haven't merged the commits, in case you want to partially accept this).

### DIFF
--- a/bleak/backends/scanner.py
+++ b/bleak/backends/scanner.py
@@ -61,6 +61,11 @@ AdvertisementDataCallback = Callable[
     Optional[Awaitable[None]],
 ]
 
+AdvertisementDataFilter = Callable[
+    [BLEDevice, AdvertisementData],
+    bool,
+]
+
 
 class BaseBleakScanner(abc.ABC):
     """Interface for Bleak Bluetooth LE Scanners"""
@@ -195,10 +200,29 @@ class BaseBleakScanner(abc.ABC):
 
         """
         device_identifier = device_identifier.lower()
+        return await cls.find_device_by_filter(lambda d, ad : d.address.lower() == device_identifier)
+
+    @classmethod
+    async def find_device_by_filter(
+        cls, filterfunc, timeout: float = 10.0, **kwargs
+    ) -> Optional[BLEDevice]:
+        """A convenience method for obtaining a ``BLEDevice`` object specified by a filter function.
+
+        Args:
+            filterfunc: A function that is called for every peripheral found. It should return True only for the wanted device.
+            timeout (float): Optional timeout to wait for detection of specified peripheral before giving up. Defaults to 10.0 seconds.
+
+        Keyword Args:
+            adapter (str): Bluetooth adapter to use for discovery.
+
+        Returns:
+            The ``BLEDevice`` sought or ``None`` if not detected.
+
+        """
         stop_scanning_event = asyncio.Event()
 
         def stop_if_detected(d: BLEDevice, ad: AdvertisementData):
-            if d.address.lower() == device_identifier:
+            if filterfunc(d, ad):
                 stop_scanning_event.set()
 
         async with cls(
@@ -211,5 +235,5 @@ class BaseBleakScanner(abc.ABC):
             return next(
                 d
                 for d in scanner.discovered_devices
-                if d.address.lower() == device_identifier
+                if filterfunc(d, None)
             )
--- a/bleak/backends/scanner.py
+++ b/bleak/backends/scanner.py
@@ -62,7 +62,7 @@ AdvertisementDataCallback = Callable[
 ]
 
 AdvertisementDataFilter = Callable[
-    [BLEDevice, AdvertisementData],
+    [BLEDevice, Optional[AdvertisementData]],
     bool,
 ]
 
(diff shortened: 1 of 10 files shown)


bleak/backends/scanner.py
```diff
@@ -61,6 +61,11 @@ def __repr__(self) -> str:
     Optional[Awaitable[None]],
 ]
 
+AdvertisementDataFilter = Callable[
+    [BLEDevice, Optional[AdvertisementData]],
```
dlech, line 65: I don't thing `AdvertisementData` should be `Optional`. Am I missing something?

examples/scanner.py
```diff
@@ -19,6 +20,8 @@
     if platform.system() != "Darwin"
     else "B9EA5233-37EF-4DD6-87A8-2A875E821C46"  # <--- Change to your device's address here if you are using macOS
 )
+if len(sys.argv) == 2:
+    address = sys.argv[1]
```
dlech, line 24: :+1: for adding this to all of the examples that take an address

examples/scanner_byname.py
```diff
@@ -0,0 +1,30 @@
...
+wanted_name = sys.argv[1].lower()
+
+
+async def run():
+    device = await BleakScanner.find_device_by_filter(lambda d, ad : d.name and d.name.lower() == wanted_name)
+    print(device)
+
+
+loop = asyncio.get_event_loop()
+loop.run_until_complete(run())
```
dlech, line 30: I've been using `asyncio.run()` in examples since `loop.run_until_complete()` doesn't always clean up properly if the program is interrupted (even though we technically still support Python 3.6 and `asyncio.run()` requires Python 3.7).

dlech (commented):
This is really useful, thanks! I have made a few comments and it looks like we need to run `black` on the changes.

dlech:
And please add a changelog entry as well.

bleak/backends/scanner.py
jackjansen: The filter function is called in two places: once from the discovery callback but then later when filtering the results for returning to the caller. The second time the advertisement data isn't available. I see various workable solutions, please advise:
1. The current situation where (the the filter function is called once with 2 args once with 1 arg)
2. Getting rid of the advertisement data arg to the filter function (but that means you can no longer filter on interesting things in the advertisement data, which may be useful in some cases)
3. Getting rid of the filtering of the scan results and simply returning the first result (possibly raising an exception if more than one result was somehow available, I'm not 100% sure this can't happen with all the async stuff going on)
4. Change the signature/semantics of the filter function, stating that it returns either `None` or the `BLEDevice` that should be returned from `find_device_by_filter`, but I don't know whether it is possible to do more in the callback routine than simply setting a flag to make the discovery stop.

examples/scanner_byname.py
jackjansen: Hmm, just copied this from scanner.py. I'm not sure about the exact implications of run_until_complete verses run (I'm still pretty much an async virgin), so I would prefer if you pick this up for all examples at once....

bleak/backends/scanner.py
dlech: My choice would be 3. What I usually do is create an `asyncio.Queue()` ([example](https://github.com/hbldh/bleak/blob/develop/examples/uart_service.py)). This allows the callback to queue more than one match in case it receives more before we can stop the scanner but we only use the first item from the queue since it was the first match.

examples/scanner.py
jackjansen: Done, insofar as I had the possibility to do at least minimal testing.

examples/scanner_byname.py
dlech: Sure, no problem.

jackjansen:
Except for the filter function signature I think it's all there now. I did accidentally mark the "args for examples" discussion as done and I can't see how to undo that.

bleak/backends/scanner.py
jackjansen: Ok, for this use case I'll just do the simple thing and return the first match (after the scan has been stopped), or `None` if there's no matches.

But your `asyncio.Queue()` example makes me wonder: couldn't the whole discovery process be converted to using these? That way the whole filtering stuff would become a lot easier, and various other things would be possible with discovery too, as the results would become available as they were received...

jackjansen: Implemented (3).

hbldh (approved):
This looks good; a welcome improvement.

bleak/backends/scanner.py
```diff
@@ -208,8 +236,6 @@ def stop_if_detected(d: BLEDevice, ad: AdvertisementData):
                 await asyncio.wait_for(stop_scanning_event.wait(), timeout=timeout)
```
dlech, line 236: We should be able to get rid of the `stop_scanning_event` altogether and do something like this instead.

```suggestion
                return await asyncio.wait_for(found_device_queue.get(), timeout=timeout)
```

dlech (commented):
Looking good, just one more suggestion.

bleak/backends/scanner.py
jackjansen: I tried that first, but there was an issue with it so I put the event back in... But now I don't remember the exact issue, but if the filter didn't trigger something went wrong.

jackjansen: BTW: I won't be able to test/change anything until next weekend or later, so if you test the removal of the event and you're happy please just do it.

dlech: ok, I went ahead and took care of it.

dlech:
Merged, thanks!

PR #571, opened 2021-06-07
Adding a stop event in WinRT scanner

hbldh:
An alternate way of solving #556 and imo preferable to solving it by sacrificing the logging.

The stop event is originating from the Windows `BluetoothLEAdvertisementWatcher` class, so however the the scanning is stopped, it should generate an callback regardless of how it is stopped. This is, of course, you use the `BleakScanner`'s `start` method to initiate it...


dlech:
This looks good to me. I assume we need the same fix in the dotnet backend?

devanlai:
This fix looks good to me - it eliminates the race condition that I was using the reproduce the error in #556 and I've been running a test for a few minutes now running a scan with the WinRT backend every second for a few minutes now to stress test it and it seems fine.

hbldh:
Good!

Ah, yes. The two backends. I will fix it in the morning.

dlech:
Don't forget the changelog entry as we..

PR #583, opened 2021-06-27
16bit UUID support

jackjansen:
Small fix, but convenient: allow 16bit uuids to be passed to `get_service()` and `get_characteristic()`, by extending them to 128 bit as per the BLE specification.


hbldh:
It looks good. There has been multiple reports of devices with multiple services with the same uuid, and it is also in the Bluetooth LE specifications that it is allowed. Thus, referring to services and characteristics by UUID is a bad choice. You should use the handles for your specific device instead.

Bleak was initially built with the intention to only use UUIDs, because I felt using handles was a hassle and that the UUIDs was enough. That idea turned out to be erroneous, but the UUID usage is still present in Bleak. I think this PR is still a valuable addition, but know that you would have more stable code using handles.

hbldh (approved):

PR #609, opened 2021-08-02
bleak-winrt

dlech:
I got tired of waiting for Microsoft to update the `winrt` package with my fixes and improvements. So, I've created a new `bleak-wrint` package that contains my changes and published it. (It also only includes the namespaces used by bleak, so the binary is about 6MB instead of 40-something MB.) The main change is a fix for a major memory leak that was leaking a Windows COM object on every `await` call.

I've also dropped the `dotnet` backend since PythonNET hasn't been keeping up with Python releases and maintaining two backends for Windows seems like too much work. 

Technically, this means we would be dropping support for Python 3.6 on Windows since I've only compiled `bleak-winrt` for Python 3.7 and up. But 3.6 reaches end of life at the end of this year anyway.


dlech:
It would be interesting to do the analysis from https://github.com/hbldh/bleak/issues/255 on this to see if it fully fixes the memory leak issues seen there.

hbldh:
This is impressive work! It is a shame that Microsoft seems to have abandoned the winrt solution.

I had no idea if the winrt solution had memory problems when we implemented the first Bleak integration, but I did it with the assumption that it ought to be more rigorous than the `pythonnet` solution since it was a project from a well funded company.

Have you got any estimate on how much work it would take to make your winrt fork build for 3.10? This is far beyond my experience, given that it seems to involves a lot of really low level knowledge...

dlech:
It should be just a matter of adding `3.10` to the github actions workflow (assuming the `setup-python` action supports it already).

hbldh:
Lovely. I will try this out this week; I finally have some time that I want to actively invest in Bleak once again.

I cannot tell you how delighted I am to see the `BleakBridge` disappear. It was one of the hackiest things I have ever let out into the world: it depended on some bug/feature of the first available version (5.0.0) of `Microsoft.NETCore.UniversalWindowsPlatform`. I couldn't upgrade the version of that dependecy, because if I did, the `BleakBridge` namespace couldn't be imported by `pythonnet`. I also never figured out if that actually was a problem or not, since I never got around to verify whether `pythonnet` actually got version 5.0.0 of the `Windows` namespace components or found something newer. The `pythonnet` package is, in my mind, a semi-magical solution that I like a lot, but not completely trust...

aliencaocao:
Hi all, is there a replacement for the status attribute of the new WinRT backend? I am following the official doc but that was only for dotnet backend, and now when I try to access the status attribute of the scanner, it give me `'BleakScannerWinRT' object has no attribute 'status'`.

dlech:
Please start a new [discussion](https://github.com/hbldh/bleak/discussions) for this and include your full source code.

aliencaocao:
Ok, I have started one.
https://github.com/hbldh/bleak/discussions/638

PR #620, opened 2021-08-10
Added keyword argument option to force using Notifications instead of Indications in WinRT backend

### LINKED ISSUE #526: How can I activate notify if both notify and indicate are in the characteristic at the same time?
* bleak version: 0.11.0
* Python version: 3.7.8
* Operating System: windows 10

### Description

Hello,
Thanks for sharing a great library.

i'm using bleak to control BLE(bluetooth low energy) devices in Python. 
and i was trying to control a BLE device that included notify and indicate properties in one characteristic, 
so there was a problem that only indicate is activated.

so when i checked the bleak source code, it was found that if indicate exists, activate it as indicate, and if not, check notify and activate it as notify, and start_notify works.

the following is the relevant source code and link.

https://github.com/hbldh/bleak/blob/develop/bleak/backends/winrt/client.py

        async def start_notify(
                self,
                char_specifier: Union[BleakGATTCharacteristic, int, str, uuid.UUID],
                callback: Callable[[int, bytearray], None],
                **kwargs
            ) -> None:
                ...
                ...
                characteristic_obj = characteristic.obj
                if (
                    characteristic_obj.characteristic_properties
                    & GattCharacteristicProperties.INDICATE
                ):
                    cccd = GattClientCharacteristicConfigurationDescriptorValue.INDICATE
                elif (
                    characteristic_obj.characteristic_properties
                    & GattCharacteristicProperties.NOTIFY
                ):
                    cccd = GattClientCharacteristicConfigurationDescriptorValue.NOTIFY
                else:
                    cccd = GattClientCharacteristicConfigurationDescriptorValue.NONE
        
                fcn = _notification_wrapper(bleak_callback, asyncio.get_event_loop())
                event_handler_token = characteristic_obj.add_value_changed(fcn)
                self._notification_callbacks[characteristic.handle] = event_handler_token
                status = await characteristic_obj.write_client_characteristic_configuration_descriptor_async(
                    cccd
                )
        

### What I Did

what is the best way to deal with it in this case? 
so i implemented the code below to use "notify", but it doesn't look good.

        async with BleakClient(address, timeout=5.0) as client:
            client.set_disconnected_callback(on_disconnect)                        
            services = await client.get_services()        
            for service in services:
                for characteristic in service.characteristics:
                    if characteristic.uuid == read_write_charcteristic_uuid:  
                        if 'notify' in characteristic.properties:
                            # if "indicate" is exist remove it
                            if 'indicate' in characteristic.properties:
                                characteristic.properties.remove('indicate')
                            # then "notify" becomes active.
                            await client.start_notify(notify_charcteristic, callback)

thank you

hbldh:
Added `force_notify` keyword argument for WinRT backend client's `start_notify` method.

Fixes #526


dlech:
I am still of the opinion that Windows should be changed to use notifications instead of indications by default when both are available. This matches the BlueZ backend and the name of the method (and in my limited experience, notifications seem to be used much more frequently than indications).

dlech:
Random idea: should we "namespace" platform-specific kwargs like this with the platform name? I.e. `win_force_notification` or `win_force_indication` here. This would make it obvious to readers of user code that this is a platform-specific feature without having to consult the Bleak documentation or source code.

hbldh:
I agree with the default. Will change it to use notifications instead of indications.

There are multiple issue where people use some keywords in OSes that do not care for them. It is apparently not enough with documentation, since users look through issues/SO and use anything that is reminiscent of what they want to achieve.

I like the idea, but the prefixing seems horribly ugly to me. I have no better idea though...

How do other cross platform APIs handle this?

dlech:
> I like the idea, but the prefixing seems horribly ugly to me. I have no better idea though...

What if there is a standard kwarg for each platform that takes a dictionary. Looks like this: 

```python
start_notify(..., win={ "force_indication": True })
```

> How do other cross platform APIs handle this?

This is how cross-platform settings (json rather than API) in VS Code work.

hbldh:
I have given it some thought and I agree with your assessment: we need a clearer separation of OS specific arguments. I do not personally like neither the `win={}` nor the prefixing, but my reasoning is that a clear and understandable API for users trumps everything else.

Of the two I prefer the `win={}`, `macos={}`, `linux={}` (or `bluez`? not sure. `posix`?) I will open a new issue for this, because it will take a lot careful combing through all backends to get a complete coverage for this. It feels like it is a release of its own, like 0.14.0...

Flipped the default for Windows backends here now. Do you think the Changelog entry is enough or do you think it should be even more verbose about the changing of the default? I do not think many will notice it, but for those that do it might be helpful...

dlech:
It wouldn't hurt to mention the change in behavior.

hbldh:
Yes, I know and agree. I was hoping on you saying, don't bother and I would magically be free of handling the change...

dlech:
Haha, OK, only 3 people will notice. :grin:

hbldh:
That is enough. My bar for getting away for something is 0. 

It is right to do and I know it. I would have done it anyway before merging. Documentation and transparency is the only way forward in a project like this. Everything else is wrong.

I REALLY need to focus on #266 and get that done soon. It feels like 80% of our issues this spring and summer came from a poor documentation and a lack of including real users' user cases in it. I need to rewrite it.

bleak/backends/winrt/client.py
```diff
@@ -756,14 +760,22 @@ def bleak_callback(s, d):
         characteristic_obj = characteristic.obj
         if (
             characteristic_obj.characteristic_properties
-            & GattCharacteristicProperties.INDICATE
+            & GattCharacteristicProperties.NOTIFY
         ):
-            cccd = GattClientCharacteristicConfigurationDescriptorValue.INDICATE
+            if kwargs.get("force_indicate", False) and (
```
dlech, line 765: Would it be simpler to just say:

    if device has notify property and not force_indicate:
        use notify flag
    else if device has indicate property
        use indicate flag
    else
        error

bleak/backends/winrt/client.py
```diff
@@ -756,14 +760,22 @@ def bleak_callback(s, d):
...
+                cccd = GattClientCharacteristicConfigurationDescriptorValue.NOTIFY
         elif (
             characteristic_obj.characteristic_properties
-            & GattCharacteristicProperties.NOTIFY
+            & GattCharacteristicProperties.INDICATE
         ):
-            cccd = GattClientCharacteristicConfigurationDescriptorValue.NOTIFY
+            cccd = GattClientCharacteristicConfigurationDescriptorValue.INDICATE
         else:
             cccd = GattClientCharacteristicConfigurationDescriptorValue.NONE
```
dlech, line 780: Not related to this PR, but `NONE` would disable notifications/indications. Should we raise an error here instead?

dlech:
I added a commit with my suggested changes.

PR #622, opened 2021-08-11
Only warn if handles are added again

elupus:
Bluez seem to be re-adding interface:s causing exceptions to be thrown during message processing.

```
File "/usr/local/lib/python3.9/site-packages/dbus_next/message_bus.py", line 668, in _process_message
result = handler(msg)
File "/usr/local/lib/python3.9/site-packages/bleak/backends/bluezdbus/client.py", line 1057, in _parse_msg
self.services.add_characteristic(
File "/usr/local/lib/python3.9/site-packages/bleak/backends/service.py", line 160, in add_characteristic
raise BleakError(
bleak.exc.BleakError: This characteristic is already present in this BleakGATTServiceCollection!
2021-08-11 02:08:36 ERROR (MainThread) [root] A message handler raised an exception: This characteristic is already present in this BleakGATTServiceCollection!.
```

Just warn instead of throwing exception.

   guidelines in that document.
    - OS X / macOS >= 10.11


elupus:
Might even want to just make this debug level?

elupus:
I could also be that we are getting InterfacesRemoved that is not currently handled. Still i don't think making this a warning is a bad idea. Crashing on exception on this case seem bad.

dlech:
Since hitting this condition is a result of a bug in the Bleak library itself, I think this should be at least `logger.error` rather than warning. If you have a reproducible test case, lets try handling `InterfacesRemoved`. This has been on my todo list for a long time, but I haven't been able to get to it yet.

dlech:
Also, won't this mean that the service will now have old data (D-Bus property values) since the new data is discarded?

elupus:
I suppose it might have old data yes. I don't know if it is a bug in bleak. Are you sure bluez is not allowed to resend InterfacesAdded without removing them?

I think in my case it's a matter of a somewhat bad signal causing the data to change somewhat over time. Ie the interfaces goes away then comes back again.  I don't have a test case per-say i have a kitchen fan with ble that triggers this. Then suddenly crashes it's internal ble server. Unsure if it's my fault for hammering it with requests yet.

I've had the error occur 5000 times the last 20 hours. So it's not rare here. I have a scanner running continuously.

dlech:
> Are you sure bluez is not allowed to resend InterfacesAdded without removing them?

yes

elupus:
Switch it to error instead. I started looking at implementing InterfacesRemoved. But I don't feel comfortable in doing that. If you can do a draft implementation i can test it and possible refine it.

elupus:
Hmm. No that is likely not the case. This is what the devices announces:
```2021-08-11 02:36:11 DEBUG (MainThread) [custom_components.fjaraskupan] Detection: COOKERHOOD_FJAR CC:F9:57:96:4B:7B: COOKERHOOD_FJAR - AdvertisementData(local_name='COOKERHOOD_FJAR', manufacturer_data={20296: b'ODFJAR\x00\x02\x00\x00\x00\x00\x00'}, service_uuids=[
'00001800-0000-1000-8000-00805f9b34fb',
'00001801-0000-1000-8000-00805f9b34fb',
'77a2bd49-1e5a-4961-bba1-21f34fa4bc7b'
])

2021-08-11 02:36:11 DEBUG (MainThread) [custom_components.fjaraskupan] Detection: COOKERHOOD_FJAR CC:F9:57:96:4B:7B: COOKERHOOD_FJAR - AdvertisementData(local_name='COOKERHOOD_FJAR', manufacturer_data={20296: b'ODFJAR\x00\x02\x00\x00\x00\x00\x00'}, service_uuids=[
'00001800-0000-1000-8000-00805f9b34fb',
'00001801-0000-1000-8000-00805f9b34fb',
'0000180a-0000-1000-8000-00805f9b34fb',
'77a2bd49-1e5a-4961-bba1-21f34fa4bc7b'])
```

It's doing this in quick succession. So it really is announcing first with one set of services, then another set, one of the services going away. So it's quick clearly the lack of InterfacesRemoved that is the main issue here.

dlech:
Do we still need this? In #625, it seems like we determined that this was triggered by a BlueZ bug. So leaving the exception might not be a bad thing after all.

dlech:
In my testing, if we got this exception, it meant that the new attributes (services/characteristics/descriptors) provided by BlueZ were wrong, so I would rather have the exception thrown with a message that says you need to do the remove device/manually delete cache file workaround instead. Otherwise, we will get many reports of issues of "my service is missing" or "the characteristic returns the wrong value" or "enabling notifications does not work" because of the incorrect attributes reported by BlueZ. These issues are much harder to troubleshoot.

elupus:
Right. In my case thou the device is constantly changing. At least that is how it looks. So cache will always be wrong from time to time.

dlech:
Can we get more logs to show what changes on your device and when it changes? If it is at runtime, does it use the GATT Database Hash characteristic or Service Changed characteristic to send notifications of the changes? If this is the case, then I think we can handle it in a better way instead of just suppressing this error.

elupus:
I will try to get trace instead of the logs i posted.

Den mån 4 okt. 2021 20:49David Lechner ***@***.***> skrev:

Issue #376, opened 2020-12-03
"bleak.exc.BleakError: This service is already present in this BleakGATTServiceCollection!" when reconnecting a device


sandhu0409:
* bleak version: 0.9.1
* Python version: 3.7
* BlueZ version (`bluetoothctl -v`) in case of Linux: 5.50

### Description
For very first time the central code for scanning BLE Devices  worked and connected with uart device. But after connection I was not able to disconnect it. 

### What I Did
I had scanned BLE Devices and make connection. It worked and shown

disconnected, scanning
connected

Then I had stopped peripheral service. After service was stopped I had ran scanner class again then i was showing below error.

```
disconnected, scanning
Traceback (most recent call last):
  File "echoClient.py", line 31, in <module>
    ble.connect(advertisement)
  File "/usr/local/lib/python3.7/dist-packages/adafruit_ble/__init__.py", line 305, in connect
    connection = self._adapter.connect(advertisement.address, timeout=timeout)
  File "/usr/local/lib/python3.7/dist-packages/_bleio/adapter_.py", line 373, in connect
    return self.await_bleak(self._connect_async(address, timeout=timeout))
  File "/usr/local/lib/python3.7/dist-packages/_bleio/adapter_.py", line 134, in await_bleak
    return future.result(timeout)
  File "/usr/lib/python3.7/concurrent/futures/_base.py", line 432, in result
    return self.__get_result()
  File "/usr/lib/python3.7/concurrent/futures/_base.py", line 384, in __get_result
    raise self._exception
  File "/usr/local/lib/python3.7/dist-packages/_bleio/adapter_.py", line 384, in _connect_async
    await client.connect(timeout=timeout)
  File "/usr/local/lib/python3.7/dist-packages/bleak/backends/bluezdbus/client.py", line 157, in connect
    await self.get_services()
  File "/usr/local/lib/python3.7/dist-packages/bleak/backends/bluezdbus/client.py", line 405, in get_services
    BleakGATTServiceBlueZDBus(service, object_path)
  File "/usr/local/lib/python3.7/dist-packages/bleak/backends/service.py", line 116, in add_service
    "This service is already present in this BleakGATTServiceCollection!"
bleak.exc.BleakError: This service is already present in this BleakGATTServiceCollection!
```

hbldh:
Do you have two services with the same uuid on the peripheral? In that case this is related to #362.

sandhu0409:
No I have ran only one service with 
[...]

PR #625, opened 2021-08-15
Add support for removal of interfaces

elupus:
Support removal of interfaces (service, characteristics, descriptors) for bluez backends.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -26,6 +26,8 @@ jobs:
                 run: pip install -r requirements.txt
             -   name: Install development dependencies
                 run: pip install -r requirements_dev.txt
+            -   name: Install module self
+                run: pip install -e .
             -   name: Test with pytest
                 run: |
                     pytest tests --junitxml=junit/test-results-${{ matrix.os }}-${{ matrix.python-version }}.xml --cov=com --cov-report=xml --cov-report=html
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,7 @@ Added
 ~~~~~
 * Allow 16-bit UUID string arguments to ``get_service()`` and ``get_characteristic()``.
 * Added ``register_uuids()`` to augment the uuid-to-description mapping.
+* Support interface removal on bluez backends
 
 Fixed
 ~~~~~
--- a/bleak/backends/bluezdbus/characteristic.py
+++ b/bleak/backends/bluezdbus/characteristic.py
@@ -94,6 +94,22 @@ class BleakGATTCharacteristicBlueZDBus(BleakGATTCharacteristic):
         """
         self.__descriptors.append(descriptor)
 
+    def del_descriptor(self, handle: int):
+        """Remove a :py:class:`~BleakGATTDescriptor` to the characteristic.
+
+        Should not be used by end user, but rather by `bleak` itself.
+        """
+        for index, desc in enumerate(self.__descriptors):
+            if desc.handle == handle:
+                del self.__descriptors[index]
+                break
+        else:
+            raise ValueError(
+                "Descriptor {} doesn't exists on characteristics {}".format(
+                    handle, self.__service_handle
+                )
+            )
+
     @property
     def path(self) -> str:
         """The DBus path. Mostly needed by `bleak`, not by end user"""
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -25,6 +25,7 @@ from bleak.backends.bluezdbus.service import BleakGATTServiceBlueZDBus
 from bleak.backends.bluezdbus.signals import MatchRules, add_match, remove_match
 from bleak.backends.bluezdbus.utils import (
     assert_reply,
+    extract_handles_from_path,
     extract_service_handle_from_path,
     unpack_variants,
 )
@@ -1068,6 +1069,24 @@ class BleakClientBlueZDBus(BaseBleakClient):
                 )
         elif message.member == "InterfacesRemoved":
             path, interfaces = message.body
+            (
+                _,
+                service_handle,
+                characteristic_handle,
+                descriptor_handle,
+            ) = extract_handles_from_path(path)
+
+            if defs.GATT_SERVICE_INTERFACE in interfaces:
+                assert service_handle is not None
+                self.services.del_service(service_handle)
+
+            if defs.GATT_CHARACTERISTIC_INTERFACE in interfaces:
+                assert characteristic_handle is not None
+                self.services.del_characteristic(characteristic_handle)
+
+            if defs.GATT_DESCRIPTOR_INTERFACE in interfaces:
+                assert descriptor_handle is not None
+                self.services.del_descriptor(descriptor_handle)
 
         elif message.member == "PropertiesChanged":
             interface, changed, _ = message.body
--- a/bleak/backends/bluezdbus/service.py
+++ b/bleak/backends/bluezdbus/service.py
@@ -36,6 +36,22 @@ class BleakGATTServiceBlueZDBus(BleakGATTService):
         """
         self.__characteristics.append(characteristic)
 
+    def del_characteristic(self, handle: int):
+        """Remove a :py:class:`~BleakGATTCharacteristicBlueZDBus` characteristics from service.
+
+        Should not be used by end user, but rather by `bleak` itself.
+        """
+        for index, desc in enumerate(self.__characteristics):
+            if desc.handle == handle:
+                del self.__characteristics[index]
+                break
+        else:
+            raise ValueError(
+                "Characteristics {} doesn't exists on service {}".format(
+                    handle, self.__handle
+                )
+            )
+
     @property
     def path(self):
         """The DBus path. Mostly needed by `bleak`, not by end user"""
--- a/bleak/backends/bluezdbus/utils.py
+++ b/bleak/backends/bluezdbus/utils.py
@@ -51,3 +51,33 @@ def extract_service_handle_from_path(path):
         return int(path[-4:], 16)
     except Exception as e:
         raise BleakError(f"Could not parse service handle from path: {path}") from e
+
+
+def extract_handles_from_path(path: str):
+    segments = reversed(path.split("/"))
+    segment = next(segments)
+    if segment.startswith("desc"):
+        desc = int(segment[4:], 16)
+        segment = next(segments)
+    else:
+        desc = None
+
+    if segment.startswith("char"):
+        char = int(segment[4:], 16)
+        segment = next(segments)
+    else:
+        char = None
+
+    if segment.startswith("service"):
+        service = int(segment[7:], 16)
+        segment = next(segments)
+    else:
+        service = None
+
+    if segment.startswith("dev_"):
+        device = segment[4:]
+        segment = next(segments)
+    else:
+        device = None
+
+    return device, service, char, desc
--- a/bleak/backends/characteristic.py
+++ b/bleak/backends/characteristic.py
@@ -91,3 +91,11 @@ class BleakGATTCharacteristic(abc.ABC):
         Should not be used by end user, but rather by `bleak` itself.
         """
         raise NotImplementedError()
+
+    @abc.abstractmethod
+    def del_descriptor(self, handle: int):
+        """Remove a :py:class:`~BleakGATTDescriptor` to the characteristic.
+
+        Should not be used by end user, but rather by `bleak` itself.
+        """
+        raise NotImplementedError()
--- a/bleak/backends/service.py
+++ b/bleak/backends/service.py
@@ -55,6 +55,14 @@ class BleakGATTService(abc.ABC):
         """
         raise NotImplementedError()
 
+    @abc.abstractmethod
+    def del_characteristic(self, characteristic: BleakGATTCharacteristic):
+        """Add a :py:class:`~BleakGATTCharacteristic` to the service.
+
+        Should not be used by end user, but rather by `bleak` itself.
+        """
+        raise NotImplementedError()
+
     def get_characteristic(
         self, uuid: Union[str, UUID]
     ) -> Union[BleakGATTCharacteristic, None]:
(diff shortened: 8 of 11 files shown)


elupus:
I've not tested this yet. And the test suite is quite miniscule :)

elupus:
To do some type of verification i had to add some test cases. I'm not 100% sure i got everything correct. I also had to create a new path parser. I've not implemented it's use everywhere where handles are parsed from paths. Though that would make sense to do as a separate pull.

dlech:
This approach was my first though as well. However, after looking at what would be needed to handle this cross-platform, I think a different approach may be needed.

Backing up a bit, we need to understand why we are getting `InterfacesRemoved` in the first place. I recently encountered a bug with WebBluetooth on Linux that I think is basically the same cause as https://github.com/hbldh/bleak/issues/376 in Bleak. In my case, we added a PnP ID characteristic to the Device Information service. This apparently triggers something in BlueZ that causes it to remove the service momentarily before adding it back again.

IIRC, the sequence of events is like this:

- Connect to device
- InterfacesAdded signals
- ServicesResolved: True (PropertiesChanged signal)
- ServicesResolved: False (PropertiesChanged signal)
- InterfacesRemoved signals
- InterfacesAdded signals
- ServicesResolved True (PropertiesChanged signal)

Now changing subjects for a moment to discuss the cross-platform aspect. An interesting observation is the change in the ServicesResolved property. The corresponding APIs on other platforms are the [BluetoothLEDevice.GattServicesChanged Event](https://docs.microsoft.com/en-us/uwp/api/windows.devices.bluetooth.bluetoothledevice.gattserviceschanged?view=winrt-20348) on Windows and the [CBPeripheralDelegate peripheral:didModifyServices: method](https://developer.apple.com/documentation/corebluetooth/cbperipheraldelegate/1518865-peripheral?language=objc) on Mac. After both of these events, services must be manually re-enumerated. Mac gives a list of services that were invalidated, but Windows does not give any extra information. So the ServicesResolved: False signal in BlueZ is essentially the same event as the others just described. BlueZ just does the re-enumeration automatically instead of manually.

So here is my thinking...

## The quick hack solution

Disable the handling of InterfacesAdded when the `connect()` method exits.

The `connect()` method in the BlueZ backend waits for ServicesResolved: True. We need to handle InterfacesAdded before this, but we can disable it after this. This basically gets us on par with the other platforms. If services really change, we don't handle this - but this is already case on other platforms as well (see solution below). If BlueZ does its quirky thing of removing and adding the same services back (with same characteristics and descriptors), then the existing `BleakGATTServiceCollection` is still valid.

One potential issue here is the window of time between ServicesResolved: False and ServicesResolved: True where D-Bus objects may be removed and not added back yet. If we try to write a characteristic (or enable notifications) during this time, we will get a D-Bus error that the object does not exist on the bus. We could potentially work around this by adding a check to every method that access these D-Bus objects to say "if ServicesResolved == False then await ServicesResolved == True" before trying to send the D-Bus message.

## The cross-platform solution

Technically, BLE services can change dynamically after connecting. This is why the events describe above exist. However, I don't think this is used often (and I don't recall any Bleak issues requesting support for this) so we could probably live without this unless there is actually demand for it.

To implement this, basically we would need to add a callback registration similar to the disconnect callback to handle this event. When this event occurs, users would be expected to call `get_services()` to re-enumerate the services and discard any existing service/characteristic/descriptor objects since these are no longer valid.

elupus:
Just fixing up the tests. I'm not entirely sure there is anything wrong with supporting the removal of interfaces like this pull does.

I think there could be a race problem with your above indicated semi solution. I think in my case i have race situation where an announce from the device occurs right when we are attempting to connect to it. I noticed i had setup timing in such a way so they lined up. 

So i think the client registered for signals, then got the announce from the device (which added interfaces), then connected or something like that. That could be why the services list is different.

This race does seem to remain in the code. So maybe the client should not be listening to ServicesAdded/Removed at all. Only rely on the data it get from GetManagedObjects() on connect?

dlech:
It could also be helpful to have `bluetoothctl` running in a terminal while the problem happens.

dlech:
> It could also be helpful to have `bluetoothctl` running in a terminal while the problem happens.

I was able to get a reproducible case for this issue. See https://github.com/hbldh/bleak/issues/628 for such a log. 

I'm also curious if the workaround suggested in https://github.com/hbldh/bleak/issues/628#issuecomment-904181977 fixes the issue without changing Bleak.



> I'm not entirely sure there is anything wrong with supporting the removal of interfaces like this pull does.

Assuming we are seeing the same issue, I mostly agree. However, since there is no similar mechanism to remove individual services/characteristics/descriptors on other platforms, I think there may be a similar alternative that doesn't involve adding `del_*` methods that will only apply to BlueZ. 

This new suggestion is also motivated by the fact that `GetManagedObjects` has a major performance impact on constrained systems (like routers) and will hopefully be changed in the future anyway.

### New suggestion

In `scanner.py` we basically make a cache of all D-Bus object properties received in a dictionary so that we don't have to call individual `Get` or `GetAll` to get the properties. We could do the same here (with the idea of eventually only calling `GetManagedObjects` once globally for Bleak instead of each time we connect or start scanning - so if the implementations are mostly the same now, there will be less work to do later). 

So instead of creating and modifying the `BleakGATTServiceCollection` object when we call `"GetManagedObjects"` and received the `"InterfacedAdded"`, `"InterfacesRemoved"` and `"PropertiesChanged"` signals, we could just maintain a simple dictionary with the cached object paths and properties. Then when we receive ServicesResolved: True, we populate the `BleakGATTServiceCollection` object all at once.

This would probably also reduce some duplicate code between https://github.com/hbldh/bleak/blob/a83283411729f7fa42a68f827efe3cf2285ae03a/bleak/backends/bluezdbus/client.py#L254-L286 and  https://github.com/hbldh/bleak/blob/a83283411729f7fa42a68f827efe3cf2285ae03a/bleak/backends/bluezdbus/client.py#L1045-L1070.

elupus:
```
[DEL] Characteristic (Handle 0x1dd1)
        /org/bluez/hci0/dev_CC_F9_57_96_4B_7B/service000e/char000f
        00002a29-0000-1000-8000-00805f9b34fb
        Manufacturer Name String
[DEL] Characteristic (Handle 0x1dd1)
        /org/bluez/hci0/dev_CC_F9_57_96_4B_7B/service000e/char0011
        00002a24-0000-1000-8000-00805f9b34fb
        Model Number String
[DEL] Characteristic (Handle 0x1dd1)
        /org/bluez/hci0/dev_CC_F9_57_96_4B_7B/service000e/char0013
        00002a26-0000-1000-8000-00805f9b34fb
        Firmware Revision String
[DEL] Characteristic (Handle 0x1dd1)
        /org/bluez/hci0/dev_CC_F9_57_96_4B_7B/service000e/char0015
        00002a28-0000-1000-8000-00805f9b34fb
        Software Revision String
[DEL] Secondary Service (Handle 0x0027)
        /org/bluez/hci0/dev_CC_F9_57_96_4B_7B/service000e
        0000180a-0000-1000-8000-00805f9b34fb
        Device Information
[CHG] Device CC:F9:57:96:4B:7B UUIDs: 00001800-0000-1000-8000-00805f9b34fb
[CHG] Device CC:F9:57:96:4B:7B UUIDs: 00001801-0000-1000-8000-00805f9b34fb
[CHG] Device CC:F9:57:96:4B:7B UUIDs: 77a2bd49-1e5a-4961-bba1-21f34fa4bc7b
[NEW] Secondary Service (Handle 0xbef9)
        /org/bluez/hci0/dev_CC_F9_57_96_4B_7B/service000e
        0000180a-0000-1000-8000-00805f9b34fb
        Device Information
[NEW] Characteristic (Handle 0x9149)
        /org/bluez/hci0/dev_CC_F9_57_96_4B_7B/service000e/char000f
        00002a29-0000-1000-8000-00805f9b34fb
        Manufacturer Name String
[NEW] Characteristic (Handle 0x9149)
        /org/bluez/hci0/dev_CC_F9_57_96_4B_7B/service000e/char0011
        00002a24-0000-1000-8000-00805f9b34fb
        Model Number String
[NEW] Characteristic (Handle 0x9149)
        /org/bluez/hci0/dev_CC_F9_57_96_4B_7B/service000e/char0013
        00002a26-0000-1000-8000-00805f9b34fb
        Firmware Revision String
[NEW] Characteristic (Handle 0x9149)
        /org/bluez/hci0/dev_CC_F9_57_96_4B_7B/service000e/char0015
        00002a28-0000-1000-8000-00805f9b34fb
        Software Revision String
[CHG] Device CC:F9:57:96:4B:7B UUIDs: 00001800-0000-1000-8000-00805f9b34fb
[CHG] Device CC:F9:57:96:4B:7B UUIDs: 00001801-0000-1000-8000-00805f9b34fb
[CHG] Device CC:F9:57:96:4B:7B UUIDs: 0000180a-0000-1000-8000-00805f9b34fb
[CHG] Device CC:F9:57:96:4B:7B UUIDs: 77a2bd49-1e5a-4961-bba1-21f34fa4bc7b
[CHG] Device CC:F9:57:96:4B:7B ServicesResolved: yes
```

Hard to get those logs. I seem to get spammed with rssi updates :). Let me see if removing the cache drops the error.

elupus:
The cache does not seem to include this address on this system.

elupus:
Scratch that. Found the file. Missed the controller subdirectory. Removed the cache and indeed. It might actually solve the issue. Rather ugly workaround thou :)

dlech:
> The cache does not seem to include this address on this system.

Do you have 

    [GATT]
    Cache = no

Set in `/etc/bluetooth/*.conf`?

Also, the cache file is in a protected directory, so you won't see the file unless you are root.

---

Is the updated layout correct (handle numbers and UUIDS)? On my device, it had new firmware with additional characteristics. The old characteristics were deleted (`[DEL]` messages), but only the same number of attributes were added back. The last handle was 0x0013 (`/org/bluez/hci0/dev_90_84_2B_00_A8_8D/service0010/char0011/desc0013`) but should have been 0x0019 (`/org/bluez/hci0/dev_90_84_2B_00_A8_8D/service0014/char0017/desc0019`).

elupus:
image

Stupidly i forgot to grab copies of the file before i deleted it. I managed to get this screenshot. I'm just trying to see if i can spot a difference.

elupus:
Issue re-occured. Let me see if i can grab the file now.

elupus:
It feels like this devices might be doing weird announces from time to time. I've had it crash, and suddenly be announcing with a different name and different service set. Felt like some sort of fallback values.

I deleted the file. Connected, to the device and directly grabbed the cache. Waited a while for the announces to come in, and voila! Several additional attributes are suddenly added to the set.

image

So it seems it might be giving different data in announce and when connected?

dlech:
> So it seems it might be giving different data in announce and when connected?

When you say "announce" do you mean advertising data or the Service Changed characteristic?

elupus:
I'm not sure what i mean to be honest. I think i mean advertising data. I'm not at all up to date on BLE protocols. I mean the services and characteristics found when there is only announces going on. But now that you question it, I think there is no known characteristics from just advertisements? They are only resolved once we connect right.

My device is on the limit of radio access. So i get these early disconnects when connecting that i opened the other issue on. Could bluez internally get a messed up cache when we get early disconnects?

dlech:
The advertising data may contain a list of service UUIDs, but generally not characteristics. If you aren't using this, you could try setting 

    [GATT]
    Cache = yes

in `/etc/bluetooth/main.conf` to disable caching for devices that don't pair (also called trusted or bonded). There is a different bug in BlueZ that breaks the service UUIDs in advertising data when this setting is set to `yes` or `no` (default is `always`), so this workaround isn't for everyone.

You can use Wireshark to capture Bluetooth traffic to see what exactly is going on. https://bleak.readthedocs.io/en/latest/troubleshooting.html#linux

elupus:
In my case i could live without advertisement of services. It is detected on name. But i can't run wireshark here. It's an embedded system hassos for home assistant. Only linux system currently here. I suppose I could see if i can spin up something more standard somwhere to test out. Maybe i can try to grab logs on osx.

dlech:
I think there are ways to grab the capture data from the command line (e.g. `btmon`) and open it on another computer too. Running Bleak on Mac and capturing that would not be as good since it doesn't have BlueZ and would likely handle certain things differently.

dlech:
Re-open if I am incorrect, but it seems like we agree that this change is not currently needed.

elupus:
Well it might be. But the bug remains here. I need advertising data, so can't really use it if it's another bug. We still some fix for the exceptions thrown.

elupus:
Oh sorry. It was just the advertisment handling of service uuids that had bugs. So maybe i could till get the manufacture data. But i still think we need some fix for the lib throwing exceptions.

dlech:
FYI, some of the BlueZ bugs will be fixed in 5.62 when it is released if upgrading BlueZ is a possibility.

dlech:
> bthci_evt.bd_addr == cc:f9:57:96:4b:7b

FWIW, I had better luck with this filter: `(!(bthci_evt.code == 0x3e)) && !(frame.protocols == "bluetooth:hci_mon")`

elupus:
Ps. I managed to temporary hack hassos (bind mounted a main.conf onto the readonly file system :)) to run with Cache=yes, and the error goes away. That doesn't feel like a solution thou.

dlech:
> That doesn't feel like a solution thou.

Actually, it should be a good workaround for the BlueZ bug, but I found when doing this, it triggered a different BlueZ bug that prevented services UUIDs from being returned when scanning the device for a second time. But, if you don't need service UUIDs from advertising data, then this should work.

dlech:
It could very well be that the services change before connecting. In the packet logs though, for some reason, it gives a different response to the same request at different times. I've highlighted it in the screenshot below.

image

I think this is what is triggering the removal and re-adding of D-Bus interfaces.

dlech:
At this point, I feel like I understand what is going on well enough to make a decision. I think `Cache=yes` would probably be the best "fix", but you mentioned this was difficult to implement (and it can trigger other BlueZ bugs), so I will consider merging #622 if that works for you.

PR #643, opened 2021-09-25
Handle lost services on osx

elupus:
When a device has gone out of range and then returned again, the service list is reset to None. So try to get it again.


bleak/backends/corebluetooth/PeripheralDelegate.py
```diff
@@ -87,14 +87,21 @@ def futures(self) -> Iterable[asyncio.Future]:
...
 
         future = self._event_loop.create_future()
         self._services_discovered_future = future
         self.peripheral.discoverServices_(None)
         await future
 
-        return self.peripheral.services()
+        services = self.peripheral.services()
+        if services is None:
+            raise RuntimeError("Unable to get services")
```
dlech, line 103: This should be `BleakError` rather than `RuntimeError`.

bleak/backends/corebluetooth/PeripheralDelegate.py
```diff
@@ -87,14 +87,21 @@ def futures(self) -> Iterable[asyncio.Future]:
     @objc.python_method
     async def discover_services(self, use_cached: bool = True) -> NSArray:
         if self._services_discovered_future.done() and use_cached:
```
dlech, line 89: I think we should be able to get rid of the future done check here.


```suggestion
        if use_cached:
```

bleak/backends/corebluetooth/PeripheralDelegate.py
```diff
@@ -87,14 +87,21 @@ def futures(self) -> Iterable[asyncio.Future]:
...
-            return self.peripheral.services()
+            services = self.peripheral.services()
+            if services is None:
+                logger.debug("Services lost retrying")
```
dlech, line 92: We should be able to do without this debug log (especially if the previous suggestion is taken). In the case that we have to call `discoverServices_()`, the delegate callback should trigger a log anyway, so we have another way to know that code path is taken.

elupus:
I've not tested yet. But your example should work perfectly fine. It's just a more extreme version of mine. Never use the cache. So i'm perfectly good with that.

dlech:
Sounds good.

PR #654, opened 2021-10-06
BleakClient: add optional services argument

dlech:
This adds an optional `services` argument to `BleakClient()` to make use of OS API capabilities of only enumerating services that will be used. These APIs are available on Windows and Mac, but not BlueZ, so on the BlueZ backend, it is faked. Theoretically, this can save battery power on peripheral devices, but this is probably only relevant to super-low-powered devices.


karlp:
Not just saves power, it can also be quite a bit faster!

PR #664, opened 2021-10-08
BlueZ: add lazy-get/check version method

### LINKED ISSUE #602: Bleak dbus client opens bluetooth process on each instanciation
* bleak version: 0.12.1
* Python version: 3.9
* Operating System: Linux

### Description

With this type of workflow: Open connection, give device command, close connection using a async context manager the dbus bluez backend spawns subprocesses on each iteration here: https://github.com/hbldh/bleak/blob/7e0fdae6c0f6a78713e5984c2840666e0c38c3f3/bleak/backends/bluezdbus/client.py#L88

Feel like this should be a cached check.

### What I Did

```python
while True:
  async with BleakClient(ble_address) as client:
    client.read..
  await asyncio.sleep(10)
```

I suppose i could do as this:

```python
client = BleakClient(ble_address)
while True:
  async with client:
    client.read..
  await asyncio.sleep(10)
```

But it's less clean, and I seem to recall i had problem with that type of code on Osx.

dlech:
This adds a new method to the BlueZ bakckend for checking the BlueZ version. This lazy-gets the version from bluetoothctl so that the subprocess only runs once.

Fixes #602.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -46,7 +46,7 @@ Fixed
 * Fixed protocol error code descriptions on WinRT backend. Fixes #532.
 * Fixed race condition hitting assertation in BlueZ ``disconect()`` method. Fixes #641.
 * Fixed enumerating services on a device with HID service on WinRT backend. Fixes #599.
-
+* Fixed subprocess running to check BlueZ version each time a client is created. Fixes #602.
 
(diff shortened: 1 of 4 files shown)


dlech:
This is an alternate to #619.

PR #679, opened 2021-10-26
Remove hardcoded timeout values

linnik:
This PR removes hardcoded timeout values in all backends in favor of default `BaseBleakClient._timeout` or kwargs `timeout=...` argument.

Otherwise it would be impossible to connect with slow devices even if timeout option is given:
```
  File "/usr/lib64/python3.9/asyncio/locks.py", line 226, in wait
    await fut
asyncio.exceptions.CancelledError

During handling of the above exception, another exception occurred:

  File "/usr/lib64/python3.9/asyncio/tasks.py", line 492, in wait_for
    fut.result()
asyncio.exceptions.CancelledError

The above exception was the direct cause of the following exception:

  File "/home/user/.../test.py", line 70, in <module>
    asyncio.run(main(address))
  File "/usr/lib64/python3.9/asyncio/runners.py", line 44, in run
  File "/usr/lib64/python3.9/asyncio/base_events.py", line 642, in run_until_complete
    return future.result()
  File "/home/user/.../test.py", line 30, in main
    await client.connect(timeout=60)
  File "/home/user/.../__pypackages__/3.9/lib/bleak/backends/bluezdbus/client.py", line 317, in connect
    await self.get_services()
  File "/home/user/.../__pypackages__/3.9/lib/bleak/backends/bluezdbus/client.py", line 592, in get_services
    await asyncio.wait_for(self._services_resolved_event.wait(), 5)
  File "/usr/lib64/python3.9/asyncio/tasks.py", line 494, in wait_for
    raise exceptions.TimeoutError() from exc
asyncio.exceptions.TimeoutError
```

### DIFF
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -90,7 +90,7 @@ class BleakClientBlueZDBus(BaseBleakClient):
         """Connect to the specified GATT server.
 
         Keyword Args:
-            timeout (float): Timeout for required ``BleakScanner.find_device_by_address`` call. Defaults to 10.0.
+            timeout (float): Defaults to 10.0.
 
         Returns:
             Boolean representing connection status.
@@ -377,8 +377,11 @@ class BleakClientBlueZDBus(BaseBleakClient):
             self.services = BleakGATTServiceCollection()
             self._services_resolved = False
 
-    async def disconnect(self) -> bool:
+    async def disconnect(self, **kwargs) -> bool:
         """Disconnect from the specified GATT server.
+
+        Keyword Args:
+            timeout (float): Defaults to 10.0.
 
         Returns:
             Boolean representing if device is disconnected.
@@ -396,10 +399,11 @@ class BleakClientBlueZDBus(BaseBleakClient):
             logger.debug(f"already disconnected ({self._device_path})")
             return True
 
+        timeout = kwargs.get("timeout", self._timeout)
         if self._disconnecting_event:
             # another call to disconnect() is already in progress
             logger.debug(f"already in progress ({self._device_path})")
-            await asyncio.wait_for(self._disconnecting_event.wait(), timeout=10)
+            await asyncio.wait_for(self._disconnecting_event.wait(), timeout=timeout)
         elif self.is_connected:
             self._disconnecting_event = asyncio.Event()
             try:
@@ -413,7 +417,9 @@ class BleakClientBlueZDBus(BaseBleakClient):
                     )
                 )
                 assert_reply(reply)
-                await asyncio.wait_for(self._disconnecting_event.wait(), timeout=10)
+                await asyncio.wait_for(
+                    self._disconnecting_event.wait(), timeout=timeout
+                )
             finally:
                 self._disconnecting_event = None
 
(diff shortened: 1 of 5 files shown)


bleak/backends/bluezdbus/client.py
```diff
@@ -90,7 +90,7 @@ async def connect(self, **kwargs) -> bool:
         """Connect to the specified GATT server.
 
         Keyword Args:
-            timeout (float): Timeout for required ``BleakScanner.find_device_by_address`` call. Defaults to 10.0.
+            timeout (float): Defaults to 10.0.
```
mdxs, line 93: Perhaps specify the unit as seconds; so "timeout (float): Defaults to 10.0 seconds."

bleak/backends/bluezdbus/client.py
```diff
@@ -377,9 +377,12 @@ def _cleanup_all(self) -> None:
             self.services = BleakGATTServiceCollection()
             self._services_resolved = False
 
-    async def disconnect(self) -> bool:
+    async def disconnect(self, **kwargs) -> bool:
         """Disconnect from the specified GATT server.
 
+        Keyword Args:
+            timeout (float): Defaults to 10.0.
```
mdxs, line 384: Perhaps specify the unit as seconds; so "timeout (float): Defaults to 10.0 seconds."

bleak/backends/bluezdbus/client.py
```diff
@@ -575,6 +581,9 @@ def mtu_size(self) -> int:
     async def get_services(self, **kwargs) -> BleakGATTServiceCollection:
         """Get all services registered for this GATT server.
 
+        Keyword Args:
+            timeout (float): Defaults to 10.0.
```
mdxs, line 585: Perhaps specify the unit as seconds; so "timeout (float): Defaults to 10.0 seconds."

bleak/backends/winrt/client.py
```diff
@@ -258,14 +258,18 @@ def _ConnectionStatusChanged_Handler(sender, args):
 
         return True
 
-    async def disconnect(self) -> bool:
+    async def disconnect(self, **kwargs) -> bool:
         """Disconnect from the specified GATT server.
 
+        Keyword Args:
+            timeout (float): Defaults to 10.0.
```
mdxs, line 265: Perhaps specify the unit as seconds; so "timeout (float): Defaults to 10.0 seconds."

mdxs:
@hbldh As I read this PR code changes, it just tries to remove hard-coded timeouts with something that can be specified. If not provided it defaults to 10.0 seconds; perhaps you would rather see that default timeout be set to 5.0 seconds?

Then as I read the PR it would allow someone to specify the timeout if so desired, to a longer or a shorter timeframe.

dlech:
> it just tries to remove hard-coded timeouts with something that can be specified

I think the question is: is there an actual technical problem that is fixed with a different timeout or are we just doing this for the sake of making everything configurable?

mdxs:
Though I'm not the OP, for me this looks like something that could help me in testing. The use case being that I know something about the Device Under Test (DUT) such as a unique name or the MAC address from reading a NFC tag or scanning a QR code; after which I would like to connect to the DUT via BLE within a given amount of time, then discover its services within another given amount of time. The purpose being to figure out if the DUT is able to meet certain minimal criteria (which of course are parameters in a test script to be executed against the DUT when the test starts).

Thus for me the "make everything configurable" does seem interesting enough indeed; the alternative being of course that I would schedule time-outs differently (another thread running a clock and scheduling the time-out for each part of the test); and then if the DUT doesn't meet the test, I will "fail" it and terminate the DUT in some way... which may cause some side effects, so more graceful controlled time-outs sound interesting.

But my use case might be very specific and not fit the purpose of bleak, so it is just an input for your consideration of course.

Issue #230, opened 2020-06-27
Scan peripherals and filter by service UUID


joelviel:
Hi,

Is it possible to scan peripherals and only consider those with a specific service UUID?
BLE addresses change often if using Android peripheral, what is your solution to detect compliant peripherals with your central in that case?

I am currenty using NodeJS noble ( noble.startScanning(uuid) ) but I am looking for alternative

Thanks

bsiever:
By OS:

- The Linux backend appears to allow this via an optional `filters` dictionary (I haven't tested it). 
- PR #209 should add identical support for CoreBluetooth too.  (I have tested some of it)
- I don't know much about the Windows backend, but it doesn't look like it does filtering yet. 

The BlueZ documentation includes a section on `SetDiscoveryFilters` that gives the keywords allowed and their meaning. See: https://www.mideiros.net/doc/system/packages/bluez/dbus-apis/adapter-api.txt (Search for SetDiscoveryFilter)

Here's an example of scanning in the CoreBluetooth version:
```
    scanner = bleak.BleakScanner(filters={"UUIDs":["1d93af38-9239-11ea-bb37-0242ac130002"], "DuplicateData":False})
    scanner.register_detection_callback(detection_callback)
    await scanner.start()
    await asyncio.sleep(3.0)
    await scanner.stop()
    devices = await scanner.get_discovered_devices()
```

dlech:
Is there a compelling reason (like performance) that the filtering should be done at the OS level?

I've been thinking it would be nice to have an async iterator in the scanner. This could be used to implement filters in Python and it would not be restricted to the lowest common denominator of OS capabilities. (Off topic: this feature would also nicely replace `register_detection_callback()`)

For example, get the first device with matching service UUID (ignore that it doesn't match the current API - this is more like pseudo-code):

```python
MY_UUID = 'xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx'

async with BleakScanner() as scanner:
    # this has an iteration every time a new advertisement is seen
    async for advertisement in scanner.scan(timeout=5.0):
        if advertisement.service_uuids and MY_UUID in advertisement.service_uuids:
            my_device_id = advertisement.device_id
            # Breaking out of the for loop stops scanning. If there were any advertisements
            # already received after this one, they are ignored.
            break
    else:
        # The for loop completed without `break` so we must have reached timeout.
        print('device not found!')
        exit(1)

    my_device = await scanner.connect(my_device_id)
    ...
[...]

PR #692, opened 2021-11-27
corebluetooth: work around macOS 12 scanner bug

### LINKED ISSUE #635: No results when scanning on MacOS 12
* bleak version: 0.12.1
* Python version: 3.9.6
* Operating System: MacOS 12 Beta (Monterey)

### Description

I installed bleak via pip, and used the scanning example from the docs. When the code is run, no devices are found. I can run a different BLE scanner and see many devices. Running as root doesn't change anything.

Happy to help investigate, but I'm not sure where to go from here.

dlech:
macOS has a bug (feature?) where advertising data is no longer received unless one or more service UUIDs are given in `scanForPeripheralsWithServices:options:`.

This implements a new kwarg on `BleakScanner` to allow users to provide such a list of UUIDs. This commit only implements it in the CoreBluetooth backend with other backends to follow.

If macOS 12 is detected and this kwarg is not given, an error will be logged explaining how to fix the problem.

Fixes #635.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,7 +14,7 @@ Added
 -----
 
 * Added ``service_uuids`` kwarg to  ``BleakScanner``. This can be used to work
-  around issue of scanning not working on macOS 12. Issue #635.
+  around issue of scanning not working on macOS 12. Fixes #230. Works around #635.
 
 Changed
 -------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,12 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 `Unreleased`_
 =============
 
+Added
+-----
+
+* Added ``service_uuids`` kwarg to  ``BleakScanner``. This can be used to work
+  around issue of scanning not working on macOS 12. Issue #635.
+
 Changed
 -------
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -21,6 +21,8 @@ Changed
 
 * Changed WinRT backend to use GATT session status instead of actual device
   connection status.
+* Changed handling of scan response data on WinRT backend. Advertising data
+  and scan response data is now combined in callbacks like other platforms.
 
 Fixed
 -----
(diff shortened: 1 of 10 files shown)


hbldh:
This is great, but a sad state of affairs that it is needed. I was wrong in my comment in the issue then: there was a workaround that was reasonable...

It does require prior knowledge of the device though, or could you pass e.g. Device Information service uuid if nothing is sent in in Bleak and get all devices anyway? The DI is always present, or am I wrong again? 🤔

dlech:
> The DI is always present, or am I wrong again?

The services have to be specifically included in the advertising data, so if a device advertises no services, we won't be able to connect. The [Accessory Design Guidelines for Apple Devices](https://developer.apple.com/accessories/Accessory-Design-Guidelines.pdf) specifically says:

> The accessory shall implement the Device Information Service. The service UUID for this service should not be advertised in the Advertising Data

So I'm afraid you are wrong about counting on the DI service. 

However, WebBluetooth requires advertised services to be specified like this already (on all OSes), so if a device works with WebBluetooth, then it can work with Bleak.

dlech:
All platforms have been implemented now. Incidentally, this also fixes #230.

bri3d:
Unless I'm missing something, this doesn't fix `find_device_by_address` and by proxy the use of `connect` with a string device identifier on Mac OS 12, because there's no way to send the new kwarg through that call chain.

dlech:
The kwargs from `find_device_by_address` are passed through to the `BleakScanner` constructor, so that should work with the new `service_uuids` kwarg.

Using `async with BleakClient(address) as client:` won't work currently though as you have pointed out. I suppose we could add a `scanner_args` kwarg there to pass to the scanner. Or maybe we should consider deprecating/discouraging that use - it seems like a good number of issues are resolved by advising people to use a separate `BleakScanner` object for scanning, so if that was only/recommended way to do it, we could save people some trouble.

bri3d:
From my standpoint as a user, I don't see any reason not to deprecate that use - switching my project away from it was trivial and I agree there are wins there. 👍

challiwill:
What is the path to get this merged to `master`? I have been using it and it has been working for me.

ph4r05:
Is there pls any estimate on when this PR gets to a new release on pipy? Thanks

elupus:
How does this work with manufacturer data? They seem to come as separate announcements from the list of services? Will the scanner still get these manufacturer data if the device at least once announce a selected service?

dlech:
MacOS combines the information from the advertising data packet and scan response packet so I don't expect any problems in that regard.

elupus:
I can confirm that it seem to work. Seams to take a bit longer for first manufacturer data to show up, but no issue otherwise.

Issue #623, opened 2021-08-12
Separating OS specific keyword arguments


hbldh:
* bleak version: 0.12.0
* Python version: All
* Operating System: All

Bleak has several OS specific keyword, all of which are integrated into the regular `**kwargs` input to methods. The primary means of making users aware of what they can do to affect their current environment is through documentation. The documentation does provide this, but it is evidently not enough, given the amount of issues that e.g. asks why `device="hci1` does not work in Windows to select another Adapter.

@dlech proposed two different ways of improving Bleak in this regards:

1. Using a `win_`, `macos_`, and `linux_` prefix to keyword arguments that are os specific.
2. Using a `win={"x"=1. "y"=2, ...}` style of OS specific grouping of parameters.

Both will render nicely with `black` and both relays the desired effect in my mind. @hbldh prefers the second way, but given that Bleak _currently_ has few OS specific parameters it will be almost equal. Given that I, creator of Bleak, has much less use of it that most everyone else that will read this will have, I am more that ready to yield to public demand. 

I vote 2, and everyone else that votes will have his or her's vote counted equally. At the first of September, I will start to implement the winning choice. IF someone beats me to the implementation of one of the choices, I will take that one.

dlech:
Option 2 seems like it could be adapted to use [typing.TypedDict](https://docs.python.org/3/library/typing.html#typing.TypedDict) in the future (requires Python 3.8) as well for better type hints.

dlech:
This has been implemented.

PR #695, opened 2021-12-02
Use Cached Services option for WinRT backend

### LINKED ISSUE #686: `BleakClientWinRT.get_services` takes longer time in Windows 11 than in Windows 10
* bleak version: 0.14.0a1 (`feature/wtd` branch)
* Python version: 3.7.8
* Operating System: Windows 11 ('10', '10.0.22000', 'SP0', 'Multiprocessor Free')

### Description

`get_services` call take a lot longer time in Windows 11 than in Windows 10.

### What I Did

A near 10 second connection delay in Windows 11 comparing to 1-5 seconds in Windows 10 was surprising to me. This issue is more of a mental note than an error report.  Based on memory, it was faster than 10 seconds earlier.

If anyone else using Bleak and noticing the same, please verify that it takes longer time in Windows 11 if you have experienced the same, or refute it if you have experienced the opposite.

```
import os

os.environ["BLEAK_LOGGING"] = "1"

import asyncio

address = "24:71:89:cc:09:05"
MODEL_NBR_UUID = "00002a24-0000-1000-8000-00805f9b34fb"


async def run(address):
    device = await BleakScanner.find_device_by_address(address)
        model_number = await client.read_gatt_char(MODEL_NBR_UUID)
        print("Model Number: {0}".format("".join(map(chr, model_number))))

asyncio.run(run(address))
```

Output:

```powershell
2021-11-06 21:54:40,931 bleak.backends.winrt.scanner DEBUG: Received 24:71:89:CC:09:05: Unknown.
2021-11-06 21:54:40,933 bleak.backends.winrt.scanner DEBUG: 2 devices found. Watcher status: 3.
2021-11-06 21:54:40,933 bleak.backends.winrt.client DEBUG: Connecting to BLE device @ 24:71:89:CC:09:05
2021-11-06 21:54:41,551 bleak.backends.winrt.client DEBUG: _ConnectionStatusChanged_Handler: 1
2021-11-06 21:54:41,551 bleak.backends.winrt.client DEBUG: Get Services...
2021-11-06 21:54:53,966 bleak.backends.winrt.client INFO: Services resolved for BleakClientWinRT (24:71:89:CC:09:05)
2021-11-06 21:54:54,085 bleak.backends.winrt.client DEBUG: Read Characteristic 000C : bytearray(b'CC2650 SensorTag')
Model Number: CC2650 SensorTag
2021-11-06 21:54:54,086 bleak.backends.winrt.client DEBUG: Disconnecting from BLE device...
2021-11-06 21:54:57,205 bleak.backends.winrt.client DEBUG: _ConnectionStatusChanged_Handler: 0
```

hbldh:
This allows Bleak to use the Windows Bluetooth cache for services, characteristics and descriptors. Since Windows 11 took 10 seconds to fetch services on each connect, this will provide a faster connect on devices that do not change.

Fixes #686

Also modified Characteristic class in WinRT to use description property from super class in case of empty string description from Windows object.

The SensorTag example is a bit expanded to also light up a LED.


bleak/backends/winrt/client.py
```diff
@@ -113,10 +113,14 @@ class BleakClientWinRT(BaseBleakClient):
     a package that enables Python developers to access Windows Runtime APIs directly from Python.
 
     Args:
-        address_or_ble_device (`BLEDevice` or str): The Bluetooth address of the BLE peripheral to connect to or the `BLEDevice` object representing it.
+        address_or_ble_device (``BLEDevice`` or str): The Bluetooth address of the BLE peripheral
```
dlech, line 116:
```suggestion
        address_or_ble_device (str or BLEDevice): The Bluetooth address of the BLE peripheral
```

Would be nice to put type in same order as descriptions. Also, no backticks should be necessary here.

bleak/backends/winrt/client.py
```diff
@@ -113,10 +113,14 @@ class BleakClientWinRT(BaseBleakClient):
...
+            to connect to or the ``BLEDevice`` object representing it.
 
     Keyword Args:
         timeout (float): Timeout for required ``BleakScanner.find_device_by_address`` call. Defaults to 10.0.
+        use_cached_services (bool): ``True`` allows Windows to fetch the services, characteristics and descriptors
```
dlech, line 121: I think this would be a good time to start implementing #623 and do something to indicate that this kwarg is windows-only.

bleak/backends/winrt/client.py
```diff
@@ -113,10 +113,14 @@ class BleakClientWinRT(BaseBleakClient):
...
+            from the Windows cache instead of reading them from the device. Can be very much faster for known devices.
```
dlech, line 122: We should probably mention that this is not a good idea for DIY devices where the attribute layout may change frequently.

bleak/backends/winrt/client.py
```diff
@@ -152,12 +157,17 @@ async def connect(self, **kwargs) -> bool:
 
         Keyword Args:
             timeout (float): Timeout for required ``BleakScanner.find_device_by_address`` call. Defaults to 10.0.
+            use_cached_services (bool): ``True`` allows Windows to fetch the services, characteristics and descriptors
```
dlech, line 160: Adding the kwarg to the connect method too seems a bit much. What is the use case?

bleak/backends/winrt/client.py
```diff
@@ -442,9 +458,14 @@ async def get_services(self, **kwargs) -> BleakGATTServiceCollection:
...
             logger.debug("Get Services...")
+            use_cached_services = kwargs.get(
+                "use_cached_services", self._use_cached_services
+            )
             services: Sequence[GattDeviceService] = _ensure_success(
                 await self._requester.get_gatt_services_async(
-                    BluetoothCacheMode.UNCACHED
+                    BluetoothCacheMode.CACHED
+                    if use_cached_services
+                    else BluetoothCacheMode.UNCACHED
```
dlech, line 468: there would be less duplicate code if we just did this conversion from `bool` to `BluetoothCacheMode` once above instead of in each function call.

dlech (commented):
Looks good. I would be curious to see the Bluetooth packet logs to see why it is taking so long to fetch the attributes.

bleak/backends/winrt/client.py
hbldh: Hadn't noticed that. You are very right.

Will Google-style do string args render differently than other text? Will check it out and change.

dlech: It should pick up the type and automatically create a link.

bleak/backends/winrt/client.py
hbldh: Did I write that issue?! Ok. Let's start changing the standard now then. I will change for this method now and do the rest in the documentation rewriting.

bleak/backends/winrt/client.py
hbldh: I was planning on having default False and write about the `use_cached_services` as an optional boost in the documentation, for those who are confident about their device.

Do you think it should be the other way around? Use cached unless you know that your device is changing between connects? It would be faster for general users, but really annoy the Ones not only playing around. I don't know what is preferable.

How many of Bleak users also write their own peripheral? 25% at best is my guess.

bleak/backends/winrt/client.py
hbldh: It much, I agree. But it is what initiates the `get_services` in the non-context manager way of connecting, and it felt appropriate to add it there as well if it was also in `get_services`. Maybe I should remove it from both and just go with the constructor...

bleak/backends/winrt/client.py
hbldh: Yes. Good call. This was lazy coding.

hbldh:
I was also thinking about that. Will do, but for now I just did the implementation.

bleak/backends/winrt/client.py
dlech: We actually used to have the cached enabled by default, but I removed it because we kept having issues reported that "bleak isn't working". I don't know that it has been long enough to know if that actually solved any problems or not. But I think you have the right idea here that it should not use cache by default.

dlech: > How many of Bleak users also write their own peripheral? 25% at best is my guess.

I don't have a good guess, but that is a significant percentage. I also don't think everyone is experiencing the long time delays when discovering services like you are.

bleak/backends/winrt/client.py
dlech: We could leave them out for now and we can always add them later if there is a real use case for it.

hbldh: Removed it from connect. It is not needed there.

bleak/backends/winrt/client.py
hbldh: Added text warning about using it with devices created by oneself.

bleak/backends/winrt/client.py
hbldh: I tried both way presented in #623, but I didn't like either one of them now. Went with solution number two: the `win` dict in the end. It did however produce some problems documenting this in docstrings. I ended up with a list and emulating the emphasis on the dict keys to fit the RTD theme. If you have a better suggestion on how to document it, I will gladly change to that.

It looks ok in html. In the actual docstring, a bit cluttered.

examples/sensortag.py
```diff
@@ -98,7 +98,7 @@
 
 
 async def main(address):
-    async with BleakClient(address) as client:
+    async with BleakClient(address, win={"use_cached_services": True}) as client:
```
dlech, line 101: I don't have strong feelings about it but I kind of like `win=dict(use_cached_services=True)` to use the kwargs syntax for the `win` args too.

bleak/backends/winrt/client.py
```diff
@@ -133,13 +146,21 @@ def __init__(self, address_or_ble_device: Union[BLEDevice, str], **kwargs):
         self._session_closed_events: List[asyncio.Event] = []
         self._session: GattSession = None
 
+        if "address_type" in kwargs:
+            raise PendingDeprecationWarning(
```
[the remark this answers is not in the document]
hbldh: Ah, yes. Don't know what I thought of...

examples/sensortag.py
hbldh: I am not sure I understand. Do you mean only document them in the docstrings and not specify them in the method parameters? Still let them enter methods in the `kwargs`? Or do you mean how the dict is constructed with a `=` instead of a `:`?

dlech: I just mean how the dict is constructed.

hbldh: Ok. Changing.

dlech:
Hi @hbldh, I hope you don't mind that I updated this and made a few changes.

- Merged develop (only conflict was changelog)
- Made the backend-specific arg consistent with similar changes I made in the BlueZ backend recently.
- Added a 3rd state for `use_cached_services` that calls a different WinRT method overload. The docs aren't really clear if the no-arg version of the overload is the same as the 1-arg version with a specific argument or a different behavior. I asked a [similar question about the WinRT characteristic read API](https://github.com/MicrosoftDocs/winrt-api/issues/840) a while back and the answer was a 3rd behavior, so I figured it was a good idea to add it just in case it helps with some of the issues that have been reported with timeout while getting gatt services.

hbldh:
I do not mind at all. I'm sorry, I have been a bad maintainer of Bleak for quite some time...

It is good that is is merged. I was meant to do a log capture to see why this happened, but I never got around to it and forgot all about this...

I tried Bleak on my new Surface Pro with Windows 11 and had trouble getting my devices disconnected, so I will be doing some Bleak debugging in the near future. At least I am planning on doing that...

dlech:
Great, it will be good to get this merged.

> I tried Bleak on my new Surface Pro with Windows 11 and had trouble getting my devices disconnected, so I will be doing some Bleak debugging in the near future. At least I am planning on doing that...

I noticed some time ago that it was taking about 30 seconds for devices to disconnect on Windows 11 but today, it seems much better so maybe Microsoft has fixed this recently?

PR #706, opened 2021-12-30
Use predefined service uuids

mkoderer:
On macos the service_uuids need to be filled. Main issue for
beginners is to retrieve some valid uuids. Instead of using an
external tool like nRF we could use predefined ones in bleak/uuids.

Implementation works with 128bit and 16bit UUIDs.

Related-to #635

Signed-off-by: Marc Koderer <marc@koderer.com>

### DIFF
--- a/examples/detection_callback.py
+++ b/examples/detection_callback.py
@@ -15,6 +15,7 @@ import sys
 from bleak import BleakScanner
 from bleak.backends.device import BLEDevice
 from bleak.backends.scanner import AdvertisementData
+from bleak.uuids import uuid16_dict, uuid128_dict
 
 logger = logging.getLogger(__name__)
 
@@ -24,6 +25,16 @@ def simple_callback(device: BLEDevice, advertisement_data: AdvertisementData):
 
 
 async def main(service_uuids):
+    if len(service_uuids) > 0 and service_uuids[0] == "all":
+        # in Macos Monterey the service_uuids need to be specified.
+        # Instead of discovering valid uuids with a different tool
+        # you can add `all` as argument and it will add all defined
+        # uuid's from bleak/uuids as a starting point.
+        logger.info("Adding all known service uuids")
+        service_uuids.pop(0)
+        for item in uuid16_dict:
+            service_uuids.append("{0:04x}".format(item))
+        service_uuids.extend(uuid128_dict.keys())
     scanner = BleakScanner(service_uuids=service_uuids)
     scanner.register_detection_callback(simple_callback)
 


examples/detection_callback.py
```diff
@@ -24,6 +25,13 @@ def simple_callback(device: BLEDevice, advertisement_data: AdvertisementData):
 
 
 async def main(service_uuids):
+    if service_uuids[0] == "all":
```
dlech, line 28: Won't this raise a `IndexError` if `service_uuids` is empty?

examples/detection_callback.py
```diff
@@ -24,6 +25,13 @@ def simple_callback(device: BLEDevice, advertisement_data: AdvertisementData):
...
+        logger.info("Adding all known service_uuids")
```
dlech, line 29: can leave out `_` in the message.

examples/detection_callback.py
```diff
@@ -24,6 +25,13 @@ def simple_callback(device: BLEDevice, advertisement_data: AdvertisementData):
...
+        service_uuids.pop(0)
+        for item in uuid16_dict:
+            service_uuids.append("{0:04x}".format(item))
```
dlech, line 32: f-strings are ok too

examples/detection_callback.py
```diff
@@ -24,6 +25,13 @@ def simple_callback(device: BLEDevice, advertisement_data: AdvertisementData):
...
+        for item in uuid128_dict:
+            service_uuids.append(item)
```
dlech, line 34: If you like one-liners.

```suggestion
            service_uuids.extend(uuid128_dict.keys())
```

dlech (commented):
This is a great idea! Can we add some comments about why this is needed?

I found one potential problem. The rest of my comments are bike-shedding, so feel free to ignore if you don't think they make it better.

examples/detection_callback.py
mkoderer: yep you are right. Fixed.

examples/detection_callback.py
mkoderer: done

examples/detection_callback.py
mkoderer: done

dlech:
Thanks for updating. The `black` code formatter is complaining about one line. Maybe there is a trailing newline or a tab instead of spaces?

mkoderer:
Should pass now

dlech:
Thanks!

hbldh:
Very good, thank you! This is an essential addition to the example.

PR #710, opened 2021-12-31
github: convert android build to manual trigger

dlech:
The Android build is failing and I don't have time to fix it. This converts it to a manual trigger so that it doesn't cause pull requests to go red but keeps the job so we can fix it later.


hbldh:
The android solution should not break anything. I do not regard it as an OS with a stable Python option like the others. I will try to get to it during the docs work...

PR #715, opened 2022-01-03
small bug fix to set the metadata on the device on discovery callback

dlee1j1:

   guidelines in that document.
3. The pull request should work for Python 3.7+ on the following platforms:
    - OS X / macOS >= 10.11

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -33,6 +33,7 @@ Fixed
   of the same characteristic are used. Fixes #675.
 * Fixed reading a characteristic on CoreBluetooth backend also triggers notification
   callback.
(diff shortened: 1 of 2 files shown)


dlech:
This information is already in the `AdvertismentData` parameter of `detection_callback` so it would be preferred to use that instead. `metadata` was a bit of a hack to get the advertising data before we had a proper `detection_callback`. But, I guess this change doesn't hurt anything either.

Could you also add a changelog entry describing the change?

dlee1j1:
Got it. My understanding from the documentation indicates that metadata is the preferred way to get the data rather than advertisement.

dlech:
Thanks!

Issue #716, opened 2022-01-03
example/detection_callback.py doesn't work with all on bluez


mkoderer:
* bleak version:
develop branch
* Python version:
3.9.2
* Operating System:
Linux raspberrypi 5.10.89-v7+ #1506 SMP Fri Dec 31 14:38:59 GMT 2021 armv7l GNU/Linux
* BlueZ version (`bluetoothctl -v`) in case of Linux:
bluetoothctl: 5.55

### Description

Registering all known uuids leads to an exception:

```
> python examples/detection_callback.py all
2022-01-03 19:38:50,872 __main__ INFO: Adding all known service uuids
Traceback (most recent call last):
  File "/home/pi/git/bleak/examples/detection_callback.py", line 55, in <module>
    asyncio.run(main(service_uuids))
  File "/usr/lib/python3.9/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/usr/lib/python3.9/asyncio/base_events.py", line 642, in run_until_complete
    return future.result()
  File "/home/pi/git/bleak/examples/detection_callback.py", line 44, in main
    await scanner.start()
  File "/home/pi/git/bleak/.venv/lib/python3.9/site-packages/bleak-0.14.0a1-py3.9.egg/bleak/backends/bluezdbus/scanner.py", line 166, in start
    assert_reply(reply)
  File "/home/pi/git/bleak/.venv/lib/python3.9/site-packages/bleak-0.14.0a1-py3.9.egg/bleak/backends/bluezdbus/utils.py", line 23, in assert_reply
    raise BleakDBusError(reply.error_name, reply.body)
bleak.exc.BleakDBusError: [org.bluez.Error.InProgress] Operation already in progress
```

dlech:
Interesting. This seems to trigger a bug in BlueZ. After running the program with `all`, then even using a single UUID triggers the error until `bluetoothd` is restarted. I'm guessing that sending it so many UUIDs overwhelms BlueZ somehow.

I'm not too worried about this since `all` is just a hack to work around a macOS 12 bug and should never be used otherwise.

mkoderer:
@dlech Should we active the "all" behavior by default for Mac (without any parameter)?

The bluez bug is interesting it starts when you specify more than 64 service_uuids.
I will try to reproduce it  directly and update the version

mkoderer:
It's an clear bluez bug/limitation:

```
$ bluetoothctl
[bluetooth]# menu scan
[...]

PR #717, opened 2022-01-03
Limit "all" option to Mac.

dlech:
Adding 100s of UUIDs to the scan filter on BlueZ causes BlueZ to fail
with:

    File "examples/detection_callback.py", line 53, in <module>
        asyncio.run(main(service_uuids))
    File "/usr/lib/python3.8/asyncio/runners.py", line 44, in run
    File "/usr/lib/python3.8/asyncio/base_events.py", line 616, in run_until_complete
        return future.result()
    File "examples/detection_callback.py", line 42, in main
        await scanner.start()
    File "/home/david/work/pybricks/bleak/bleak/backends/bluezdbus/scanner.py", line 169, in start
        assert_reply(reply)
    File "/home/david/work/pybricks/bleak/bleak/backends/bluezdbus/utils.py", line 23, in assert_reply
        raise BleakDBusError(reply.error_name, reply.body)
    bleak.exc.BleakDBusError: [org.bluez.Error.InProgress] Operation already in progress

`bluetoothd` has to be restarted to clear the error.

So don't allow this option on OS other than Mac.

Fixes: https://github.com/hbldh/bleak/issues/716

### DIFF
--- a/examples/detection_callback.py
+++ b/examples/detection_callback.py
@@ -10,6 +10,7 @@ Updated on 2020-10-11 by bernstern <bernie@allthenticate.net>
 
 import asyncio
 import logging
+import platform
 import sys
 
 from bleak import BleakScanner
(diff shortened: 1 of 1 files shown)


hbldh (approved):
Does the adding of this many uuids make scanning in Monterey slower than previous macOS versions? Regardless, it is needed for proper functionality...

Do we need to update and expand the list of uuids to cover devices not in that list? And in that case, where do we find them?

dlech:
> Does the adding of this many uuids make scanning in Monterey slower than previous macOS versions?


It should not affect the "speed" at all. It is just a filter that excludes advertising data that doesn't match any of the UUIDs.

> Do we need to update and expand the list of uuids to cover devices not in that list? And in that case, where do we find them?

Getting the advertised UUIDs for every BLE device ever made seems like an impossible task. I would just leave it up to users to make a pull request to add a UUID if it is important to them.

hbldh:
I agree. I dislike leaving difficult things like "knowing your device before connecting to it and seeing what's on it" to the user, but I don't think it is a viable option to try to include more known uuids in bleak without PRs.

Bleak on new macOS becomes less beginner-friendly than I would like. But this is as good as it gets, and hopefully good enough for most applications.

examples/detection_callback.py
```diff
@@ -25,11 +25,18 @@ def simple_callback(device: BLEDevice, advertisement_data: AdvertisementData):
 
 
 async def main(service_uuids):
-    if len(service_uuids) > 0 and service_uuids[0] == "all":
-        # in Macos Monterey the service_uuids need to be specified.
+    if sys.platform == "darwin" and service_uuids and service_uuids[0] == "all":
```
mkoderer, line 28: I would suggest that we enable this by default (without any parameter). And we use the logger.info block (maybe logging.warining would be better) below to give the user an hint that we only scan for predefined uuids.

dlech: Updated.

mkoderer:
Thanks David - LGTM

Issue #729, opened 2022-01-11
Cannot scan for connected devices


MagnusModig:
* bleak version: 0.13.0
* Python version: 3.9
* Operating System: Win 11
* BlueZ version (`bluetoothctl -v`) in case of Linux:

### Description

I try to scan connected devices.

### What I Did

Ran this code...
```python
import asyncio
from bleak import BleakScanner

async def main():
    devices = await BleakScanner.discover()
    for d in devices:
        print(d)

asyncio.run(main())
```

But I get this error..

```
Traceback (most recent call last):
  File "D:\scan.py", line 2, in <module>
    from bleak import BleakScanner
  File "D:\bleak.py", line 6, in <module>
    from bleak import BleakClient, BleakScanner
ImportError: cannot import name 'BleakClient' from partially initialized module 'bleak' (most likely due to a circular import) (D:\bleak.py)`

```

dlech:
You cannot name your script `bleak.py`. This causes the circular import.

MagnusModig:
Thanks, that helped to rename the file

Issue #515, opened 2021-04-19
ImportError: cannot import name 'BleakScanner'


Gladiatech:
* bleak version:0.11.0
* Python version:3.8.4
* Operating System:Windows 10 19041.867
* BlueZ version (`bluetoothctl -v`) in case of Linux:

### Description

I've just installed bleak, trying to execute the first script 

```python
import asyncio
from bleak import BleakClient

address = "24:71:89:cc:09:05"
MODEL_NBR_UUID = "00002a24-0000-1000-8000-00805f9b34fb"

async def run(address):
    async with BleakClient(address) as client:
        model_number = await client.read_gatt_char(MODEL_NBR_UUID)
        print("Model Number: {0}".format("".join(map(chr, model_number))))

loop = asyncio.get_event_loop()
loop.run_until_complete(run(address))
```

I get : ImportError: cannot import name "BleakScanner" from partially initialized module 'bleak' (most likely due to a circular import) (C:\Users\Fra\Desktop\bleak.py)

I tried to uninstall/reinstall again and change python version from 3.9 but not luck

cheers

PR #730, opened 2022-01-11
docs: warn against bleak.py

dlech:
We frequently get bug reports about circular import errors. This is due to naming your script `bleak.py`. So put warnings  everywhere so hopefully people won't miss it.

Fixes: #729
Fixes: #704
Fixes: #515


hbldh:
This is indeed a universal Python issue, one that I do not think any level of documentation in Bleak will solve. But it is worth to try.

Issue #736, opened 2022-01-13
Some BLE Devices showing up as Unknown


quasi635:
* bleak version: 0.14.1
* Python version: 3.7.9
* Operating System: Windows 10

### Description

After upgrading to bleak 0.14.1 from 0.13.0, my devices no longer show up when doing `discover()`

### What I Did

I ran the example code
```python
import asyncio
from bleak import BleakScanner

async def main():
    devices = await BleakScanner.discover()
    for d in devices:
        print(d)

asyncio.run(main())
```

On 0.13.0 I see a number of devices, including:
```
A0:E6:F8:3A:84:A1: Temperature 835-041>19
E0:7D:EA:6B:1E:5A: //code.Node 020-122>G0
7E:68:78:BB:3F:32: Apple, Inc. (...)
69:BB:ED:8E:0A:97: Microsoft (...)
34:3C:5B:F3:EC:52: Microsoft (...)
75:59:95:91:07:00: Apple, Inc. (...)
7C:BE:1C:6C:9E:C1: Apple, Inc. (...)
A0:9E:1A:34:DB:6D: Polar H7 34DB6D24
```

On 0.14.1 those same devices show up as:
```
A0:E6:F8:3A:84:A1: Unknown
E0:7D:EA:6B:1E:5A: Unknown
7E:68:78:BB:3F:32: Apple, Inc. (...)
69:BB:ED:8E:0A:97: Microsoft (...)
34:3C:5B:F3:EC:52: Microsoft (...)
75:59:95:91:07:00: Apple, Inc. (...)
7C:BE:1C:6C:9E:C1: Apple, Inc. (...)
A0:9E:1A:34:DB:6D: Polar H7 34DB6D24
```

Any help? I can't figure out what changed between the two versions to cause this. Both version show all of the Apple/Microsoft/Polar devices around me (I removed the byte value and replaced with three dots).

dlech:
Can you enable logging and share the results?

The change that broke this is probably https://github.com/hbldh/bleak/commit/3f26be60c3e69d61c5327a0fe0bbf0aee20f1456. In particular, the logic that determines the name is at: https://github.com/hbldh/bleak/blob/3f26be60c3e69d61c5327a0fe0bbf0aee20f1456/bleak/backends/winrt/scanner.py#L211-L232

dlech:
I could see how this might fail if `args.advertisement.local_name` returns an empty string instead of `None`.

quasi635:
> 

[...]

PR #748, opened 2022-01-26
winrt: fix name is 'Unknown'

dlech:
Commit 3f26be60c "winrt: combine adv data and scan rsp data" incorrectly assumed that `local_name` would return None instead of '' when no name was present in the advertising data. This changes the test to not strictly check for None.

Fixes: https://github.com/hbldh/bleak/issues/736


kva85:
This works for me.

PR #755, opened 2022-01-30
Add a Multiple Device Troubleshooting Section

### LINKED ISSUE #754: Collect notify data from two devices (different MACs) with same UUIDs + Identify devices
* bleak version: 0.14.2
* Python version: 3.9.7
* Operating System: Ubuntu 21.10
* BlueZ version (`bluetoothctl -v`) in case of Linux: 5.60

### Description

*I apologize in advance, I am sure the library is working fine here, this is just my limited BLE knowledge causing issues*
- I will hopefully repay any help here with an example / documentation update

#### My use case / setup

I have two Lithionics Batteries that each have different Bluetooth MAC addresses that I wish to pull the telemetry from each. I wish to store it in a way so I can create counters in prometheus for each respective device. They send ascii strings that I can split into a data class and then I will export to Prometheus (time series DB) via it's exporter API.

When I try to create a client per device via `BleakClient(mac_address)` the second device always fails to connect / stream data. It seems to just sit there blocked. So I then thought the one notify callback must get the data from both batteries, but that does not seem the case either as I always get the same "sender id" sent to the handler:

```
[2022-01-30 17:09:10,294] INFO: Received Li3 telementary from 17 (li3.py:50)
```
- Always client 17

My code connecting to the two device can be seen here: https://github.com/cooperlees/vanD/blob/main/src/vand/li3.py#L71
- The two devices UUIDs etc. can be found in the config: https://github.com/cooperlees/vanD/blob/main/vand.json#L9

What's the right way to get data and identify each battery (device) here so I can collect respective statistics? Do I need separate processes for each device?

Screen Shot 2022-01-30 at 8 57 01 AM

### What I Did

I run a daemon that listens and converts the ascii bluetooth data into a data class and then exposes via the Prometheus Export API. For one device this is working, but I can't get it to work for two.

Any tips / tricks / info here would be appreciated. I thank you for this library and again, apaologize as I bet this is just all my limited knowledge of Bluetooth LE. This is my first time ever working with it.

cooperlees:
- Add information / tips to help with:
  - Connection issues to multiple devices
  - Pass more parameters to notify callback function

Fixes #754

### DIFF
--- a/docs/troubleshooting.rst
+++ b/docs/troubleshooting.rst
@@ -37,12 +37,39 @@ Windows Command Prompt::
 
 Then run your Python script in the same terminal.
 
+
 ------------------------------------------------
 Handling multiple devices with same Service UUID
 ------------------------------------------------
 
 If you're having difficulty connecting to multiple devices, try to do a scan first and
 pass the returned ``BLEDevice`` objects to ``BleakClient``.
+
+Issue #759 might fix this in the future.
+
+Python::
+
+    import asyncio
+    from typing import Sequence
+
+    from bleak import BleakClient, BleakScanner
+    from bleak.backends.device import BLEDevice
+
+
+    async def find_all_devices_services()
+        scanner = BleakScanner()
+        devices: Sequence[BLEDevice] = scanner.discover(timeout=5.0)
+        for d in devices:
+            async with BleakClient(d) as client:
+                print(await client.get_services())
+
+
+    asyncio.run(find_all_devices_services())
+
+
+-----------------------------------------
+Pass more parameters to a notify callback
+-----------------------------------------
 
 If you need a way to pass more parameters to the notify callback, please use
 ``functools.partial`` to pass in more arguments.
@@ -51,7 +78,7 @@ Python::
 
     from functools import partial
 
-    from Bleak import BleakClient
+    from bleak import BleakClient
 
 
     def my_notification_callback_with_client_input(
--- a/docs/troubleshooting.rst
+++ b/docs/troubleshooting.rst
@@ -38,14 +38,12 @@ Windows Command Prompt::
 Then run your Python script in the same terminal.
 
 
-------------------------------------------------
-Handling multiple devices with same Service UUID
-------------------------------------------------
+-----------------------------------------------
+Connecting to multiple devices at the same time
+-----------------------------------------------
 
 If you're having difficulty connecting to multiple devices, try to do a scan first and
-pass the returned ``BLEDevice`` objects to ``BleakClient``.
-
-Issue #759 might fix this in the future.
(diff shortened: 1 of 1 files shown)


dlech:
> Add information / tips to help with:
>     * Connection issues to multiple devices
>     * Pass more parameters to notify callback function

This seems like to unrelated problems, so I would expect two separate sections in the docs.

> * Pass more parameters to notify callback function


I'm wondering if we should fix this in code instead of fixing it in documentation. See my previous comment at https://github.com/hbldh/bleak/issues/754#issuecomment-1025196739.

dlech:
> > * Pass more parameters to notify callback function
> 
> I'm wondering if we should fix this in code instead of "fixing" it in documentation. See my previous comment at [#754 (comment)](https://github.com/hbldh/bleak/issues/754#issuecomment-1025196739).

I opened https://github.com/hbldh/bleak/issues/759 to discuss this separately.

dlech:
OK, I will try again.

> > Add information / tips to help with:
> > * Connection issues to multiple devices
> > * Pass more parameters to notify callback function
> 
> This seems like two unrelated problems, so I would expect two separate sections in the docs.

These are two unrelated issues, so combining them into one troubleshooting section seems like it would make it harder to find the answer for one or the other.

> * Connection issues to multiple devices

There is no debate about this issue. Having a troubleshooting section that explains that you should use a separate `BleakScanner` to scan first and then stop scanning before attempting to connect when you have multiple devices sounds like a good idea to me.

> * 
> 
> * Pass more parameters to notify callback function

I don't have a strong opinion one way or the other if we include this now or wait to see what the resolution of #759 is. But if we include it now, it should be a separate section from the scanning issue.

docs/troubleshooting.rst
```diff
@@ -38,6 +38,63 @@ Windows Command Prompt::
 Then run your Python script in the same terminal.
 
 
+------------------------------------------------
+Handling multiple devices with same Service UUID
```
dlech, line 42: I think this applies to multiple devices in general, not just ones with the same service UUID.

Suggested title: Connecting to multiple devices at the same time.

docs/troubleshooting.rst
```diff
@@ -38,6 +38,63 @@ Windows Command Prompt::
...
+------------------------------------------------
+
+If you're having difficulty connecting to multiple devices, try to do a scan first and
+pass the returned ``BLEDevice`` objects to ``BleakClient``.
+
+Issue #759 might fix this in the future.
```
dlech, line 48: Issue 759 doesn't apply here, it applies to the section below.

dlech (commented):
Thanks for updating. I just have a few suggestions.

dlech:
Thanks!

Issue #699, opened 2021-12-22
Unable to unpair with device


harshdeepsehgal:
* bleak version: 0.13.0
* Python version: 3.8.5
* Operating System: Win10

### Description

I am trying to pair and unpair with an Arduino Nano 33 BLE. I want to pair since there is an HID service I want to use which won't work using the standard connect method since it throws an Access Denied.

### What I Did

Pairing works fine and I can use it as an HID. I am planning to use unpair for disconnecting. Calling unpair does nothing.
The issue seems to be that unpair uses the following boolean variable -
`self._requester.device_information.pairing.is_paired`

```
        client=BleakClientWinRT(device_address)
        is_connected=await client.connect()
        is_paired=await client.pair(2)
        print(is_paired)
        await asyncio.sleep(10)
        await client.unpair()
```

Even though the `pair` function returns true, the above boolean variable remains false. Possibly something on the firmware end and I can't figure out how to fix that.

Is it possible to have another boolean instance variable to store the pairing status at the software end and raise an exception if unpair fails?
Is it also possible that the above boolean variable isn't being refreshed after pairing in Bleak?

This is a bit urgent so I'd appreciate some quick help!
TIA

dlech:
Pairing isn't implemented in Windows yet. See #640.

hbldh:
I did implement pairing in my first winrt reimplementation of the previous backend. There are some elements of it there, right? A just works-pairing?

Try removing the offending check in your local installation of Bleak and see if something breaks. However, I do not think that the Bleak `unpair` method will lead to a disconnection. You should disconnect with the `disconnect` method instead.

harshdeepsehgal:
Yes, the elements of your old implementation do work.
From what I have understood, the current implementation of pair/unpair uses UWP Bluetooth. Unpairing using that leads to automatic disconnection and I'm hoping would also trigger the disconnect callback (have not checked this yet).
I tried removing the check on my local installation and it works fine. I could just put in a try/except just in case.

In case there are no immediate plans to re-implement this, maybe I could open a pull request with the temporary changes to make this work for more devices, by removing the check and adding a try/except instead.

I have to use this in some production-level code and it would be tedious to give my own build of bleak to everyone.

harshdeepsehgal:
An update:
I tried refreshing the `_requester` object before the check, and it seems that fixes it as well. So I guess the issue isn't device-specific after all. The object needs to be refreshed after pairing/unpairing.

[...]

PR #757, opened 2022-02-04
Local device info object for pair/unpair

harshdeepsehgal:
Added a local `DeviceInformation` object to make sure flags are updated to enable pairing and unpairing.

Solves #699

### DIFF
--- a/bleak/backends/winrt/client.py
+++ b/bleak/backends/winrt/client.py
@@ -31,6 +31,7 @@ from bleak_winrt.windows.devices.bluetooth.genericattributeprofile import (
     GattSession,
 )
 from bleak_winrt.windows.devices.enumeration import (
+    DeviceInformation,
     DevicePairingKinds,
     DevicePairingResultStatus,
     DeviceUnpairingResultStatus,
@@ -350,15 +351,18 @@ class BleakClientWinRT(BaseBleakClient):
             Boolean regarding success of pairing.
 
         """
-
+        # New local device information object created since the object from the requester isn't updated
+        device_information = await DeviceInformation.create_from_id_async(
+            self._requester.device_information.id
+        )
         if (
-            self._requester.device_information.pairing.can_pair
-            and not self._requester.device_information.pairing.is_paired
+            device_information.pairing.can_pair
+            and not device_information.pairing.is_paired
         ):
 
             # Currently only supporting Just Works solutions...
             ceremony = DevicePairingKinds.CONFIRM_ONLY
-            custom_pairing = self._requester.device_information.pairing.custom
+            custom_pairing = device_information.pairing.custom
 
             def handler(sender, args):
                 args.accept()
@@ -395,7 +399,7 @@ class BleakClientWinRT(BaseBleakClient):
                 )
                 return True
         else:
-            return self._requester.device_information.pairing.is_paired
+            return device_information.pairing.is_paired
 
     async def unpair(self) -> bool:
         """Attempts to unpair from the device.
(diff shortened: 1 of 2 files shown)


bleak/backends/winrt/client.py
```diff
@@ -406,10 +408,10 @@ async def unpair(self) -> bool:
             Boolean on whether the unparing was successful.
 
         """
-
-        if self._requester.device_information.pairing.is_paired:
+        device_information=await DeviceInformation.create_from_id_async(self._requester.device_information.id)
```
dlech, line 411: Can we add a comment here explaining that the old `device_information` is out of date?

bleak/backends/winrt/client.py
```diff
@@ -395,7 +397,7 @@ def handler(sender, args):
                 )
                 return True
         else:
-            return self._requester.device_information.pairing.is_paired
+            return device_information.pairing.is_paired
```
dlech, line 400: Should we just return `False` here?

dlech (commented):
Thanks!

Can you run `black` on the code to fix the formatting and add a changelog entry?

bleak/backends/winrt/client.py
dlech: Never mind, I see how this could be `True` now.

bleak/backends/winrt/client.py
harshdeepsehgal: Done

harshdeepsehgal:
I have added the changelog entry. I did run black, it has reformatted files, but somehow it still fails the check.

dlech:
Black just came out with a new stable version. The CI is using `black-22.1.0`. Maybe you are using an older version? Or maybe we haven't run the new version on the code yet?

harshdeepsehgal:
Yes I was running an older version of black. It passed the checks now.

bleak/backends/corebluetooth/characteristic.py
```diff
@@ -63,7 +63,7 @@ def __init__(self, obj: CBCharacteristic):
         # self.__props = obj.properties()
         self.__props: List[str] = [
             _GattCharacteristicsPropertiesEnum[v][0]
-            for v in [2 ** n for n in range(10)]
+            for v in [2**n for n in range(10)]
```
dlech, line 66: It would be nice to put unrelated black format changes in a separate commit.

bleak/backends/winrt/client.py
```diff
@@ -350,15 +351,18 @@ async def pair(self, protection_level: int = None, **kwargs) -> bool:
             Boolean regarding success of pairing.
 
         """
-
+        # New local device information oject created since the object from the requester isn't updated
```
dlech, line 354: spelling

```suggestion
        # New local device information object created since the object from the requester isn't updated
```

bleak/backends/winrt/client.py
```diff
@@ -407,10 +411,12 @@ async def unpair(self) -> bool:
 
         """
 
-        if self._requester.device_information.pairing.is_paired:
-            unpairing_result = (
-                await self._requester.device_information.pairing.unpair_async()
-            )
+        # New local device information oject created since the object from the requester isn't updated
```
dlech, line 414:
```suggestion
        # New local device information object created since the object from the requester isn't updated
```

bleak/backends/corebluetooth/characteristic.py
dlech: Actually, I'll just go ahead and do that, then you can just rebase on the develop branch.

dlech:
Thanks!

harshdeepsehgal:
Happy to contribute :)

Issue #762, opened 2022-02-11
Is there a leak in corebluetooth or am I using it wrong?


Jc2k:
* bleak version: 0.14.2
* Python version: 3.9
* Operating System: macOS 10.15.7
* BlueZ version (`bluetoothctl -v`) in case of Linux:

### Description

I am working on a HomeKit BLE client for Home Assistant to support things like Eve or Nanoleaf "home automation" devices. Polling is working well, and so are what Apple calls "disconnected" events (incrementing a counter in advertising data to trigger a poll). To finish this off i'm looking to add support for "connected" events (`start_notify`).

For context, HomeKit devices don't do pairing, they implement their own encryption scheme on top of unpaired GATT write/read. HomeKit devices tend to close the connection after less than 30s-ish (haven't timed this). They stay connected in case e.g. you were dragging a brightness slider on your phone and wanted fast successive updates. I don't know if thats normal for BLE or not, but it means that short of polling every 10s I need to reconnect to the device a lot.

I'm at a bit of a loss with I guess what you would call lifetime and ownership issues.

* Is it expected that a `BleakClient` is one-shot, that if the BLE device becomes disconnected I need a new one?
* Or, Is it expected that I can reuse the same `BleakClient`, and "reconnect" if a BLE device drops the connection?
  * If i reuse it, do i have to re-register the notify callbacks?
* How do I call `stop_notify` if the connection was closed by the device?

I've guessed and struggled to find an ownershipe/lifetime model that works for all my test cases. The closest seems to be that `BleakClient` can only be used once, but there is a bug that means they are leaking right now. So given the frequency of disconnects, if i poll one device once a minute then i'll leak 1440 `BleakClient` instances a day.

### What I Did

This is a reproduction of the things I tried. I don't fully remember them all, I think there was also a situation where I had no errors but no notifications which I can't reproduce.

I believe this simple test case is working, but I don't know if its considered valid (as it reuses a BleakClient).

Note that characteristic 71 is just a random one that happened to have `indicate` in its properties.

```python
from bleak import BleakClient
import asyncio

async def main():
    def _disconnected(*args, **kwargs):
        print("DISCONNECT", args, kwargs)

    client = BleakClient("E8EA177F-0419-44A9-B1DD-D96F348B1184")
    client.set_disconnected_callback(_disconnected)

    def _callback(*args, **kargs):
        print("NOTIFY", args, kwargs)

    async with client:
        char = client.services.characteristics[71]
[...]

PR #763, opened 2022-02-11
Fix BleakClient leak on macOS

Jc2k:
Please see https://github.com/hbldh/bleak/issues/762 which contains my "working out" for this problem.

Previously the disconnect handler was only called if the peripheral wasn't already disconnected. This is a problem when devices might disconnect before you are done with them.

It could lead to a leak where the handler was never disconnected and you could end up with duplicate disconnect callbacks for the same device.

The client objects themselves also leaked.

The existing code had the effect of masking a duplicate callback. corebluetooth seems to send a notification with no error when we disconnect the disconnect handler. So users previously only saw disconnect events caused by errors (the device going away).

Now that we disconnect from the handler we get an extra notification. To avoid an API break, we now only invoke the users callback if there was an error condition. This simulates the old behaviour.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,7 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 
 * Fixed reading the battery level returns a zero-filled bytearray on linux with bluez >= 5.48 Fixes #750
 * Fixed unpairing does not work on windows with winrt. Fixes #699
+* Fixed a leak of ``BleakClient`` instances in macOS. Fixes #762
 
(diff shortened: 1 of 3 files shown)


dlech:
Thanks for the pull request. I'm not sure yet that this is the right way to fix the problem without breaking other things. I thought I was seeing some similar issues with disconnecting on macOS in my own program last week. Unfortunately, I'm no longer able to reproduce this issue today. I also tried the example from #762 but was not able to reproduce the problem.

While looking into it, though, I found some other problem. Could you try #778 and see if it makes any difference with this problem?

dlech:
Closing due to no activity.

PR #796, opened 2022-03-30
Use AuthAnnonymous() when connecting to DBus over TCP

someDude12341:
DBus-next defaults to using AuthExternal() when autenticating but this only works when we are working on the same machine.

Tested on Ubuntu 21.10.

### DIFF
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -30,7 +30,7 @@ from . import defs
 from .characteristic import BleakGATTCharacteristicBlueZDBus
 from .manager import get_global_bluez_manager
 from .scanner import BleakScannerBlueZDBus
-from .utils import assert_reply, get_dbus_authenticator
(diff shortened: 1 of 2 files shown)


dlech:
This seems like it could be a security hole. What is the use case? It seems like it would be better to run Bleak on the same machine as BlueZ and use some sort of RPC instead.

someDude12341:
> This seems like it could be a security hole.

Absolutely, just not in Bleak. For this to be an thing you would first have to edit /etc/dbus-1/system-local.conf and /lib/systemd/system/dbus.socket on the host machine and re-define DBUS_SYSTEM_BUS_ADDRESS on your local machine.

I would argue that if someone make these kind of changes and don't know better than to remain on an isolated LAN then bigger security holes are at stake.

dlech:
We recently made some similar changes in https://github.com/hbldh/bleak/pull/1182, so if this is still needed, would like to add a `BLEAK_DBUS_AUTH_ANON` environment variable check in the new `get_dbus_authenticator()` function.

Issue #822, opened 2022-05-02
Improved type annotations


davclark:
* bleak version: latest
* Python version: 3.9.11
* Operating System: Linux / CircuitPython
* BlueZ version (`bluetoothctl -v`) in case of Linux: not relevant

### Description

At PyCon 2022, we're sprinting to improve type annotations in the adafruit circuitpython libraries. bleak is one of the foundational libraries in the adafruit bluetooth stack, and so we can't quite get complete typing without bleak *also* having complete typing (or alternatively, stubs).

Is there interest in improved typing? We could produce a PR, but I wanted to check if that would be likely to be accepted first!

### What I Did

- Clone the repo
- Install requirements (including dev)
- Install mypy (which is not currently in either of the above)
- Check types with mypy

```
mypy bleak
```

Found 85 errors in 22 files (let me know if you'd like the full output pasted here)

dlech:
Bleak should already have complete type hints. Happy to take a PR to fix mypy errors.

It could be better if we did something like #582 though.

JPHutchins:
I've put about 1.5 hours into an initial investigation.

For hacks to make mypy "pass" (but still be useful) I would estimate 40 - 80 hours of work.

Per backend it varies.  The android implementation is particularly worrisome and I'm certain that the static analysis is revealing true runtime bugs.

Generally, the architecture itself, due to supporting multiple backends, is not very strongly typed.  For example, while interfaces are defined, they are not generic and the implementations tend to "add what they need".

What I'd recommend is splitting up the approach:

* sunset python 3.8 support -  it's EOL is less than a year away!
* define statically typed interfaces that are consistent across all backends - careful test coverage!
* improve pyproject.toml "groups" to allow for "cross platform" static analysis - e.g. install windows dependencies on linux, install android dependencies on mac, etc. - dev and CI environments only, of course
* start adding mypy requirements one module at a time; probably the scope would be 15-20 PRs making particular folders pass static analysis and adding to the CI
  * probably these steps would update the interfaces definitions, introduce Generics and TypeVars as needed, etc.
  
Total maybe 120 hours for one person.  For the project, more, managing reviews and test coverage of 30-40 PRs.

For reference, here's the initial mypy error list as of posting:

```
typings/CoreBluetooth/__init__.pyi:3: error: Skipping analyzing ".libdispatch": module is installed, but missing library stubs or py.typed marker  [import-untyped]
[...]

Issue #582, opened 2021-06-27
Want to use BleakClient within bleak package


jackjansen:
I would like to have at least `BleakClient` available within bleak, in other words: move the platform-dependent imports from `__init__.py` to a new module (say `api.py`). `__init__.py` would then import them from there, but so could other modules within bleak.

If you're willing accept a pull request for this let me know.

> The reason I want this is that I'm working on a Pythonic interface to get and put values, using 2904 descriptors to marshal and unmarshal values, plus accessor classes for services and characteristics. Hopefully you'll eventually be able to do things like 

```
async with PythonicBleakClient(address, timeout=10) as client:
    batteryLevel = await client.services['Battery Service'].characteristics['Battery Level'].get()
```

But `PythonicBleakClient` and the accessor classes are backend-independent, and I don't want to put all the functionality in `BaseBleakClient`.

hbldh:
I can understand your desire for this; I would also have liked to make Bleak more Pythonic and generate Service, Characteristic and Descriptor instances in runtime to match what the connected peripheral responds that it has. I failed when encountering multiple backends and I retreated to a lower level API instead, in order to get a cross-platform API working before moving on. I never got around to doing that.

One [main design criteria of Bleak](https://github.com/hbldh/bleak/blob/feature/wtd/docs/nutshell.rst) is that it should always be pip-installable and importable right off the bat, and I also want it to be as simple as `from bleak import BleakClient` and that the user should not be required to learn how to any other class in order to get something working. I am fine with moving the platform-dependent parts to another file, but I think that having a fully functional client class at root level is the best option here.

And, given that your `PythonicBleakClient` is truly backend-independent, then it could be implemented either in the `BaseBleakClient` or in a new inheriting class in between the `BaseBleakClient` and the backend implementations. Also, given that Bleak should always be importable in any backend, having platform dependent imports in the `__init__.py` should not pose any problem either. 

Somewhere down the line you will have to get dirty with the specific implementations anyway, so I cannot see the real world value of an independent import of the `BaseBleakClient`. I am not trying to be difficult here, but I do not see a way forward to achieve both what you want and what I want Bleak to be. If you find a path that I do not see I would gladly accept that PR.

A sidenote: there is an [add-on package](https://pypi.org/project/bleak-sigspec/) to Bleak that adds a rather large XML with specifications of services and characteristics. This might fit better in with that? Or being a separate package in a similar fashion?

jackjansen:
[...]

PR #833, opened 2022-05-19
Add py.typed file

ljodal:
This is needed for mypy to read the type signatures defined in the project, ref https://mypy.readthedocs.io/en/stable/installed_packages.html#creating-pep-561-compatible-packages

There's also a note about needing `zip_safe=False`, but I'm not to familiar with setuptools so I didn't want to add that unless it's strictly required.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,11 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 `Unreleased`_
 =============
 
+Changed
+-------
+
+* Add `py.typed` file so mypy discovers Bleak's type annotations
+
 
 `0.14.3`_ (2022-04-29)
 ======================
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -3,6 +3,7 @@ include CHANGELOG.rst
 include CONTRIBUTING.rst
 include LICENSE
 include README.rst
+include bleak/py.typed
 
 recursive-include tests *
 recursive-exclude * __pycache__


dlech:
Thanks! Could you also add a changelog entry?

> There's also a note about needing `zip_safe=False`, but I'm not to familiar with setuptools so I didn't want to add that unless it's strictly required.

To test this, you could run `python setup.py sdist bdist_wheel` (in an environment with `setuptools` and `wheel` packages installed) and inspect the resulting files.

ljodal:
Thanks for the quick feedback! Wasn't sure if this was big enough to warrant a changelog entry, but I'll get that fixed and check the zip_safe flag over the weekend :)

ljodal:
Hey, sorry about the late follow up here. I researched the zip_safe flag and it seems to only be relevant for .egg distraction, which from what I understood is deprecated anyway. Looking into the PRs when the warning was added to mypy it seems like it's not a common issue, so I'm inclined skip that bit. Does that seem reasonable?

dlech:
As long as the empty `py.typed` file ends up in the binary wheel, it's fine with me. :smile:

ljodal:
It does, running `python setup.py sdist bdist_wheel` it gets included in both the `.whl` and `.tar.gz` files :) I'll add a changelog entry and leave it at that.

Note that the typing experience with bleak isn't perfect after this, as noted in #822. At least in strict mode mypy will still complain:

```
scanner.py:5: error: Module "bleak" does not explicitly export attribute "BleakClient"; implicit reexport disabled  [attr-defined]
scanner.py:5: error: Module "bleak" does not explicitly export attribute "BleakScanner"; implicit reexport disabled  [attr-defined]
```

dlech:
Thanks!

dlech:
> Note that the typing experience with bleak isn't perfect after this, as noted in #822. At least in strict mode mypy will still complain:
> 
> ```
> scanner.py:5: error: Module "bleak" does not explicitly export attribute "BleakClient"; implicit reexport disabled  [attr-defined]
> scanner.py:5: error: Module "bleak" does not explicitly export attribute "BleakScanner"; implicit reexport disabled  [attr-defined]
> ```

Another related discussion on these two types in particular is #582. It is a rather large chore, but would be nice for consuming the API with static analysis tools like mypy and pyright.

ljodal:
Yeah, I saw that issue and completely agree :)

PR #894, opened 2022-07-25
Fix documentation of raised error

mdxs:
Simple fix in documentation of method describing the error that could be raised.


dlech:
Thanks again!

Issue #888, opened 2022-07-14
Deprecating `handle`


dlech:
In #883 we realized that Android has an solution for differentiating services/characteristics that isn't compatible with the current Bleak notion of an attribute handle. The reason we expose the `handle` property in Bleak to begin with is to differentiate between two services or characteristics that have the same UUID. To get the value, we are using the attribute handle of the service/characteristic. However, this value is not the same cross-platform - it can be off-by-one between OSes. On macOS, we are using a private ObjC property to get the value (which could disappear some day and break Bleak). On BlueZ we have to do some string scraping to get the value from the D-Bus object path. And as we have just realized, on Android, there isn't a way to get the value at all.

But Android actually has IMHO a rather elegant solution. Each service/characteristic is identified by two pieces of information, the UUID and an "instance id" that is essentially a 0 based index of the occurrences of that UUID.  We could use this idea in Bleak by allowing a tuple of a UUID and instance id as the `char_specifier` argument.

This seems much nicer that the existing situation. Currently, I think many users are just hard-coding the handle value, but as mentioned already, this doesn't work cross-platform since the value varies by OS, and it also breaks if the device has a firmware upgrade that rearranges the GATT attributes. The current way to avoid these pitfalls is to iterate all of the characteristics to find the one you want and then pass the characteristic instance itself as the `char_specifier`. But we could avoid the iteration step by allowing the new form of the argument as describe above.

If we deprecate the use of `handle`s, then we don't have to come up with a hack to fix #883 and we can just not implement handles on Android. (And there seems to be a decent amount of support for #759 which is going in the same direction of not using `handle`).

dlech:
According to a book I just read, using the 0x2901 "Characteristic User Description" is another recommended way to uniquely identify a characteristic. Also, in some cases, the service probably needs to be part of the characteristic "path".

jackjansen:
I think that completely hiding handles (if the backend supports them) is the wrong way to go, because really handles are the common low-level addressing mechanism for attributes and therefore for all higher level abstractions (services, characteristics, values, ...).

But steering users away from using handles if they don't have to is a good idea. For example, by not allow a bare integer handle as parameter to all types of calls, but introducing a `BleakGATTAttribute` type that encapsulates the handle, and allowing that in stead.

[...]

PR #904, opened 2022-07-28
Export BleakClient and BleakScanner as toplevel objects (#582)

jackjansen:
This pull request adds toplevel objects `BleakClient` and `BleakScanner` and documents them. The objects are basically the underlying backend objects for the current platform, but this way they can be subclassed within bleak (as well as allowing more understandable user documentation). See #582 and related issues for the rationale.

The pull request is not complete yet (I still have to remove the GitHub action changes), I'm opening it mainly for discussion.

One of the things I would like feedback on: I would like to follow a similar pattern for `BleakGATTService` and `BleakGATTCharacteristic`, but I'd want to hear the opinion of the bleak maintainers. The advantage, especially from a documentation/usability point of view, would be that all objects needed for normal use would be exported from the top level, and the whole `backends` business becomes essentially an implementation detail.


dlech:
> The advantage, especially from a documentation/usability point of view, would be that all objects needed for normal use would be exported from the top level, and the whole `backends` business becomes essentially an implementation detail.

Yes, I agree this would be a nice improvement.

bleak/_api.py
```diff
@@ -0,0 +1,117 @@
...
+            Optional function that will be called each time a device is
+            discovered or advertising data has changed.
+        service_uuids:
+            Optional list of service UUIDs to filter on. Only advertisements
+            containing this advertising data will be received.
+        scanning_mode:
+            Set to "passive" to avoid the "active" scanning mode.
+    """
+
+    pass
```
dlech, line 94: I was hoping that we could put all of the methods/properties here so all of the documentation would be in once place and not have to be repeated in the backends. And also so that when you jump to the definition of a method/property using an IDE you end up here.

docs/api.rst
```diff
@@ -1,77 +1,119 @@
...
+Client interface
+~~~~~~~~~~~~~~~~
+
+.. autoclass:: bleak.BleakClient
     :members:
+    :inherited-members:
 
-Windows
-~~~~~~~
+Client Windows implementation
```
dlech, line 34: It would be nice to move the backend docs to a separate page.

dlech (commented):
This looks like a good start. Just a few high level comments.

jackjansen:
@dlech can you have a look? All relevant classes are now exported from toplevel. The documentation has be reorganised to have the low-level stuff and the base classes in a separate page. I've also added some prose to help people along, _but_: that's the prose that would have helped _me_ along, it'd be good to have more pairs of eyes on it.

bleak/_api.py
jackjansen: I haven't moved the documentation here, because frankly I'm still a bit out of my water with the whole ABC and typing stuff, and how I could document a method in the toplevel class (in an abstract method?) and then override the implementation from the subclass.

docs/api.rst
jackjansen: Done.

bleak/__init__.py
```diff
@@ -8,13 +8,39 @@
...
+    "BleakScanner",
+    "BleakClient",
+    "BLEDevice",
+    "BleakGATTServiceCollection",
+    "BleakGATTService",
+    "BleakGATTCharacteristic",
+    "BleakGATTDescriptor",
+    "discover",
+    "cli",
+]
```
dlech, line 43: I think we also need `AdvertisementData`, `AdvertisementDataCallback` and `AdvertisementDataFilter` (from `backends.scanner`) here.

bleak/__init__.py
```diff
@@ -8,13 +8,39 @@
...
+    "BleakError",
+    "BleakScanner",
+    "BleakClient",
+    "BLEDevice",
+    "BleakGATTServiceCollection",
+    "BleakGATTService",
+    "BleakGATTCharacteristic",
+    "BleakGATTDescriptor",
+    "discover",
+    "cli",
```
dlech, line 42: `cli` has never really been developed or used AFAIK, so could probably be dropped.

docs/apiimplementations.rst
```diff
@@ -0,0 +1,88 @@
+API Implementations
+===================
+
+Bleak hides most of the details of the Bluetooth LE stack provided by the operating system,
+but sometimes it may be useful to access those details because it provides useful functionality 
+(even though your code may no longer be portable to other operating systems).
```
dlech, line 6: We should also mention that the backends do not have a stable API.

docs/index.rst
```diff
@@ -53,6 +53,7 @@ Contents:
    usage
    backends/index
    api
+   apiimplementations
```
dlech, line 56: I think `backends` would be a better name (also, I have a hard time reading words squashed together like this - my brain likes to change the letters when I see words I don't recognize).

bleak/__init__.py
```diff
@@ -8,13 +8,39 @@
...
-import platform
 import asyncio
 
 from bleak.__version__ import __version__  # noqa: F401
 from bleak.exc import BleakError
-
-_on_rtd = os.environ.get("READTHEDOCS") == "True"
+from bleak._api import (
+    BleakScanner,
+    _BleakScannerImplementation,
```
dlech, line 17: To be consistent with the existing naming conventions, `_BleakXyzBackend` instead of `_BleakXyzImplementation` seems like the more obvious name.

bleak/_api.py
```diff
@@ -0,0 +1,237 @@
+import os
+import platform
+from bleak.exc import BleakError
+from bleak.backends.scanner import BaseBleakScanner
+from bleak.backends.client import BaseBleakClient
+
+
+_on_rtd = os.environ.get("READTHEDOCS") == "True"
```
dlech, line 8: If this is only used once, we probably just inline it instead of using a variable.

bleak/_api.py
```diff
@@ -0,0 +1,237 @@
...
+
+if _on_rtd:
+    from bleak.backends.scanner import (
+        BaseBleakScanner as _BleakScannerImplementation,
```
dlech, line 12: To make the docs correct, all of the doc strings of the abstract base classes would need to be updated to document the concrete implementation rather than the abstract base class. Is this your intention?

bleak/_api.py
```diff
@@ -0,0 +1,237 @@
...
+else:
+    raise BleakError(f"Unsupported platform: {platform.system()}")
+
+# Now let's tie together the abstract class and the backend implementation
+class BleakScanner(_BleakScannerImplementation):
+    """
+    Interface for Bleak Bluetooth LE Scanners.
+
+    The actual implementation is dependent on the backend used, and some methods (notably the
+    constructor) may have additional optional arguments.
```
dlech, line 134: For me, one of the main reasons for making this change is to be able to have all of the OS-specific quirks/args listed in one place rather than having read the same docs 5 times for each platform. This makes it seem like we are not accomplishing that goal.

docs/api.rst
```diff
@@ -1,93 +1,87 @@
-Interfaces, exceptions and utils
-================================
+API
+===
 
-Connection Clients
-------------------
+The naming of things can be confusing when using Bluetooth LE, due to the many possible meanings of
+words like "device" or "service" depending on the context. Here is a quick breakdown of the main object
+types used in bleak and their meaning, roughly in the order in which you will often use them in your code:
```
dlech, line 6: This is a really nice way to introduce the API.

docs/api.rst
```diff
@@ -1,93 +1,87 @@
...
-~~~~~~~~~~~~~~~~~~~~~~~
+Client interface
+~~~~~~~~~~~~~~~~
 
-.. automodule:: bleak.backends.p4android.scanner
+.. autoclass:: bleak.BleakClient
     :members:
+    :inherited-members:
 
+Addressing items in the BLE device
```
dlech, line 60: `BleakGATTServiceCollection` and friends are more about enumerating rather than addressing or accessing. For devices with only one instance of each service/characteristics, using theses classes is not necessary unless you want to print meta info about the services/characteristics.

docs/index.rst
dlech: In fact, it looks like there are already `backends/index` with individual pages for each backend.

bleak/__init__.py
dlech: And perhaps other types reference via type hints of the top-level APIs?

bleak/__init__.py
```diff
@@ -8,13 +8,39 @@
...
+    BLEDevice,
+    _BLEDeviceImplementation,
+    BleakGATTServiceCollection,
+    BleakGATTService,
+    _BleakGATTServiceImplementation,
+    BleakGATTCharacteristic,
+    _BleakGATTCharacteristicImplementation,
+    BleakGATTDescriptor,
+    _BleakGATTDescriptorImplementation,
+    discover,
```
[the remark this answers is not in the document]
jackjansen: will do.

bleak/__init__.py
jackjansen: Will do for the `backends.scanner` objects. But I'm not sure I understand what you mean by 

> And perhaps other types reference via type hints of the top-level APIs?

can you explain?

jackjansen: Hmm, I'm also getting a funny signature for `AdvertisementDataCallback`, and `AdvertisementDataFilter`, and they aren't the same type as what is used in the `BleakScanner` signatures... Any suggestion why that could be the case?

bleak/__init__.py
jackjansen: ok.

docs/apiimplementations.rst
jackjansen: ok.

docs/index.rst
jackjansen: Yeah, but the current `backends/index` (and the stuff below it) is really a completely different beast. A lot of it is implementation or usage notes.

Should I try to merge the two sets of pages? Or keep them separate and rename (Backend APIs and Backend Notes)? Or is there a middle ground?

bleak/_api.py
jackjansen: I was planning to burn that bridge when we get there, but apparently we have:-)

We really need two abstract base classes: one to specify how things look for the user and another one to specify how things look for a backend implementor. So the former wouldn't include the methods that the end user isn't supposed to call, for example. But I've experimented a bit with it and I keep getting into trouble with MRO...

Suggestions?

bleak/_api.py
jackjansen: Yes, this needs fixing.

docs/api.rst
jackjansen: ok.

bleak/__init__.py
jackjansen: Agreed. Will do.

bleak/_api.py
jackjansen: Okay, I think I have something that works. I'm going to create another module `abstract_api.py` that has the user-visible abstract classes (and their user-visible documentation). the abstract backend classes are going to be subclassed from these (and include the methods and documentation for backend developers). When `_on_rtd` we inherit from the `abstract_api` classes.

But the backend documentation section will still include the "hidden" stuff.

Now only the constructors for `BleakClient` and `BleakScanner` need to be documented in a funny place.

bleak/_api.py
jackjansen: I think this is resolved with the new `abstract_api.py`.

bleak/__init__.py
jackjansen: I need help with these ones. I _think_ `TypeAlias` might help, or `NewType`, but I'm not familiar with them at all, and I also think they're not backward compatible to 3.7.

bleak/_api.py
jackjansen: It is now used twice:-)

bleak/_api.py
jackjansen: Implemented and pushed. Please check.

bleak/_api.py
jackjansen: I think this is fixed with the `abstract_api` changes.

docs/api.rst
jackjansen: I've added a few notes on attributes and handles too.

bleak/_api.py
dlech: > Suggestions?

I suppose a bit too late now :smile: but my first thought on this was to do something more like dependency injection instead of multiple inheritance. For example, in `BleakClient` might look something like this:

```python
from bleak.backends.client import get_platform_backend

...

class BleakClient:
    def __init__(self, ...):
        self._backend = get_platform_backend()
        ...

    async def connect(self):
        # common code like: if already connected, raise exception
        ...
        await self._backend.connect()
```

dlech: > When _on_rtd we inherit from the abstract_api classes.

This works for the oneline docs, but what about `help()` in Python and docs for IDE intellesense? Wouldn't they still be using the doc string from one of the backends? So we would still need to write all of the docs strings 5 times?

docs/index.rst
dlech: I think it would be OK to merge them.

docs/api.rst
dlech: The purpose of handles in Bleak is to differentiate between two attributes with the same UUID. More info in #888.

bleak/__init__.py
dlech: > can you explain?

I didn't go through and check every method to see if there are other custom types like these. We will need to do this eventually to make sure we didn't miss anything.

dlech: > Hmm, I'm also getting a funny signature

Do you mean in the docs generated with Sphinx? We are using such and old version of Sphinx, we really should update.

bleak/_api.py
jackjansen: > my first thought on this was to do something more like dependency injection instead of multiple inheritance.

This would work only for `BleakClient` and `BleakScanner`, right? But there is definitely something to be said for it. 

It would also provide a nice hook for something that I need: using bleak in a framework that is not `asyncio`-based. I have something but it's hand-coded and not ready for prime time.

My suggestion is to keep this in mind as a separate issue. But it's your call...

jackjansen: > what about help() in Python and docs for IDE intellesense?

I've completely ignored them:-)

Will check, and I think the writing the doctoring 5 times may only be needed for some very select methods, I think most can just use the doctoring of the abstract base class.

I'll check, but I will need help if things need fixing (don't know anything about the Android and Windows BLE stacks)

dlech: > This would work only for BleakClient and BleakScanner, right?

Possibly. I haven't really looked at the other classes. But I think we should eventually be able to make most, if not all, of the other classes not backend-specific, so we could just ignore those for now.

We don't need to refactor the common code from the various backends just yet, but would it be too much work to just make the new `BleakClient` and `BleakScanner` classes a thin wrapper around each backend at this point in time (each method is implemented by `return self._backend.method()`) ?

docs/api.rst
jackjansen: Ok, rephrased.

jackjansen:
I think I'm getting close. Please check.

Also, it would be good to test that the code still runs on all platforms. I can only test on macOS. And I've had to do some restructuring to make things documentable in a sane way (for example adding `BLEDevice` subclasses to all backends that didn't have them yet, even though they are empty).

And, who knows, maybe when I removed the docstrings I've removed a line or two of useful code too...

bleak/_api.py
```diff
@@ -0,0 +1,214 @@
...
+            containing this advertising data will be received.
+    :type service_uuids: Optional[List[str]]
+    :param scanning_mode:
+            Set to "passive" to avoid the "active" scanning mode.
+    :type scanning_mode: Literal['active', 'passive']
+    """
+
+    # Gross hack to get the best docstring:
+    if len(__doc__) < len(_BleakScannerImplementation.__doc__):
+        __doc__ = _BleakScannerImplementation.__doc__
```
dlech, line 153: Yes, this doesn't seem like something we want.

bleak/backends/bluezdbus/device.py
```diff
@@ -0,0 +1,12 @@
+from bleak.backends.device import BLEDevice
+
+
+class BLEDeviceBlueZDBus(BLEDevice):
```
dlech, line 4: Ooof. This seems like going in the wrong direction. I know it is not the case right now, but ideally `BleakScanner` and `BleakClient` should be the only classes that have a platform-specific implementation. With some effort, `BleDevice` and the service/characteristic/descriptor classes could be made universal. We don't have to do that now, but let's not make more work just to be undone later.

bleak/backends/bluezdbus/scanner.py
```diff
@@ -45,26 +46,21 @@ class BlueZScannerArgs(TypedDict, total=False):
...
-            Optional list of service UUIDs to filter on. Only advertisements
-            containing this advertising data will be received. Specifying this
-            also enables scanning while the screen is off on Android.
-        scanning_mode:
-            Set to ``"passive"`` to avoid the ``"active"`` scanning mode.
-        **bluez:
-            Dictionary of arguments specific to the BlueZ backend.
-        **adapter (str):
-            Bluetooth adapter to use for discovery.
+    :param detection_callback: Optional function that will be called each time a device is discovered or advertising data has changed.
```
dlech, line 57: Let's not change the arg style and keep using the [google doc style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html).

If we have type hints in the actual method declarations, we should not need to repeat them in the doc strings.

dlech (commented):
There's lots of great stuff here. But its starting to be too much all at once for me to review.

It would be nice if we could break this down into some smaller chunks. For example, the first thing we should probably do is update to a recent version of Sphinx and make sure all of that is working. Then we could pick just one class, e.g. `BleakScanner`, and focus on that until we get it right. Then the rest of it should follow easier.

bleak/_api.py
jackjansen: Fully agreed. But we somehow need to distinguish between two cases, and this was the quick-and-dirty way I came up with:

- If the base class is incomplete (i.e. we're building for RTD) we should provide the docstring for the constructor,
- if the base class is complete (i.e. we're at runtime, and we need to provide help for introspection tools and `help()` it's better to provide the base class docstring because it's better.

Suggestions?

bleak/backends/bluezdbus/device.py
jackjansen: Oops, sorry. And fully agreed. I wanted to make things more orthogonal between backend implementations, but your suggestion is much better. 

OTOH: I'm not sure I've introduced "more work": now all classes are treated the same way for all implementations. So that basically means that all classes could get a `backend_private` property where the backend could store whatever bits they wanted and if they wanted.

Hmm, I see an analogy to your idea in the other thread about making `BleakClient` backend-independent...

bleak/backends/bluezdbus/scanner.py
jackjansen: Grmpf, I wasn't aware of that style guide (and apparently there were enough things documented in a way that didn't adhere to it that I didn't notice there was a style guide in effect), so I just picked RTD style because I'm more familiar with that. I'll revert.

bleak/_api.py
jackjansen: Fixed, by simply adding a mock base class for RTD. Also has the advantage that we have a constructor signature now.

bleak/backends/bluezdbus/scanner.py
jackjansen: Reverted back to google style.

jackjansen:
Next try.

It's getting increasingly more bothersome to merge develop into this branch, mainly because I've had to move some backend base class implementation methods into abstract classes (the stuff related to the async context manager). So now when there's changes in those classes on develop I have to handcraft those.

jackjansen:
@dlech I want to abandon this merge request, #982 is really all I need.

But that means we would lose any documentation fixes you hint at maybe wanting to pick up.

What do you suggest?

dlech:
You mentioned before that some of the stuff you wrote were things you would have like to have known when you first started using Bleak. So if you could copy and paste that into a new pull request, I'm sure other newbies would appreciate having that in the docs.

dlech:
Closing since this has been broken down into smaller pull requests and issues.

Issue #798, opened 2022-04-06
Don't use bluetoothctl


kq98:
* bleak version: 0.14.2
* Python version: 3
* Operating System: GNU/Linux
* BlueZ version: 5.60

### Description

Trying to use bleak inside a Flatpak sandbox fails because bluetoothctl, like all other system utils, can't be accessed.

Inside the bluezdbus backend in `__init__.py` a bluetoothctl subprocess is spawned to get the current bluez version. 
I suggest to use the bluez dbus interface directly to get the version.

hbldh:
I am pretty sure there is no such thing available in the dbus api for BlueZ. At least there wasn't when Bleak was written.

It is regrettable that bluetoothctl is needed there, and I would like to remove it if possible.

Simplest solution for you would be to modify the init file to have your version hard coded, for this very specific use case.

dlech:
This was fixed by #907 by not crashing if `bluetoothctl` is not available.

PR #907, opened 2022-07-28
Check the bluetoothctl version with asyncio.create_subprocess_exec

bdraco:
Partial fix for #798


dlech:
> Partial fix for #798

This change still uses `bluetoothctl` so I don't think that issue is relevant.

bleak/backends/bluezdbus/client.py
```diff
@@ -30,11 +30,14 @@
 from bleak.backends.device import BLEDevice
 from bleak.backends.service import BleakGATTServiceCollection
 from bleak.exc import BleakDBusError, BleakError
-
+from .version import BlueZFeatures
 
 logger = logging.getLogger(__name__)
 
 
+_on_ci = "CI" in os.environ
```
dlech, line 38: This should no longer be needed. It was just a hack to prevent import fail so we didn't have to install bluez on CI runs.

dlech (changes requested):
Looks good. Just one suggestion for change.

bleak/backends/bluezdbus/client.py
bdraco: Removed, re-squashed, and re-tested.  all good 👍

bleak/__init__.py
```diff
@@ -12,7 +12,6 @@
 import asyncio
 
 from bleak.__version__ import __version__  # noqa: F401
-from bleak.backends.bluezdbus import check_bluez_version
 from bleak.exc import BleakError
 
 _on_rtd = os.environ.get("READTHEDOCS") == "True"
```
dlech, line 17: We can remove the `_on_ci = "CI" in os.environ` line below this line too since it is no longer used.

bdraco: Re-squashed and pushed with this removed

dlech:
Thanks!

PR #912, opened 2022-08-01
Make callback logging lazy

bdraco:
Fixes logging-format-interpolation (W1202)

Fixes `__repr__` being called and thrown away
Screen Shot 2022-07-31 at 4 06 25 PM

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,11 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 `Unreleased`_
 =============
 
+Fixed
+-----
+* Made BlueZ D-Bus signal callback logging lazy to improve performance.
+
+
 `0.15.0`_ (2022-07-29)
 ======================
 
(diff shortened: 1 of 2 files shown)


dlech (changes requested):
Looks good. Can you please add a changelog entry? Also it would be nice to change the commit message to explain this only changes one instance in the bluez backend (there are many other callbacks that could use the same fix).

bdraco:
Will fix it when I get back home, looks like I forgot to push or commit since I remember writing it 🤦‍♂️

bdraco:
Sorry about that. I had pushed it to the wrong remote and didn't notice it rejected

CHANGELOG.rst
```diff
@@ -10,6 +10,11 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 `Unreleased`_
 =============
 
+Fixed
+-----
+* Made BlueZ callback logging lazy to improve performance.
```
dlech, line 15: It would be more accurate to say `BlueZ D-Bus signal callback` since we haven't done this in general for all callbacks.

bdraco: Squashed and repushed

dlech (approved):
Thanks!

PR #916, opened 2022-08-02
Switch to using async_timeout for timeouts

bdraco:


bleak/backends/corebluetooth/PeripheralDelegate.py
```diff
@@ -10,6 +10,7 @@
 import itertools
 import logging
 from typing import Callable, Any, Dict, Iterable, NewType, Optional
+import async_timeout
```
dlech, line 13: Since this isn't a builtin module, it would be nice to group it with the other 3rd party modules below.

bleak/backends/p4android/scanner.py
```diff
@@ -7,6 +7,8 @@
 
 from typing_extensions import Literal
 
+import async_timeout
```
dlech, line 10: This can be grouped with the `typing_extensions` import.

setup.py
```diff
@@ -19,6 +19,7 @@
 AUTHOR = "Henrik Blidh"
 
 REQUIRED = [
+    "async_timeout>=4.0.1",
```
dlech, line 22: There is also a pipenv file with yet another copy of the requirements.

dlech (commented):
Can you share some before/after measurements so we can see what the actual performance improvement is?

bleak/backends/corebluetooth/PeripheralDelegate.py
bdraco: Adjusted in dfae35b8cba0908e36d1bdd2e3bae9ea21b04731

bleak/backends/p4android/scanner.py
bdraco: Adjusted in dfae35b8cba0908e36d1bdd2e3bae9ea21b04731

setup.py
bdraco: I added the changes to `Pipfile`, but I didn't find any other place.  Is there another file?

bdraco:
> Can you share some before/after measurements so we can see what the actual performance improvement is?

Its best summed up by this stackoverflow by one of the asyncio authors

https://stackoverflow.com/questions/47000770/async-timeout-timeout-vs-asyncio-wait-for

dlech:
merged, thanks!

PR #919, opened 2022-08-02
Handle the race in the BlueZ D-Bus backend where the device disconnects during the connection process

bdraco:
Fixes
```
Logger: bleak.backends.bluezdbus.client
Source: components/switchbot/cover.py:94
First occurred: 10:36:23 (1 occurrences)
Last logged: 10:36:23
Failed to cancel connection (/org/bluez/hci0/dev_XX_XX_XX_XX_XX_XX): 'NoneType' object has no attribute 'call'
```


dlech:
thanks!

Issue #882, opened 2022-07-13
Bluez GATT attribute stale caching issues


projectgus:
* bleak version: 0.14.3
* Python version: 3.9.13
* Operating System: Arch Linux amd64 (with a downgraded Python version)
* BlueZ version (`bluetoothctl -v`) in case of Linux: 5.65

### Description

I ran into some unexpected behaviour with BLE attribute caching and Bluez. As mentioned in other issues, there are definitely some Bluez bugs here (that I intend to report to them as well). However, I think that in some of the cases Bleak can be made to correctly handle the cache invalidation.

The goal is to be able to update the firmware for a paired/trusted device in a way that changes GATT attributes and have this reflected by Bleak. It should always work when using the Service Changed attribute correctly.  As a bonus it may sometimes work when not using the Service Changed attribute correctly(!)

This is very similar to discussion in #625, #628 and probably some other issues. I couldn't find an open issue where it fit so I've started a new one - apologies if there was a better place I should have used.

### What I Did

Bleak code is this:

```python
import logging
from bleak import BleakClient

logging.basicConfig(encoding='utf-8', level=logging.DEBUG)
logger = logging.getLogger(__file__)

ADDRESS = 'F4:12:FA:41:8D:B2'

async def connect_and_pair():
    c = BleakClient(ADDRESS)
    await c.connect()
    await c.pair()
    services = await c.get_services()
    for char in services.characteristics.values():
        logger.info(f'SVC {char.service_uuid} CHAR {char}')

asyncio.run(connect_and_pair())
```

The device is a modified ESP-IDF NimBLE `bleprph` example running on an ESP32-S3, which exposes the standard Service Changed characteristic and a custom GATT service with a single custom GATT attribute. I don't think the details of the device are relevant, though.

## 01 - Initial run

Works as expected, Bluez does service discovery and caches the services. 

```
INFO:SVC 00001801-0000-1000-8000-00805f9b34fb CHAR 00002a05-0000-1000-8000-00805f9b34fb (Handle: 7): Service Changed
INFO:SVC 22222222-2222-2222-2222-222222222222 CHAR bbbbbbbb-bbbb-bbbb-bbbb-bbbbbbbbbbbb (Handle: 11): Unknown
```

See [01-initial.log](https://github.com/hbldh/bleak/files/9099585/01-initial.log) for output from Bleak, contents of cache after running, and debug output from bluetoothd.

## 02 - Change characteristic UUID

[...]

PR #923, opened 2022-08-06
Cache service resolution on connect when using BlueZ/Dbus backend when requested

bdraco:
If the device was previously connected and we have the
the service collection already built, it can
be reused for the next connection if the device has not
since been removed from the bus when the dangerous_use_bleak_cache
flag is passed.

For devices that connect and disconnect frequently this
can make operations up to an magnitude faster once
the first connection has been made and the device
has not been removed from the bus since the last
connection.

In testing this significantly reduced the time it took to execute operations on Ble devices. The goal is to be able to quickly operate Bluetooth locks (and other similar devices)


dlech:
This will not work with devices that suffer from the issues described in #882. We need to wait for "ServicesResolved" after connect to ensure that BlueZ is not longer adding or removing service/characteristic/descriptor d-bus objects.

I suppose we could make this opt-in for devices that are known to be well behaved (meaning they don't trigger the BlueZ bugs).

bleak/backends/bluezdbus/client.py
```diff
@@ -469,9 +476,14 @@ def mtu_size(self) -> int:
...
     # GATT services methods
 
-    async def get_services(self, **kwargs) -> BleakGATTServiceCollection:
+    async def get_services(
+        self, use_cached: bool = False, **kwargs
+    ) -> BleakGATTServiceCollection:
         """Get all services registered for this GATT server.
 
+        Args:
+            use_cached (bool): Use cached services if available.
```
dlech, line 485: This should probably be called something like `dangerous_use_bleak_cache`. Dangerous since is subject to the issues described in #882 and `bleak` because the caching is done by bleak (we use `use_cached` elsewhere to mean us something cached by the OS bluetooth stack).

bdraco:
I reworked it to be opt-in and renamed it to be `dangerous_use_bleak_cache`

CHANGELOG.rst
```diff
@@ -41,6 +44,7 @@ Fixed
 
 Changed
 -------
+* Switch to using async_timeout instead of asyncio.wait_for for performance.
```
bdraco, line 47:
```suggestion
```

bleak/backends/bluezdbus/manager.py
```diff
@@ -508,7 +511,9 @@ def remove_device_watcher(self, watcher: DeviceWatcher) -> None:
         """
         self._device_watchers.remove(watcher)
 
-    async def get_services(self, device_path: str) -> BleakGATTServiceCollection:
+    async def get_services(
+        self, device_path: str, use_cached: bool
```
dlech, line 515: Can we add `use_cached` to the Args in the doc comment?

dlech: It took care of it.

bdraco:
Thanks

PR #928, opened 2022-08-08
Adjust default timeout for ``read_gatt_char()`` with CoreBluetooth to 20s

### LINKED ISSUE #926: corebluetooth read_characteristic timeout is too low
Seen in https://github.com/Jc2k/aiohomekit/issues/136

Devices with a low MTU that talks HAP over bluetooth will fail to pair because the timeout is too lower

e1c4ced242bf2f054948b7baf367465dc9a16347 raised the timeout, but unfortunately it wasn't high enough.

I'm going to do some experiments to see if 15 is enough or if 20 is needed.  I suspect we are just close enough now.

BlueZ does not have the issue

bdraco:
fixes #926

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,6 +19,7 @@ Fixed
 * Handle the race in the BlueZ D-Bus backend where the device disconnects during
   the connection process which presented as ``Failed to cancel connection``. Merged #919.
 * Ensure the BlueZ D-Bus scanner can reconnect after DBus disconnection. Merged #920.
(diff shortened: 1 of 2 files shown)


dlech:
I wonder if we should just remove this timeout altogether. The other backends don't have a timeout on this method.

bdraco:
That's a good thought.  I'll do some testing to make sure I can't get it to hang

bdraco:
I took out the timeout and attempt to do a homekit pairing.  After a few times I got it to fail, and its hanging forever without the timeout so I think we do need one, its just a question of how low is too low.

PR #930, opened 2022-08-09
Implement max_write_without_response_size for BleakGATTCharacteristic

### LINKED ISSUE #738: Introduce per-characteristic `max_write_without_response_size`
A while back, we added a `BleakClient.mtu_size` property. This can be used to determine what is the largest size of data you can send using write without response which is `mtu_size - 3`.

There are a couple problems with this. One is that you have to know what MTU means and that you have to subtract 3 from it. The other is that the BlueZ backend doesn't currently doesn't have a proper way to get the negotiated MTU and defaults to the minimum value of 23.

However, BlueZ recently introduced a new [MTU property on characteristics](https://github.com/bluez/bluez/blob/7a8d1be9755f39ca2d3060b4ba9f9ce58c83c0ed/doc/gatt-api.txt#L297-L301). BlueZ's reason for doing this is that with Bluetooth 5 there are enhanced attributes that allow different MTU sizes per characteristic.

So, I would like to propose adding a new `max_write_without_response_size` property on `BleakGATTCharacteristic` objects that returns `mtu_size - 3`, or in the case of BlueZ, the "MTU" d-bus property minus 3. This makes it obvious what the value should be used for and will hopefully be future-proof if other OSes add support for enhanced attributes.

### LINKED ISSUE #925: BlueZ client._acquire_mtu() raises RuntimeError on many devices
Most test devices I have result in a `RuntimeError` from `StopIteration` because there are no write without response chars on the  device.  I'm able to get the MTU of the device with this proof of concept

```python3
            from bleak.backends.bluezdbus.manager import get_global_bluez_manager
            logger.debug("%s: Acquiring MTU from BlueZ",name)
            await client.get_services()

            device_path = client._device_path
            mtu: int | None = None
            for service_path, service_ifaces in manager._properties.items():
                if (
                    not service_path.startswith(device_path)
                    or defs.GATT_CHARACTERISTIC_INTERFACE not in service_ifaces
                ):
                    continue
                if "MTU" in service_ifaces[defs.GATT_CHARACTERISTIC_INTERFACE]:
                    mtu = service_ifaces[defs.GATT_CHARACTERISTIC_INTERFACE]["MTU"]
                    break
            if not mtu:
                try:
                    await client._acquire_mtu()
                    mtu = client._mtu_size
                except RuntimeError as ex:
                    logger.debug("%s: Failed to acquire MTU via _acquire_mtu: %s",name, ex)
                    pass
            if mtu:
                KNOWN_MTUS[device.address] = mtu
                client.set_known_mtu(mtu)
                logger.debug("%s: Acquired MTU from BlueZ %s", name, mtu)
            else:
                logger.debug("%s: No MTU from BlueZ", name)
```

bdraco:
Closes #738
Closes #925

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,12 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 `Unreleased`_
 =============
 
+Added
+-----
+* BleakGATTCharacteristic now have a ``max_write_without_response_size`` property which is
+  the maximum size of a write without response that can be sent to the remote device. This
+  value is the MTU for the characteristic - 3 bytes for the ATT header.
+
 Fixed
 -----
 * Made BlueZ D-Bus signal callback logging lazy to improve performance.
--- a/bleak/backends/bluezdbus/characteristic.py
+++ b/bleak/backends/bluezdbus/characteristic.py
@@ -1,10 +1,13 @@
 from uuid import UUID
-from typing import Union, List
+from typing import Union, List, Optional
 
 from bleak.backends.bluezdbus.utils import extract_service_handle_from_path
-from bleak.backends.characteristic import BleakGATTCharacteristic
+from bleak.backends.characteristic import (
+    BleakGATTCharacteristic,
+    DEFAULT_MTU_SIZE,
+    ATT_HEADER_SIZE,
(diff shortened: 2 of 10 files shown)


bdraco:
closing in favor of #962

PR #945, opened 2022-08-15
backends\winrt set data to bytes

Artucuno:
Fixing an issue experienced on Windows.

`a bytes-like object is required, not 'list'`

Found using this repo: https://github.com/al5681/Sphero_Bolt_Multiplatform_Python_Bleak/pull/3


dlech:
The `data` argument comes from the caller, so I don't think it is Bleak's responsibility to convert it to bytes.

The bug should be fixed at https://github.com/al5681/Sphero_Bolt_Multiplatform_Python_Bleak/blob/afd6c12a4be26dfdace4c0ebaac182505e519cfc/sphero_bolt.py#L128

Artucuno:
Ah alright, thankyou

PR #962, opened 2022-08-26
Max write without response size

dlech:
This is an alternative to #930. I'm hoping that we can eventually drop the platform-specific implementations of `BleakGattCharacteristic`, so this implements it in a way that doesn't require overriding the property for each backend.

cc: @bdraco


bdraco (approved):
LGTM.  The functional difference I see is it doesn't preserve the value of `_acquire_mtu` in the bluez backend, but I couldn't actually use that function in production anyways since it wasn't workable on many devices anyways. Long term it can hopefully go away since BlueZ is now sending over the MTU anyways

dlech:
For older versions of BlueZ it will also be possible to monkey patch the value by setting `_max_write_without_response_size` either using `_acuire_mtu` or a hard-coded value, but as you pointed out, this needs to be handled per-device.

PR #967, opened 2022-08-29
BlueZ: fix removed devices included in discovered_devices

### LINKED ISSUE #942: Devices are no longer removed from `.discovered_devices` on the bluez scanner when InterfacesRemoved is fired
In 0.14.3, the devices would get removed from `self._devices` so when `BleakScannerBlueZDBus.discovered_devices` was enumerated we knew they were now gone.

Since the switch to the global bluez manager in 0.15, they now stay in `BleakScannerBlueZDBus.discovered_devices` forever since the scanner never gets a removal callback.

dlech:
This fixes a regression introduced in v0.15 where device removed from BlueZ while scanning were no longer removed from this list of discovered devices. This caused the `discovered_devices` property to return devices that could no longer be connected to by BlueZ since they were removed.

Fixes #942.

### DIFF
--- a/bleak/backends/bluezdbus/scanner.py
+++ b/bleak/backends/bluezdbus/scanner.py
@@ -132,11 +132,17 @@ class BleakScannerBlueZDBus(BaseBleakScanner):
 
         if self._scanning_mode == "passive":
             self._stop = await manager.passive_scan(
-                self._adapter_path, self._or_patterns, self._handle_advertising_data
+                self._adapter_path,
+                self._or_patterns,
+                self._handle_advertising_data,
+                self._handle_device_removed,
             )
         else:
             self._stop = await manager.active_scan(
-                self._adapter_path, self._filters, self._handle_advertising_data
+                self._adapter_path,
+                self._filters,
+                self._handle_advertising_data,
+                self._handle_device_removed,
             )
 
     async def stop(self):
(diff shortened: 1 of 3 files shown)


dlech:
@bdraco can you please test?

bleak/backends/bluezdbus/scanner.py
```diff
@@ -245,3 +245,15 @@ def _handle_advertising_data(self, path: str, props: Device1) -> None:
         )
 
         self._callback(device, advertisement_data)
+
+    def _handle_device_removed(self, device_path: str) -> None:
+        """
+        Handles a device being removed from BlueZ.
+        """
+        try:
+            self._devices[device_path]
```
bdraco, line 254:
```suggestion
            del self._devices[device_path]
```

Should this be `del`?

bdraco:
Been testing this for a few hours with the above `del` change and everything is working great.

Devices are going unavailable when they disappear now.

bdraco:
To make sure it is effective I did another test by setting up a humidity sensor.  

You can see from the graph when I took the batteries out as the device goes unavailable a few minutes later.  The holes in the graph are from the latest test where I removed the batteries and later replaced them each time.
Screen Shot 2022-08-29 at 8 44 47 PM

bdraco (approved):
Tested ok with the suggested change in https://github.com/hbldh/bleak/pull/967#discussion_r957606072

bdraco:
Screen Shot 2022-08-29 at 11 21 21 PM

Tested with an extended battery removal as well.  Recovered quickly 

This all looks good with the `del` change

bleak/backends/bluezdbus/scanner.py
dlech: yes, of course!

dlech:
Thanks for testing!

PR #968, opened 2022-08-29
BlueZ: fix not disconnecting if get_services() throws during connect

### LINKED ISSUE #951: Device remains connected if connection is successful but get_services receives TimeoutError during connect
```
2022-08-20 11:12:01.209 DEBUG (MainThread) [yalexs_ble.lock] Front Door (M10CK0F): Disconnected from lock callback
2022-08-20 11:12:01.210 DEBUG (MainThread) [yalexs_ble.lock] Front Door (M10CK0F): Disconnected from lock callback
2022-08-20 11:12:01.211 DEBUG (MainThread) [yalexs_ble.lock] Front Door (M10CK0F): Disconnected from lock callback
```

I noticed I was getting the callback multiple times and tracked it down to the `_remove_device_watcher` not being canceled when the connection fails

https://github.com/hbldh/bleak/blob/456cea854823ef13c55efcaa5e0fba380a055d8c/bleak/backends/bluezdbus/client.py#L152

dlech:
It is possible for get_services() to raise an exception, e.g. if external code cancels the task. If this happens, we need to be sure
that the device disconnects so the BlueZClient is in a consistent state (if the connect() method raises and exception, is is expected that the device will not be connected).

To fix this, we can just move the call to get_services() inside of the existing try block that will disconnect on any exception, then reraise the exception.

Fixes #951.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,7 @@ Added
 Fixed
 -----
 * Made BlueZ D-Bus signal callback logging lazy to improve performance.
+* Fixed possible bad connection state in BlueZ backend. Fixes #951.
 
 Changed
 -------
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -167,8 +167,19 @@ class BleakClientBlueZDBus(BaseBleakClient):
                 assert_reply(reply)
 
                 self._is_connected = True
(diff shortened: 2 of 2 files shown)


dlech:
@bdraco can you please test?

bdraco (approved):
I was not able to get the problem to reoccur with this change

PR #969, opened 2022-08-31
release v0.16.0

### LINKED ISSUE #738: Introduce per-characteristic `max_write_without_response_size`

### LINKED ISSUE #927: Linear searching is a performance bottleneck in `BlueZManager.get_services` with many devices
Properties are indexed well for `_wait_condition` but building the services can be expensive when the number of properties stored gets to a large number.  Its noticeable when there ~20 bluetooth devices that have their services resolved in the properties dict

### LINKED ISSUE #942: Devices are no longer removed from `.discovered_devices` on the bluez scanner when InterfacesRemoved is fired

### LINKED ISSUE #951: Device remains connected if connection is successful but get_services receives TimeoutError during connect

dlech:
Added
-----
* Added ``BleakGattCharacteristic.max_write_without_response_size`` property. Fixes #738.

Fixed
-----
* Fixed regression in v0.15 where devices removed from BlueZ while scanning
  were still listed in ``BleakScanner.discovered_devices``. Fixes #942.
* Fixed possible bad connection state in BlueZ backend. Fixes #951.

Changed
-------
* Made BlueZ D-Bus signal callback logging lazy to improve performance. Merged #912.
* Switch to using async_timeout instead of asyncio.wait_for for performance. Merged #916.
* Improved performance of ``BlueZManager.get_services()``. Fixes #927.

Removed
-------
* Removed explicit inheritance from object in class declarations. Merged #922.
* Removed first seen filter in ``BleakScanner`` detection callbacks on BlueZ backend. Merged #964.


dlech:
cc @bdraco

bdraco:
Thanks

PR #982, opened 2022-09-05
Concrete top level classes

dlech:
This is a partial alternative to some of the code changes in #904. It implements the suggestion of creating new `BleakScanner` and `BleakClient` objects. This solves some of the documentation complications compared to the approach in #904 of using multiple inheritance. Also it will allow us to refactor some duplicated code between the backends in the future, like checking if a device is still connected and wrapping coroutine callbacks.

It is not a complete alternative to #904 as there are some really nice documentation improvements there that I think we should still pick up. And there are some other classes that are not backend-specific that could be moved out of the `backends` module, but they can be handled one at a time and at a later date.

cc: @jackjansen 

Fixes: #582.


bleak/__init__.py
```diff
@@ -26,63 +47,518 @@
...
+            Additional args for backwards compatibility.
+    """
+
+    def __init__(
+        self,
+        detection_callback: Optional[AdvertisementDataCallback] = None,
+        service_uuids: Optional[List[str]] = None,
+        scanning_mode: Literal["active", "passive"] = "active",
+        *,
+        bluez: BlueZScannerArgs = {},
```
dlech, line 79: importing `BlueZScannerArgs` could cause import errors on other platforms and should probably just be `dict` here.

bleak/__init__.py
```diff
@@ -26,63 +47,518 @@
...
+            devices at the same time.
+    """
+
+    def __init__(
+        self,
+        device_or_address: Union[BLEDevice, str],
+        disconnected_callback: Optional[Callable[[BleakClient], None]] = None,
+        *,
+        timeout: float = 10.0,
+        winrt: WinRTClientArgs = {},
```
dlech, line 291: importing `WinRTClientArgs` is causing CI failures. we will just have to use `dict` as the type here.

bleak/backends/scanner.py
```diff
@@ -183,74 +159,29 @@ def discovered_devices(self) -> List[BLEDevice]:
...
-            A list of the devices that the scanner has discovered during the scanning.
 
-        """
-        warn(
-            "This method will be removed in a future version of Bleak. Use the `discovered_devices` property instead.",
-            FutureWarning,
-            stacklevel=2,
-        )
-        return self.discovered_devices
+def get_platform_scanner_backend_type() -> Type[BaseBleakScanner]:
```
bdraco, line 163: It would be nice if this was cached with https://docs.python.org/3/library/functools.html#functools.cache (probably needs `lru_cache(maxsize=None)` instead though since < 3.9 is supported

bleak/__init__.py
balloob: you can make it type checking only:

```
from __future__ import annotations

if TYPE_CHECKING:
    from .backends.bluezdbus.scanner import BlueZScannerArgs
```

dlech: Nice idea. That would fix runtime errors for sure. We'll see how it affects docs and static analysis tools.

bleak/backends/scanner.py
balloob: not if it's cached

bleak/__init__.py
```diff
@@ -26,63 +47,518 @@
...
+        **kwargs,
+    ):
+        PlatformBleakClient = get_platform_client_backend_type()
```
bdraco, line 294: It would be great if we could pass this in for the proxies so we don't have to override _backend

bleak/backends/scanner.py
dlech: This shaves of a few microseconds on my system.

```
python -m timeit -n 10000 -s 'from bleak import BleakScanner' 's = BleakScanner()' 
```

without caching:
```
10000 loops, best of 5: 14.6 usec per loop
```

with caching:
```
10000 loops, best of 5: 10 usec per loop
```

bdraco: Probably doesn't matter in that case 👍

dlech: ok, leaving out for now

bleak/__init__.py
dlech: I added a commit to handle this

bleak/__init__.py
dlech: it worked :+1:

bleak/__init__.py
bdraco: Thanks!

bleak/__init__.py
```diff
@@ -27,55 +50,517 @@
...
+        2.  Providing an address or UUID instead of a :class:`BLEDevice` causes
+            the :meth:`connect` method to implicitly call :meth:`BleakScanner.discover`.
+            This is known to cause problems when trying to connect to multiple
+            devices at the same time.
+    """
+
+    def __init__(
+        self,
+        device_or_address: Union[BLEDevice, str],
+        disconnected_callback: Optional[Callable[[BleakClient], None]] = None,
```
bdraco, line 301: Looks like this keyword arg snuck in before the `*`

dlech: no, this is intentional, the arg is now positional or keyword

PR #986, opened 2022-09-08
Fixed use of wrong enum in unpair function of WinRT backend

jochenjagers:
Hello,

The wrong enum for checking the unpair result was used which caused an error when unpairing an already unpaired device as the values for ALREADY_UNPAIRED and ALREADY_PAIRED are not the same.


dlech:
Nice one, thanks!

PR #987, opened 2022-09-08
Added tx_power in advertisement_data for Windows, Mac and Linux

facepolish:

   guidelines in that document.
3. The pull request should work for Python 3.7+ on the following platforms:
    - OS X / macOS >= 10.11

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,10 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 `Unreleased`_
 =============
 
+Added
+-----
+* ``AdvertisementData`` class now has an attribute ``tx_power``. Merged #987.
+
 Changed
 -------
 * ``BleakClient`` methods now raise ``BleakError`` if called when not connected in WinRT backend.
@@ -19,13 +23,16 @@ Changed
 * The BlueZ D-Bus backend implements a services cache between connections to significancy improve reconnect performance.
   To use the cache, call ``connect`` and ``get_services`` with the ``dangerous_use_bleak_cache``
   argument to avoid services being resolved again.
-* ``AdvertisementData`` class now has an attribute ``tx_power``. scanner.py in backends were changed to set tx_power
-  if available.
+* The BlueZ D-Bus backend now uses ``dbus-fast`` instead of ``dbus-next`` which significantly improves performance.
+* The BlueZ D-Bus backend will not avoid trying to connect to devices that are already connected. Fixes #992.
 
 Fixed
 -----
 * Fixed wrong error message for BlueZ "Operation failed with ATT error". Merged #975.
 * Fixed possible ``AttributeError`` when enabling notifications for battery service in BlueZ backend. Merged #976.
+* Fixed use of wrong enum in unpair function of WinRT backend.
+* Fixed inconsistent return types for ``properties`` and ``descriptors`` properties of ``BleakGATTCharacteristic``.
+* Handle device being removed before GetManagedObjects returns in BlueZ backend. Fixes #996.
 
 Removed
 -------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,10 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 
 `Unreleased`_
 =============
+Changed
+-------
+*  ``AdvertisementData`` class now has an attribute ``tx_power``
+*
 
 Changed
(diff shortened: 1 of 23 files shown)


bleak/backends/bluezdbus/scanner.py
```diff
@@ -231,13 +231,17 @@ def _handle_advertising_data(self, path: str, props: Device1) -> None:
         _service_data = {k: bytes(v) for k, v in props.get("ServiceData", {}).items()}
         _service_uuids = props.get("UUIDs", [])
 
+        # Get tx power data
+        tx_power = props.get("TxPower",None)
```
dlech, line 235: The default of `get()` is already `None`.
```suggestion
        tx_power = props.get("TxPower")
```

CHANGELOG.rst
```diff
@@ -19,6 +19,8 @@ Changed
 * The BlueZ D-Bus backend implements a services cache between connections to significancy improve reconnect performance.
   To use the cache, call ``connect`` and ``get_services`` with the ``dangerous_use_bleak_cache``
   argument to avoid services being resolved again.
+* ``AdvertisementData`` class now has an attribute ``tx_power``. scanner.py in backends of winrt, corebluetooth and 
+  bluezdbus were changed to set tx_power if available.
```
dlech, line 23: Can we get this for the Android backend too (even if you aren't able to test it)?

i.e. use `result.getTxPower()` in `_PythonScanCallback.onScanResult()`.

It also looks like we might want to convert `ScanResult.TX_POWER_NOT_PRESENT` to `None` to be consistent with other backends.

https://developer.android.com/reference/android/bluetooth/le/ScanResult#getTxPower()

bleak/backends/corebluetooth/scanner.py
```diff
@@ -110,12 +110,16 @@ def callback(p: CBPeripheral, a: Dict[str, Any], r: int) -> None:
                 cb_uuid_to_str(u) for u in a.get("kCBAdvDataServiceUUIDs", [])
             ]
 
+            # set tx_power data if available
+            tx_power = a.get("kCBAdvDataTxPowerLevel",None)
```
dlech, line 114:
```suggestion
            tx_power = a.get("kCBAdvDataTxPowerLevel")
```

bleak/backends/scanner.py
```diff
@@ -44,6 +45,9 @@ def __init__(self, **kwargs):
         # Tuple of platform specific data
         self.platform_data: Tuple = kwargs.get("platform_data", ())
 
+        # Tx Power data
+        self.tx_power: Optional[int] = kwargs.get("tx_power",None) # tx_power can't be 0. so 0 may be better default value?
```
dlech, line 49:
```suggestion
        self.tx_power: Optional[int] = kwargs.get("tx_power")
```

The value can be from -127 to 126, so 0 is allowed. I think `None` makes sense for "not available".

dlech (changes requested):
Thanks! I have made a few suggestions.

CHANGELOG.rst
facepolish: Yes. Actually now I have Android Studio but don't have test environment Android (will come in 2 weeks). So this time I made changes and test them later (if necessary). Thank you for your advice.

bleak/backends/p4android/scanner.py
```diff
@@ -276,13 +276,18 @@ def onScanResult(self, result):
             entry.getKey().toString(): bytes(entry.getValue())
             for entry in record.getServiceData().entrySet()
         }
+        if result.TX_POWER_NOT_PRESENT:
+            tx_power = None
+        else:
+            tx_power = result.getTxPower() # get Tx Power
```
dlech, line 282: Based on the android docs, I think this need to be:
```suggestion
        tx_power = result.getTxPower()

        # change "not present" value to None to match other backends
        if tx_power == defs.ScanResult.TX_POWER_NOT_PRESENT:
            tx_power = None
```

And we will need to add `ScanResult` to the `defs` module.

```python
ScanResult = autoclass("android.bluetooth.le.ScanResult")
```

dlech: We still need to add `ScanResult = autoclass("android.bluetooth.le.ScanResult")` to `defs.py`

facepolish: Thank you! Now I understand the structure of your code and usage of TX_POWER_NOT_PRESENT

dlech (approved):

dlech:
Merged, thanks!

PR #988, opened 2022-09-09
Switch from dbus_next to dbus_fast

bdraco:
Removes workaround since `.disconnect()` now closes the socket and fixes the leak.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,6 +19,7 @@ Changed
 * The BlueZ D-Bus backend implements a services cache between connections to significancy improve reconnect performance.
   To use the cache, call ``connect`` and ``get_services`` with the ``dangerous_use_bleak_cache``
   argument to avoid services being resolved again.
+* The BlueZ D-Bus backend now uses `dbus-fast` instead of `dbus-next` which significancy improves performance.
 
 Fixed
 -----
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,7 +19,7 @@ Changed
 * The BlueZ D-Bus backend implements a services cache between connections to significancy improve reconnect performance.
   To use the cache, call ``connect`` and ``get_services`` with the ``dangerous_use_bleak_cache``
   argument to avoid services being resolved again.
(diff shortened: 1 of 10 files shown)


bdraco:
I think this is ready do go.  I just have to retest every device I have which will take an hour or so

CHANGELOG.rst
```diff
@@ -19,6 +19,7 @@ Changed
 * The BlueZ D-Bus backend implements a services cache between connections to significancy improve reconnect performance.
   To use the cache, call ``connect`` and ``get_services`` with the ``dangerous_use_bleak_cache``
   argument to avoid services being resolved again.
+* The BlueZ D-Bus backend now uses `dbus-fast` instead of `dbus-next` which significantly improves performance.
```
dlech, line 22: This is a `.rst` file, so we need double backticks.

pyproject.toml
```diff
@@ -25,11 +25,11 @@ classifiers = [
 python = "^3.7"
 async-timeout = "^4.0.1"
 typing-extensions = { version = "^4.2.0", python = "<3.8" }
-dbus-next = { version = "^0.2.2", markers = "platform_system=='Linux'" }
 pyobjc-core = { version = "^8.5", markers = "platform_system=='Darwin'" }
 pyobjc-framework-CoreBluetooth = { version = "^8.5", markers = "platform_system=='Darwin'" }
 pyobjc-framework-libdispatch = { version = "^8.5", markers = "platform_system=='Darwin'" }
 bleak-winrt = { version = "^1.1.1", markers = "platform_system=='Windows'" }
+dbus-fast = {version = ">=1.4.0", markers = "platform_system == \"Linux\""}
```
dlech, line 32: Let's use caret notation for the version to avoid breaking changes. And we can use single quotes to avoid the escape characters.

dlech (commented):
Thanks for all of the work on this!

CHANGELOG.rst
bdraco: 4432e85ed88d2eda01dc63a95f759ececf22b955

pyproject.toml
bdraco: 4432e85ed88d2eda01dc63a95f759ececf22b955

PR #990, opened 2022-09-09
add support for Python 3.11

dlech:
In theory, older versions of Bleak should also work with Python 3.11 as long as any binary dependencies are available for that version.

This just makes the support official and updates the dependencies to ones that include binary wheels for Python 3.11.


dlech:
macOS support is blocked by https://github.com/ronaldoussoren/pyobjc/issues/489

PR #997, opened 2022-09-10
Handle device being removed before GetManagedObjects returns

### LINKED ISSUE #996: There is a race at startup when a device is removed before get managed objects finishes
```
  File "/usr/local/lib/python3.10/site-packages/dbus_fast/message_bus.py", line 736, in _process_message
    result = handler(msg)
  File "/usr/local/lib/python3.10/site-packages/bleak/backends/bluezdbus/manager.py", line 710, in _parse_msg
    del self._properties[obj_path][interface]
KeyError: '/org/bluez/hci0/dev_00_06_80_08_D2_95'
```

bdraco:
Fixes #996

```
  File "/usr/local/lib/python3.10/site-packages/dbus_fast/message_bus.py", line 736, in _process_message
    result = handler(msg)
  File "/usr/local/lib/python3.10/site-packages/bleak/backends/bluezdbus/manager.py", line 710, in _parse_msg
    del self._properties[obj_path][interface]
KeyError: '/org/bluez/hci0/dev_00_06_80_08_D2_95'
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -28,6 +28,7 @@ Fixed
 * Fixed possible ``AttributeError`` when enabling notifications for battery service in BlueZ backend. Merged #976.
 * Fixed use of wrong enum in unpair function of WinRT backend.
 * Fixed inconsistent return types for ``properties`` and ``descriptors`` properties of ``BleakGATTCharacteristic``.
+* Handle device being removed before GetManagedObjects returns in BlueZ backend. Fixes #996.
 
 Removed
 -------
--- a/bleak/backends/bluezdbus/manager.py
+++ b/bleak/backends/bluezdbus/manager.py
@@ -706,7 +706,10 @@ class BlueZManager:
             obj_path, interfaces = message.body
 
             for interface in interfaces:
-                del self._properties[obj_path][interface]
+                try:
+                    del self._properties[obj_path][interface]
+                except KeyError:
+                    pass
 
                 if interface == defs.DEVICE_INTERFACE:
                     self._services_cache.pop(obj_path, None)


dlech:
Is the removed object present in the properties returned by `GetManagedObjects` when this happens?

CHANGELOG.rst
```diff
@@ -26,6 +26,7 @@ Fixed
 * Fixed wrong error message for BlueZ "Operation failed with ATT error". Merged #975.
 * Fixed possible ``AttributeError`` when enabling notifications for battery service in BlueZ backend. Merged #976.
 * Fixed use of wrong enum in unpair function of WinRT backend.
+* Handle device being removed before GetManagedObjects returns in BlueZ backend.
```
dlech, line 29:
```suggestion
* Handle device being removed before GetManagedObjects returns in BlueZ backend. Fixes #996.
```

bdraco:
> Is the removed object present in the properties returned by `GetManagedObjects` when this happens?

I could only get it to happen at startup before `GetManagedObjects` returned and populated `self._properties` since the message handler fired while waiting for   `GetManagedObjects`

But it seems like it can happen later as well
https://community.home-assistant.io/t/2022-9-home-assistant-birthday-release/458283/248?u=bdraco

dlech:
That doesn't answer the question though. If the removal happens while waiting for `GetManagedObjects` to return, is the returned value aware of the removal and not include it (in which case we don't have to do anything) or does it include it and we need to manually remove it?

bdraco:
> If the removal happens while waiting for `GetManagedObjects` to return, is the returned value aware of the removal and not include it (in which case we don't have to do anything) or does it include it and we need to manually remove it?

Good point.  The returned value from `GetManagedObjects` may include the device depending on if it loses the race or not so we likely need to remove it from the results.  That means that there is also a race in where a device can get added via `InterfacesAdded` and then get cleared out when `GetManagedObjects` returns  via

https://github.com/hbldh/bleak/blob/846c35f773b9fab7b7a9bed0dd0906198333a0ed/bleak/backends/bluezdbus/manager.py#L232

bdraco:
Synchronization is hard here since we have data coming in via two channels

1. `_parse_msg`
2. the return of the `GetManagedObjects`

I was thinking we could hold all the callbacks to _parse_msg until `GetManagedObjects` returns and then reply them, but we don't really have a way to know which ones reflect the state of the bus when `GetManagedObjects` responded to

dlech:
In your tests where you reproduce the problem, are you seeing both cases? I'm hoping that BlueZ doesn't actually send any signals in the time between when it receives GetManagedObjects request and when it sends the response, in which case I would expect to never see the removed device in the return value of GetManagedObjects.

bdraco:
I'm syncing everything up and testing again to make sure I have a clean test

bdraco:
I've yet to get a signal in-between after 10 restarts

```
2022-09-10 17:13:00.170 DEBUG (MainThread) [bleak.backends.bluezdbus.manager] Sending GenManagedObjects
2022-09-10 17:13:00.258 DEBUG (MainThread) [bleak.backends.bluezdbus.manager] Got back GenManagedObjects
```

```python3
                # get existing objects after adding signal handlers to avoid
                # race condition
                logger.debug("Sending GenManagedObjects")
                reply = await bus.call(
                    Message(
                        path="/",
                        member="GetManagedObjects",
                        interface=defs.OBJECT_MANAGER_INTERFACE,
                    )
                )
                assert_reply(reply)
                logger.debug("Got back GenManagedObjects")
```

bdraco:
Well after 11 I did

```
2022-09-10 17:14:22.496 DEBUG (MainThread) [bleak.backends.bluezdbus.manager] Sending GenManagedObjects
2022-09-10 17:14:22.577 DEBUG (MainThread) [bleak.backends.bluezdbus.manager] received D-Bus signal: org.freedesktop.DBus.Properties.PropertiesChanged (/org/bluez/hci0/dev_D2_7F_24_DD_CC_C2): ['org.bluez.Device1', {'ServicesResolved': <dbus_fast.signature.Variant ('b', False)>, 'Connected': <dbus_fast.signature.Variant ('b', False)>}, []]
2022-09-10 17:14:22.580 DEBUG (MainThread) [bleak.backends.bluezdbus.manager] Got back GenManagedObjects
```

dlech:
Just because Bleak logged it in that order doesn't mean that it went over the bus in that order though, right?

bdraco:
> Just because Bleak logged it in that order doesn't mean that it went over the bus in that order though, right?

Logging is a race as well so it's possible.

bdraco:
I'd probably need to strace it to be sure but I'm not sure I can actually do that since its running in a container and I would need to attach at startup

dlech:
Can you run `dbus-monitor` in the container?

bdraco:
> 



> Can you run `dbus-monitor --system` in the container (or out of the container)?

Yes, that's what I'm trying now.

bdraco:
```
method call time=1662849236.851551 sender=:1.1753 -> destination=org.bluez serial=2 path=/; interface=org.freedesktop.DBus.ObjectManager; member=GetManagedObjects
method return time=1662849236.882229 sender=:1.5 -> destination=:1.1753 serial=3591609 reply_serial=2
```
I haven't seen anything in between `method call` and `method return` after a few restarts

bdraco:
Up to 15 and nothing in between

bdraco:
22 restarts. I think we are good

bdraco:
Up to 25 and still good

I'd wager the reason we see the interlaced one above is that when the the await returns control to the event loop, its picking up an event before returning back to process GetManagedObjects.

So I think we need to move the clear right before `GetManagedObjects` as well

bdraco:
Working that through.


1. everything is cleared
2. subscribe to InterfacesAdded/InterfacesRemoved/PropertiesChanged
3. await happens to GetManagedObjects
4. InterfacesAdded callback happens and is processed
5. InterfacesRemoved callback happens and is processed
6. GetManagedObjects is sent
7. GetManagedObjects respond is received
8. GetManagedObjects is processed

bdraco:
Another possible race
1. everything is cleared
2. subscribe to InterfacesAdded/InterfacesRemoved/PropertiesChanged
3. await happens to GetManagedObjects returning control to the event loop
4. InterfacesAdded A callback happens and is processed
5. InterfacesRemoved B callback happens and is processed
6. GetManagedObjects is sent
7. GetManagedObjects respond is received
4. InterfacesAdded C callback happens and is processed
5. InterfacesRemoved D callback happens and is processed
9. GetManagedObjects is processed

bdraco:
Since everything comes into https://github.com/Bluetooth-Devices/dbus-fast/blob/6dccab637811d53dce4f1708e9ce5d50036c065a/src/dbus_fast/message_bus.py#L731

I think its very unlikely that the above race is actually possible https://github.com/hbldh/bleak/pull/997#issuecomment-1242815879 but its no implementation detail of `dbus-fast` not a guarantee.

1. subscribe to InterfacesAdded/InterfacesRemoved/PropertiesChanged
2. await happens to GetManagedObjects returning control to the event loop
3. InterfacesAdded A callback happens and is processed
4. InterfacesRemoved B callback happens and is processed
5. GetManagedObjects is sent https://github.com/Bluetooth-Devices/dbus-fast/blob/6dccab637811d53dce4f1708e9ce5d50036c065a/src/dbus_fast/aio/message_bus.py#L323 and control is returned to the event loop
6. GetManagedObjects respond is received but not yet returned to bleak and the event loop continues
7. everything is cleared
8. GetManagedObjects is processed and returned https://github.com/Bluetooth-Devices/dbus-fast/blob/6dccab637811d53dce4f1708e9ce5d50036c065a/src/dbus_fast/aio/message_bus.py#L337
9. bleak consumes GetManagedObjects

So I think this PR is probably fine as-is

dlech:
> So I think this PR is probably fine as-is

Agreed. Thanks for all of the testing.

dlech:
Can you fix the merge conflict?

bdraco:
I'd showing clean here.  But I'll resquash

Issue #992, opened 2022-09-10
Device may already be connected when calling connect with BlueZ


bdraco:
If the device is connected, and then the application/docker container is restarted and the device is not disconnected, the next connection attempt will fail because the device is already connected on the host but we think it is not connected and a connect call will fail.

```
      "/org/bluez/hci1/dev_BD_24_6F_85_AA_61": {
        "org.freedesktop.DBus.Introspectable": {},
        "org.bluez.Device1": {
          "Address": "BD:24:6F:85:AA:61",
          "AddressType": "public",
          "Name": "Dream~BD246F85AA61",
          "Alias": "Dream~BD246F85AA61",
          "Appearance": 962,
          "Icon": "input-mouse",
          "Paired": false,
          "Trusted": false,
          "Blocked": false,
          "LegacyPairing": false,
          "RSSI": -36,
          "Connected": true,
          "UUIDs": [
            "00001800-0000-1000-8000-00805f9b34fb",
            "00001801-0000-1000-8000-00805f9b34fb",
            "0000180a-0000-1000-8000-00805f9b34fb",
            "0000ffd0-0000-1000-8000-00805f9b34fb",
            "0000ffd5-0000-1000-8000-00805f9b34fb"
          ],
          "Modalias": "usb:v045Ep0040d0300",
          "Adapter": "/org/bluez/hci1",
          "ManufacturerData": {
            "20808": {
              "__type": "<class 'bytearray'>",
              "repr": "bytearray(b'364656')"
            }
          },
          "ServicesResolved": true
        },
        "org.freedesktop.DBus.Properties": {}
      },
```

bdraco:
Need to think about how to fix this one as its not strait-forward.  I wanted to document it so I don't forget about it

bdraco:
Maybe skip calling "Connect" if the `Connected` property is already true when we get here https://github.com/hbldh/bleak/blob/aa24332a39952880005705155629df423fbad1aa/bleak/backends/bluezdbus/client.py#L158

bdraco:
Maybe could use the same check that ServicesResolved uses

dlech:
This is essentially a duplicate of #367.

[...]

Issue #309, opened 2020-09-25
Pair with devices that are not connected


hbldh:
Pairing was added in version 0.8.0. It still requires `connect` to be called first, due to that the OS native device communication objects being created in the `connect` methods (setting of the `self._requester` attributes). One could refactor this so that `pair` and `unpair` could be called on unconnected `BleakClient`s as well.

dlech:
Since pairing requires communicating with the remote device, I don't see how pairing could be done without connecting first.

dlech:
The BlueZ D-Bus API docs for "Pair" says:

> This method will connect to the remote device, initiate pairing and then retrieve all SDP records (or GATT primary services).

And the Windows API doesn't really have a notion of being "connected", so it seems like it should be possible to pair without an active gatt session. 

macOS doesn't allow explicit pairing, but rather prompts for it automatically the first time you try to access a characteristic that requires authentication.

The Android API doesn't say one way or the other. https://developer.android.com/reference/android/bluetooth/BluetoothDevice#createBond()

bojanpotocnik:
> And the Windows API doesn't really have a notion of being "connected", so it seems like it should be possible to pair without an active gatt session.

This is correct, tested with https://github.com/hbldh/bleak/pull/1133/commits/dd50aa66730f80e95ca43acd69b2e7a9f2e67043#diff-dca4aa1d374b221291b543a5d1296bbf912b43d1e2ebc51ed11c1eb153fbb89bR193-R196

PR #1002, opened 2022-09-12
added remove_device function to BleakClient interface and implement it for WinRT backend

jochenjagers:
Hello,

this PR adds a remove_device function to be able to remove a device without the need of connecting to it before.
This was inspired by the remove_device function for the bluez backend from another pull request (https://github.com/hbldh/bleak/pull/640/files#diff-dca4aa1d374b221291b543a5d1296bbf912b43d1e2ebc51ed11c1eb153fbb89bR118)
As I have no possibility to test with the bluez backend at the moment I did not copy it from there.

Additionally I moved the validate_address function out of the bluez backend utils into a generic utils file to be able to use it in the WinRT backend as well.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,7 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 Added
 -----
 * ``AdvertisementData`` class now has an attribute ``tx_power``. Merged #987.
+* Added static ``remove_device`` function to ``BleakClient`` interface and implemented it for WinRT
 
 Changed
 -------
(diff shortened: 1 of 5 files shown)


dlech:
Thanks for the PR. However, BlueZ is the only backend that has a problem with not disconnecting sometimes and it is not possible to "disconnect" in any other backend without "connecting" first. Something similar was recently discussed in #992. Perhaps the feature that was added to close that issue could be of help for your use case (it allows "connecting" in the BlueZ backend without disconnecting first)?

dlech:
Can we not just make sure that calling the existing `unpair()` method works without being connected instead of adding a new method?

dlech:
Related issue: #309

jochenjagers:
What do you think about this solution?
https://github.com/hbldh/bleak/compare/develop...jochenjagers:bleak:feature/winrt-unpair-without-connect#
It allows unpair without connection for the WinRT backend without adding a new function to the inferface.

dlech:
It will be easier to make specific comments in a pull request, but in general, it looks like the right approach.

PR #1009, opened 2022-09-13
Relax async-timeout version

devbis:
Signatures of methods haven't changed since 3.0 for use cases in bleak.  Relaxing of the dependency will allow using prebuilt version of async-timeout on OpenWrt 21.03 which has version 3.0.1

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ classifiers = [
 
 [tool.poetry.dependencies]
 python = "^3.7"
-async-timeout = "^4.0.1"
+async-timeout = ">=3.0.0"
 typing-extensions = { version = "^4.2.0", python = "<3.8" }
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -23,7 +23,7 @@ classifiers = [
 
 [tool.poetry.dependencies]
 python = "^3.7"
-async-timeout = "^4.0.1"
+async-timeout = ">=3.0.0"
```
dlech, line 26: Should we also include a max version to avoid potential breaking changes in the future?
```suggestion
async-timeout = ">= 3.0.0, < 5"
```

dlech:
Can you please add a changelog entry too?

devbis:
Done.

dlech:
Thanks!

Issue #527, opened 2021-04-25
FR: Subclassing exceptions to make them easier to catch


mihtjel:
### Description

I've been writing a piece of code that's getting .. more complicated than it probably should, I guess. But I have gotten a little annoyed by one thing in particular: The difficulty in handling the exceptions I get from Bleak:

```Logger.error(f"Bleak error encountered: {e} ({e.__class__.__name__})")```
gives me lines in my log like:
```
Bleak error encountered: org.bluez.Error.Failed (BleakDBusError)
Bleak error encountered: Device with address 01:02:03:04:05:06 was not found. (BleakError)
Bleak error encountered: Not connected (BleakError)
```
etc.

The first one is maybe difficult to do anything about, okay. But I have to do a manual parsing of the contents in my ```except BleakError as e:``` handler of the second two - and I think it would be more obvious if they were for example BleakDeviceNotFoundError and BleakNotConnectedError<sup>*</sup>.

**So, the suggestion**: Make some exception classes for the expected types of returns from the library, and use BleakError for the rest.

(Obviously I'd be willing to contribute the code as well, but I won't start working on it if it's against a project principle.)

## For example:

- BleakNotConnectedException
- BleakDeviceNotFoundException
- BleakPairingFailedException (Maybe used for Unpair as well? Or a separate one.)
- BleakBluetoothDisabledException
- BleakNotificationsAlreadyStartedException / BleakNotificationsNeverStartedException (Maybe these are genuinely errors)
- BleakReadException (when reading a characteristic fails)
- BleakWriteException (when writing a characteristic fails)

<br>
<br>
<br>

*: Or indeed BleakDeviceNotFoundException, as it is really not considered an _error_ in the classical sense - a device not being found? But that's a different matter, and an issue with the Python naming conventions.

mihtjel:
If possible, I'd like just a little feedback - if you are not directly opposed to the idea I would start working on a PR :-)

bojanpotocnik:
Just one one about
> The first one is maybe difficult to do anything about, okay.

since https://github.com/hbldh/bleak/pull/522 this one usually also has description and can be further classified into above mentioned categories.

hbldh:
I am not against this in principle, but I tend to not implement specific subclass errors until I actually need them, and I have not needed them as of yet in Bleak. 

@mihtjel you are free to try. If you can improve Bleak by doing this, I am all for it. The hassle is that there are four backends in three different OS:es to implement and test, so it is a daunting task...

[...]

PR #1012, opened 2022-09-15
allow unpair without being connected on WinRT backend

jochenjagers:
This is a alternative approach to https://github.com/hbldh/bleak/pull/1002
Calling unpair when not connected should remove all device information from the OS.

This also relates to issue https://github.com/hbldh/bleak/issues/309

### DIFF
--- a/bleak/backends/winrt/client.py
+++ b/bleak/backends/winrt/client.py
@@ -73,6 +73,22 @@ _ACCESS_DENIED_SERVICES = list(
 #     protocol_error: typing.Optional[int]
 
 
+def _address_to_int(address: str) -> int:
+    """Converts the Bluetooth device address string to its representing integer
+
+    Args:
+        address (str): Bluetooth device address to convert
+
+    Returns:
+        int: integer representation of the given Bluetooth device address
+    """
+    _address_separators = [":", "-"]
+    for char in _address_separators:
+        address = address.replace(char, "")
+
+    return int(address, base=16)
+
+
 def _ensure_success(result: Any, attr: Optional[str], fail_msg: str) -> Any:
     """
     Ensures that *status* is ``GattCommunicationStatus.SUCCESS``, otherwise
@@ -182,6 +198,18 @@ class BleakClientWinRT(BaseBleakClient):
 
     # Connectivity methods
 
+    def _create_requester(self, bluetooth_address: int):
+        args = [
+            bluetooth_address,
+        ]
+        if self._address_type is not None:
+            args.append(
+                BluetoothAddressType.PUBLIC
+                if self._address_type == "public"
+                else BluetoothAddressType.RANDOM
+            )
+        return BluetoothLEDevice.from_bluetooth_address_async(*args)
+
     async def connect(self, **kwargs) -> bool:
         """Connect to the specified GATT server.
 
@@ -206,16 +234,7 @@ class BleakClientWinRT(BaseBleakClient):
 
         logger.debug("Connecting to BLE device @ %s", self.address)
 
-        args = [
-            self._device_info,
-        ]
-        if self._address_type is not None:
-            args.append(
-                BluetoothAddressType.PUBLIC
-                if self._address_type == "public"
-                else BluetoothAddressType.RANDOM
-            )
-        self._requester = await BluetoothLEDevice.from_bluetooth_address_async(*args)
+        self._requester = await self._create_requester(self._device_info)
 
         if self._requester is None:
             # https://github.com/microsoft/Windows-universal-samples/issues/1089#issuecomment-487586755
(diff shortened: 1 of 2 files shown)


bleak/backends/winrt/client.py
```diff
@@ -452,27 +453,32 @@ async def unpair(self) -> bool:
...
-                    f"Could not unpair with device: {unpairing_result.status}"
-                )
+        unpairing_result = await device_information.pairing.unpair_async()
 
-            else:
-                logger.info("Unpaired with device.")
-                return True
+        if not self._requester:
+            # close the device object if this was created locally
+            device.close()
```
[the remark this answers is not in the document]
dlech: It is unlikely but still theoretically possible that `self._requester` could be assigned while awaiting, so perhaps we should just have a local `should_close` flag instead?

bleak/utils.py
```diff
@@ -0,0 +1,37 @@
+# -*- coding: utf-8 -*-
+import re
+
+_address_separators = [":", "-"]
+_address_regex = re.compile(
+    rf"^([0-9A-Fa-f]{{2}}[{''.join(_address_separators)}]){{5}}([0-9A-Fa-f]{{2}})$"
+)
+
+
+def validate_address(address: str) -> bool:
```
dlech, line 10: This function was not used anywhere in code, so I think we can just remove it completely.

bleak/utils.py
```diff
@@ -0,0 +1,37 @@
...
+    Args:
+        address (str): Bluetooth device address to check
+
+    Returns:
+        bool: True if the given Bluetooth device address is valid
+    """
+    return _address_regex.match(address) is not None
+
+
+def address_to_int(address: str) -> int:
```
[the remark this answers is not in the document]
jochenjagers: I moved address_to_int to WinRT's client.py.

bleak/utils.py
jochenjagers: I  moved it back to where it was. It would be great to might add a check to BaseBleakClient to check the address for validity to get early feedback if you pass something wrong to the client. But this is not the topic of this pull request.

bleak/backends/winrt/client.py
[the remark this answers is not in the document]
jochenjagers: I added a try/finally block to ensure close() is called in case of an error. I also added the should_close flag.
Is it even necessary to use self._requester here at all? Can we just create a new BluetoothLEDevice for every call to unpair an simplify this function?
Something like this:
```
        device = await BluetoothLEDevice.from_bluetooth_address_async(
            _address_to_int(self.address)
        )

        try:
            unpairing_result = await device.device_information.pairing.unpair_async()

        finally:
            device.close()

        if unpairing_result.status not in (
            DeviceUnpairingResultStatus.UNPAIRED,
            DeviceUnpairingResultStatus.ALREADY_UNPAIRED,
        ):
            raise BleakError(f"Could not unpair with device: {unpairing_result.status}")

        logger.info("Unpaired with device.")
        return True
```
What do you think?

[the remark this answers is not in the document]
dlech: Not all devices have a static address. Bonded devices with privacy enabled will have a random address that changes from time to time.  So `from_bluetooth_address_async()` should only be called as a last resort if no other information is available.

In fact, we may already have `self._device_info` even if `self_requester` is None. So we could do something like this:

```python
if self._device_info is None:
    # BleakClient was initialized with address instead of BleDevice and has not connected yet
    device = await BluetoothLEDevice.from_bluetooth_address_async(
        _address_to_int(self.address)
    )
    device_information = device.device_information
    device.close()
else:
    device_information = self._device_info

...
```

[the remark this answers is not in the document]
jochenjagers: ``self._device_info`` does only contain the Bluetooth address from advertising. It is no ``DeviceInformation`` object which can be used for unpairing.
``self._device_info = device.details.adv.bluetooth_address``

I updated the PR and added a _create_requester function which takes the address type into account and can be used by ``connect`` and ``unpair`` to create the ``BluetoothLEDevice`` object. 
For ``unpair`` the address from self._device_info is used if it is available (as it was already done in ``connect`` before). Otherwise self.address is used.
I also recognized that ``from_bluetooth_address_async`` can return None if the device is not paired and not found in the system cache. So I added a check for this case.

bleak/backends/winrt/client.py
```diff
@@ -452,27 +471,31 @@ async def unpair(self) -> bool:
...
+                ):
+                    raise BleakError(
+                        f"Could not unpair with device: {unpairing_result.status}"
+                    )
                 logger.info("Unpaired with device.")
-                return True
+            finally:
+                device.close()
+        else:
+            logger.debug("No device found to unpair.")
```
dlech, line 496: Should we raise an error instead of silently ignoring unknown devices?

dlech (commented):
I still think this is potentially problematic for resolvable private addresses since they change periodically, but I don't have a device to test that and it probably deserves consideration beyond the scope of this PR anyway.

bleak/backends/winrt/client.py
jochenjagers: I'm really not sure about this. If you only want to remove a device from the system if it is available this is not really an error. But if you expect the device to be available for unpairing this is an error of course.
But raising a ``BleakError`` makes it hard to differentiate between an error caused by the call to WinRT backend and the device not being found. This is a problem if the "device not found" error is no problem for your use case but a "backend error" is. 
If we want to raise here we should introduce a ``DeviceNotFound`` exception to be able to handle both cases in different ways.

dlech: We have a discussion in #527 about error types. I would say, let's just raise a BleakError for now and consider adding a DeviceNotFound error later in a wider scope.

jochenjagers: Sounds like a plan.

dlech (approved):
LGTM, thanks!

PR #1020, opened 2022-09-22
Notify callback rework

### LINKED ISSUE #759: proposed breaking change: change handle argument type in notification callbacks
This is a commonly reported problem:  #724, #467 (it feels like it has come up more than this, but that is all I could find).

The solutions is to use `functools.partial` to wrap the callback function to include the required extra info.

But I'm wondering if it is worth making a breaking change to make it easier and more obvious. For example, we could replace the `handle` argument in the callback function with the `BleakClient` object itself. This would only break programs that are actually using the `handle` argument. I suspect that is not many because unless you are attaching the same callback to multiple characteristics, this information is not needed.

_Originally posted by @dlech in https://github.com/hbldh/bleak/issues/754#issuecomment-1025196739_

dlech:
Fixes #759. 

See commit messages and the issue for details.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,6 +20,7 @@ Changed
 * Deprecated ``BleakScanner.set_scanning_filter()``.
 * Deprecated ``BleakClient.set_disconnected_callback()``.
 * Deprecated ``BleakClient.get_services()``.
+* Refactored common code in ``BleakClient.start_notify()``.
 
 Fixed
 -----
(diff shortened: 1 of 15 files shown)


chill-git:
Hello,

I am trying to use this new feature, but not sure how.

My use case: I am connecting to multiple identical devices, and want to set up a notification for the same characteristic in all the devices. Eg. identical temperature sensors in different rooms.
I use `start_notify` to set up the same callback function, which recovers the data, and should match it to the sensor.

How does the `BleakGATTCharacteristic` tell me which `BLEDevice` sent the notification?

Thanks!

dlech:
There is still some additional work that needs to be done to add a "parent" property to the characteristics and services to be able to get the BleakClient from the characteristic object.

PR #1022, opened 2022-09-22
added BleakDeviceNotFoundError

jochenjagers:
This is a first draft for BleakDeviceNotFoundError as discussed in https://github.com/hbldh/bleak/issues/527

### DIFF
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -11,12 +11,12 @@ from uuid import UUID
 
 import async_timeout
 from dbus_fast.aio import MessageBus
-from dbus_fast.constants import BusType, ErrorType
+from dbus_fast.constants import BusType, ErrorType, MessageType
 from dbus_fast.message import Message
 from dbus_fast.signature import Variant
 
 from ... import BleakScanner
-from ...exc import BleakDBusError, BleakError
+from ...exc import BleakDBusError, BleakError, BleakDeviceNotFoundError
 from ..characteristic import BleakGATTCharacteristic
 from ..client import BaseBleakClient, NotifyCallback
 from ..device import BLEDevice
@@ -116,8 +116,8 @@ class BleakClientBlueZDBus(BaseBleakClient):
                 self._device_info = device.details.get("props")
                 self._device_path = device.details["path"]
             else:
-                raise BleakError(
-                    "Device with address {0} was not found.".format(self.address)
+                raise BleakDeviceNotFoundError(
+                    self.address, f"Device with address {self.address} was not found."
                 )
 
         manager = await get_global_bluez_manager()
@@ -175,6 +175,15 @@ class BleakClientBlueZDBus(BaseBleakClient):
                                 path=self._device_path,
                                 member="Connect",
                             )
+                        )
+                    if (
+                        reply is not None
+                        and reply.message_type == MessageType.ERROR
+                        and reply.error_name == ErrorType.UNKNOWN_OBJECT
+                    ):
+                        raise BleakDeviceNotFoundError(
+                            self.address,
+                            f"Device with address {self.address} was not found.",
                         )
                     assert_reply(reply)
 
--- a/bleak/backends/corebluetooth/client.py
+++ b/bleak/backends/corebluetooth/client.py
@@ -17,7 +17,7 @@ from CoreBluetooth import (
 from Foundation import NSArray, NSData
 
 from ... import BleakScanner
-from ...exc import BleakError
+from ...exc import BleakError, BleakDeviceNotFoundError
 from ..characteristic import BleakGATTCharacteristic
 from ..client import BaseBleakClient, NotifyCallback
 from ..device import BLEDevice
@@ -80,8 +80,8 @@ class BleakClientCoreBluetooth(BaseBleakClient):
                 self._peripheral = device.details
                 self._central_manager_delegate = device.metadata["delegate"]
             else:
-                raise BleakError(
-                    "Device with address {} was not found".format(self.address)
+                raise BleakDeviceNotFoundError(
+                    self.address, f"Device with address {self.address} was not found"
                 )
 
         if self._delegate is None:
--- a/bleak/backends/winrt/client.py
+++ b/bleak/backends/winrt/client.py
@@ -48,7 +48,7 @@ from bleak_winrt.windows.foundation import EventRegistrationToken
 from bleak_winrt.windows.storage.streams import Buffer
 
 from ... import BleakScanner
-from ...exc import PROTOCOL_ERROR_CODES, BleakError
+from ...exc import PROTOCOL_ERROR_CODES, BleakError, BleakDeviceNotFoundError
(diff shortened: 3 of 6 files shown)


jochenjagers:
I basically only replace the BleakErrors with device not found messages by BleakDeviceNotFoundError
But this covers not all functions/backends yet.

CoreBluetooth:
- [x] connect
- [ ] ~~pair~~ (``"Pairing is not available in Core Bluetooth."``)
- [ ] ~~unpair~~ (``"Pairing is not available in Core Bluetooth."``)

WinRT:
- [x] connect
- [ ] ~~pair~~ (Pair is not possible without connect at the moment)
- [x] unpair

BlueZ:
- [x] connect
- [ ] ~~pair~~ (Pair is not possible without connect at the moment)
- [ ] ~~unpair~~ (``"Unpairing is seemingly unavailable in the BlueZ DBus API at the moment."``)

p4android:
- [ ] connect
- [ ] pair
- [ ] ~~unpair~~ (``"Unpairing is seemingly unavailable in the Android API at the moment."``)

The Android backend does not have a check for the availability of the device yet.

bleak/exc.py
```diff
@@ -8,6 +8,23 @@ class BleakError(Exception):
...
+    Exception which is raised if a device can not be found by ``connect``, ``pair`` and ``unpair``.
+    This is the case if the OS Bluetooth stack has never seen this device or it was removed and forgotten.
+    """
+
+    identifier: str
+
+    def __init__(self, identifier: str, *args: object) -> None:
+        """
+        Args:
+            identifier (str): The identifier which was used to search the device (e. g. Bluetooth address or UUID)
```
jochenjagers, line 22: I only added the identifier here for now. But it can also be the ``BLEDevice`` to share some more information about the device if they are available. I'm a bit ambivalent about this.

dlech: I think it would be best to use the `BLEDevice` object to wrap the platform-specific details.

dlech (commented):
In the `connect` methods, in addition to the optional scanner not finding the device, we need to handle the case for when the device is removed from the OS between the external scan and the connection request.

For example, in the BlueZ backend, if the "Connected" D-Bus method reply contains a `org.freedesktop.DBus.Error.UnknownObject` error, then we should raise the new `BleakDeviceNotFoundError` .

I'm not sure what it would be in the other backends, so I guess we will have to do some experimentation to find out.

bleak/exc.py
jochenjagers: I changed it to the BLEDevice.

bleak/backends/corebluetooth/client.py
```diff
@@ -112,7 +113,15 @@ def disconnect_callback():
         manager = self._central_manager_delegate
         logger.debug("CentralManagerDelegate  at {}".format(manager))
         logger.debug("Connecting to BLE device @ {}".format(self.address))
-        await manager.connect(self._peripheral, disconnect_callback, timeout=timeout)
+        try:
+            await manager.connect(
+                self._peripheral, disconnect_callback, timeout=timeout
+            )
+        except asyncio.TimeoutError as error:
+            raise BleakDeviceNotFoundError(
```
jochenjagers, line 121: I seems like corebluetooth has no real mechanism for detecting "device not found" errors. the connect function just receives a Timeout error. Can we assume that the device was not found in case of this?

bleak/backends/bluezdbus/client.py
```diff
@@ -176,6 +177,15 @@ def on_value_changed(char_path: str, value: bytes) -> None:
                                 member="Connect",
                             )
                         )
+                    if (
+                        reply is not None
+                        and reply.message_type == MessageType.ERROR
+                        and reply.error_name == ErrorType.UNKNOWN_OBJECT
```
jochenjagers, line 183: Are you sure about the UnknownObject error? The documentation does not mention it:
https://github.com/bluez/bluez/blob/master/doc/device-api.txt

jochenjagers:
For Android the documentation is ambiguous. 
On the one hand: https://developer.android.com/guide/topics/connectivity/bluetooth/connect-gatt-server
> The service will first call [getRemoteDevice()](https://developer.android.com/reference/android/bluetooth/BluetoothAdapter#getRemoteDevice(java.lang.String)) on the BluetoothAdapter to access the device. If the adapter is unable to find a device with that address, getRemoteDevice() throws an [IllegalArgumentException](https://developer.android.com/reference/java/lang/IllegalArgumentException).

On the other hand: 
https://developer.android.com/reference/android/bluetooth/BluetoothAdapter#getRemoteDevice(java.lang.String)
>A [BluetoothDevice](https://developer.android.com/reference/android/bluetooth/BluetoothDevice) will always be returned for a valid hardware address, even if this adapter has never seen that device.

bleak/backends/corebluetooth/client.py
dlech: I think we should probably just leave this as a timeout. 

If we really want a "device not found" here, I suppose we could check if the peripheral is still in the list returned by `CBCentalManager.retrievePeripheralsWithIdentifiers_` before attempting to connect.

dlech:
> For Android the documentation is ambiguous.

It sounds like we shouldn't worry about it on android then.

bleak/backends/corebluetooth/client.py
dlech: but I think it is fine to just leave it as-is for now.

jochenjagers: > If we really want a "device not found" here, I suppose we could check if the peripheral is still in the list returned by `CBCentalManager.retrievePeripheralsWithIdentifiers_` before attempting to connect.

This is already done a few lines above where the device is searched using the bleak scanner. The catch for the TimeoutError was only for the edge-case in which the device goes out of sight in between.
But I reverted this change as it might be misleading if the connection attempt times out for other reasons.

jochenjagers:
From my point of view this is ready to merge now.

bleak/backends/winrt/client.py
```diff
@@ -207,7 +207,13 @@ def _create_requester(self, bluetooth_address: int):
                 if self._address_type == "public"
                 else BluetoothAddressType.RANDOM
             )
-        return BluetoothLEDevice.from_bluetooth_address_async(*args)
+        requester = BluetoothLEDevice.from_bluetooth_address_async(*args)
+        if requester is None:
+            raise BleakDeviceNotFoundError(
+                BLEDevice(self.address),
```
dlech, line 213: Sorry I didn't think about this before when I suggested using `BLEDevice`, but if we don't have any other information besides the address, then it doesn't make sense to create an incomplete `BLEDevice` object. So probably better to just stick with the "identifier" as you originally had it.

jochenjagers: No problem. I changed it back to the identifier version. 
Nevertheless storing the whole BLEDevice might be an interesting improvement for the future.

dlech:
Thanks for updating!

jochenjagers:
> Thanks for updating!

Thanks for your fast and constructive feedback.

PR #1026, opened 2022-09-23
BleakClient: fix missing mtu_size property

dlech:
This was missed in #982 since BaseBleakClient didn't have an abstract property for this.

This adds the base abstract property to the backends, fixes the missing property in BleakClient client and adds it to the documentation.

cc @bdraco


bdraco (approved):
Tested, confirmed fixed 👍

PR #1034, opened 2022-09-27
bluez: Create a bluez manager instance per-event-loop

projectgus:
Alternative to #1031. Avoids "got Future <Future pending> attached to a different loop" if Bleak instances with BlueZ backend are constructed in context of more than one event loop (for example, in default configuration of [pytest-asyncio](https://pypi.org/project/pytest-asyncio/)).

Also enables parallel access to Bleak instances from multiple event loops, provided the same Bleak object is not shared between two event loops. Each backend manager instance establishes an independent dbus connection with a unique dbus address.

Some quick code that demonstrates using Bleak in parallel from multiple threads (and therefore multiple event loops):
<details>

```python
import asyncio
import logging
from logging import getLogger, getLevelName, Formatter, StreamHandler
import threading
import time

log = getLogger()
log.setLevel(getLevelName('INFO'))
log_formatter = Formatter("%(asctime)s [%(levelname)s] %(name)s: %(message)s [%(threadName)s] ")
console_handler = StreamHandler()
console_handler.setFormatter(log_formatter)
log.addHandler(console_handler)

def scanner_callback(device, adv_data):
    log.info(f"Callback Loop {id(asyncio.get_running_loop())} device {device}")

async def scan_coro():
    log.info(f"Scanning Loop {id(asyncio.get_running_loop())}")
    async with BleakScanner(scanner_callback):
        await asyncio.sleep(2)
    log.info("Coro done")

def run_scanner():
    log.info("Running thread")
    asyncio.run(scan_coro(), debug=True)
    log.info("Thread done")

threads = [ threading.Thread(target=run_scanner),
            threading.Thread(target=run_scanner),
            threading.Thread(target=run_scanner), ]

for t in threads:
    t.start()
    # note need to stagger starting scanning as BlueZ seems to drop responses to StartDiscovery if they happen at
    # the same time (seems to be a BlueZ limitation or bug, not anything Bleak is doing).
    time.sleep(0.4) 

[t.join() for t in threads]
```

</details>

There is also a simple pytest-asyncio case in PR #1031 that fails without this patch and passes with this patch.

### DIFF
--- a/bleak/backends/bluezdbus/manager.py
+++ b/bleak/backends/bluezdbus/manager.py
@@ -21,6 +21,7 @@ from typing import (
     Set,
     cast,
 )
+from weakref import WeakKeyDictionary
 
 from dbus_fast import BusType, Message, MessageType, Variant
 from dbus_fast.aio.message_bus import MessageBus
@@ -843,15 +844,20 @@ class BlueZManager:
             callback(device_path, device.copy())
 
 
+_global_instances: Dict[Any, BlueZManager] = WeakKeyDictionary()
+
+
 async def get_global_bluez_manager() -> BlueZManager:
     """
-    Gets the initialized global BlueZ manager instance.
+    Gets an existing initialized global BlueZ manager instance associated with the current event loop,
+    or initializes a new instance.
     """
 
-    if not hasattr(get_global_bluez_manager, "instance"):
-        setattr(get_global_bluez_manager, "instance", BlueZManager())
-
-    instance: BlueZManager = getattr(get_global_bluez_manager, "instance")
+    loop = asyncio.get_running_loop()
+    try:
+        instance = _global_instances[loop]
+    except KeyError:
+        instance = _global_instances[loop] = BlueZManager()
 
     await instance.async_init()
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,11 @@ Added
 -----
 * Added support for Python 3.11.
 
+Fixed
+-----
+* On BlueZ, support creating additional instances running on a different event
+  loops (i.e. multiple pytest-asyncio cases)
+
 `0.18.1`_ (2022-09-25)
 ======================
 
--- a/docs/backends/linux.rst
+++ b/docs/backends/linux.rst
@@ -3,12 +3,10 @@
 Linux backend
 =============
 
-The Linux backend of Bleak is written using the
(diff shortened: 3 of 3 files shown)


bleak/backends/bluezdbus/manager.py
```diff
@@ -843,15 +844,21 @@ def _run_advertisement_callbacks(
...
-    Gets the initialized global BlueZ manager instance.
+    Gets an existing initialized global BlueZ manager instance associated with the current event loop,
+    or initializes a new instance.
     """
 
-    if not hasattr(get_global_bluez_manager, "instance"):
-        setattr(get_global_bluez_manager, "instance", BlueZManager())
-
-    instance: BlueZManager = getattr(get_global_bluez_manager, "instance")
+    loop = asyncio.get_event_loop()
```
dlech, line 856: `get_event_loop()` is deprecated, this should be:

```suggestion
    loop = asyncio.get_running_loop()
```

bleak/backends/bluezdbus/manager.py
```diff
@@ -843,15 +844,21 @@ def _run_advertisement_callbacks(
             callback(device_path, device.copy())
 
 
+_instances: WeakKeyDictionary[Any, BlueZManager] = WeakKeyDictionary()
```
dlech, line 847: How about calling this `_global_instances`?

dlech: Also need to use `typing.Dict` for the type hint for compatibility with older Python versions.

bleak/backends/bluezdbus/manager.py
```diff
@@ -843,15 +844,21 @@ def _run_advertisement_callbacks(
...
-    if not hasattr(get_global_bluez_manager, "instance"):
-        setattr(get_global_bluez_manager, "instance", BlueZManager())
-
-    instance: BlueZManager = getattr(get_global_bluez_manager, "instance")
+    loop = asyncio.get_event_loop()
+    try:
+        instance = _instances[loop]
+    except KeyError:
+        instance = BlueZManager()
+        _instances[loop] = instance
```
dlech, line 861: This could be squashed in to one line.

```suggestion
        instance = _instances[loop] = BlueZManager()
```

dlech (commented):
Looks good. Just a few suggestions.

projectgus:
@dlech Thanks for the quick and thorough review! Updated accordingly.

dlech:
Thanks!

PR #1047, opened 2022-09-30
Discovered devices improvements

dlech:
* Performance improvements for keeping track of discovered devices.
* New property to get advertisement data along with discovered devices.
* Adds `rssi` attribute to `AdvertisemntData`

Additional details in the commit messages.

### DIFF
--- a/bleak/backends/bluezdbus/scanner.py
+++ b/bleak/backends/bluezdbus/scanner.py
@@ -16,6 +16,7 @@ from ..scanner import AdvertisementData, AdvertisementDataCallback, BaseBleakSca
 from .advertisement_monitor import OrPatternLike
 from .defs import Device1
 from .manager import get_global_bluez_manager
+from .utils import bdaddr_from_device_path
 
 logger = logging.getLogger(__name__)
 
@@ -89,9 +90,6 @@ class BleakScannerBlueZDBus(BaseBleakScanner):
         # kwarg "device" is for backwards compatibility
         self._adapter: Optional[str] = kwargs.get("adapter", kwargs.get("device"))
 
-        # map of d-bus object path to d-bus object properties
-        self._devices: Dict[str, Device1] = {}
-
         # callback from manager for stopping scanning if it has been started
         self._stop: Optional[Callable[[], Coroutine]] = None
 
@@ -137,7 +135,7 @@ class BleakScannerBlueZDBus(BaseBleakScanner):
         else:
             adapter_path = manager.get_default_adapter()
 
-        self._devices.clear()
+        self.seen_devices = {}
 
         if self._scanning_mode == "passive":
             self._stop = await manager.passive_scan(
@@ -192,25 +190,6 @@ class BleakScannerBlueZDBus(BaseBleakScanner):
             else:
                 logger.warning("Filter '%s' is not currently supported." % k)
 
-    @property
-    def discovered_devices(self) -> List[BLEDevice]:
-        discovered_devices = []
-
-        for path, props in self._devices.items():
-            uuids = props.get("UUIDs", [])
-            manufacturer_data = props.get("ManufacturerData", {})
-            discovered_devices.append(
-                BLEDevice(
-                    props["Address"],
-                    props["Alias"],
-                    {"path": path, "props": props},
-                    props.get("RSSI", 0),
-                    uuids=uuids,
-                    manufacturer_data=manufacturer_data,
-                )
-            )
-        return discovered_devices
-
     # Helper methods
 
     def _handle_advertising_data(self, path: str, props: Device1) -> None:
@@ -221,11 +200,6 @@ class BleakScannerBlueZDBus(BaseBleakScanner):
             path: The D-Bus object path of the device.
             props: The D-Bus object properties of the device.
         """
-
-        self._devices[path] = props
-
-        if self._callback is None:
-            return
 
         # Get all the information wanted to pack in the advertisement data
         _local_name = props.get("Name")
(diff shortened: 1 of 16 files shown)


dlech:
Hi @bdraco, could you please take a look?

bdraco:
Should be able to take a look a bit later today.

👍

bleak/backends/bluezdbus/scanner.py
```diff
@@ -244,29 +217,32 @@ def _handle_advertising_data(self, path: str, props: Device1) -> None:
             manufacturer_data=_manufacturer_data,
             service_data=_service_data,
             service_uuids=_service_uuids,
-            platform_data=props,
             tx_power=tx_power,
+            rssi=props.get("RSSI", -127),
```
bdraco, line 221: Nice!

bleak/backends/bluezdbus/utils.py
```diff
@@ -50,3 +50,16 @@ def extract_service_handle_from_path(path):
...
+    """
+    Scrape the Bluetooth address from a D-Bus device path.
+
+    Args:
+        device_path: The D-Bus object path of the device.
+
+    Returns:
+        A Bluetooth address as a string.
+    """
+    return ":".join(device_path[-17:].split("_"))
```
bdraco, line 65: Might be faster to grab it and use .replace

bleak/backends/scanner.py
```diff
@@ -3,44 +3,67 @@
...
+    Tuple,
+    Type,
+)
 
 from ..exc import BleakError
 from .device import BLEDevice
 
 
-class AdvertisementData:
+class AdvertisementData(NamedTuple):
```
bdraco, line 22: I wonder if it would be faster as a dataclass?

bdraco: Nope. Named tuple is faster

bdraco: Found this while digging around https://medium.com/@jacktator/dataclass-vs-namedtuple-vs-object-for-performance-optimization-in-python-691e234253b9

But will get better data when I run in through the profiler since we will get actual use case data vs theoretical

bdraco:
This looks awesome. Cleans things up quite a bit 👍

I added some first impression thoughts above. Some need to be validated with testing

I'll do testing and profiling when other stuff calms down later tonight and I don't have as many distractions

bleak/backends/bluezdbus/utils.py
dlech: Currently, this isn't used in any critical paths.

bleak/backends/scanner.py
dlech: Part of the reason for using NamedTuple was for the immutability, but if there is a way to speed it up, we can consider alternatives.

bdraco: Makes sense.  I don't think it's going to be an issue. Will know for sure once I do the profiles on the production systems

bdraco, line 22: Not sure if it can be avoided but
`Sep 30 23:52:42 homeassistant homeassistant[487]: TypeError: AdvertisementData.__new__() missing 1 required positional argument: 'rssi'`

bdraco:
It looks like the parse_msg path is slightly more expensive.  

New
parse_msg

Original
parse_msg_orig

bdraco:
I think this could be made more performant by storing the `_advertisement_callbacks` by adapter path and doing a dict lookup to find it by adapter path so it can avoid enumerating them and calling statswith for every callback

```
        for (callback, adapter_path) in self._advertisement_callbacks:
            # filter messages from other adapters
            if not device_path.startswith(adapter_path):
                continue

            # TODO: this should be deep copy, not shallow
            callback(device_path, device.copy())
```

dlech:
> It looks like the parse_msg path is slightly more expensive.

How much of this difference could just be noise/variance between runs? For example `_async_detection_callback` increased 0.2% but that isn't even part of Bleak. And `unpack_variants` increased as well (0.04%) and it wasn't changed.

The part that was changed - `create_or_update_device` (0.24%) vs the previous (BleakDevice) `__init__` (0.40%) seems to be an improvement though.

bdraco:
I'll do a few more.  Its not an exact science since we can't control how many advertisements come in.

bdraco:
2nd run
run2

bdraco:
3rd run
run3

bdraco:
4th run
run4

bdraco:
Only thing that still stands out as easy to optimize is `startswith` since its call 80k/min

bdraco:
All of these are 1 min snapshots with % is run time on a single core of `_parse_msg` and descendants.  `_parse_msg` is the 2nd most expensive branch running.

bdraco:
I'll revert and do another one to compare.  The system is mostly idle otherwise but its a real world performance test so everything is affected by everything else

bdraco:
Original run 2

original_run2

bleak/backends/scanner.py
balloob: Don't forget to add `frozen=True` to dataclasses when profiling 👍

bdraco:
Some good news about the hot unmarshalling path, it’s now much faster as it has an optional (auto fallback to pure python if not available) which makes this path now the hottest one in production.

I’ll see if I can get unpack variants exposed in dbus-fast since it was already added in a previous PR and add an optional (auto fallback to pure python) cython backend to shave that off a bit.

bleak/backends/scanner.py
```diff
@@ -81,6 +81,13 @@ class BaseBleakScanner(abc.ABC):
             containing this advertising data will be received.
     """
 
+    seen_devices: Dict[str, Tuple[BLEDevice, AdvertisementData]]
+    """
+    Map of device identifier to BLEDevice and most recent advertisement data.
+
+    This map must be cleared when scanning starts.
+    """
+
```
bojanpotocnik, line 114: What is the point of this class variable in addition to the instance variable defined below?

PR #1050, opened 2022-10-02
docs: add async keyword to function

chris-est:
The example only works for me when adding the async to the main() function, otherwise i get this error:

```
    async with BleakScanner(callback) as scanner:
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
SyntaxError: 'async with' outside async function
```


dlech:
Thanks!

PR #1052, opened 2022-10-02
Minimal documentation additions

jackjansen:
Minor clarifications added to the docstrings, as picked up from #904. It turns out most modifications there are no longer relevant because of the separation of the backend objects.

### DIFF
--- a/bleak/__init__.py
+++ b/bleak/__init__.py
@@ -59,7 +59,7 @@ class BleakScanner:
     Interface for Bleak Bluetooth LE Scanners.
 
     The scanner will listen for BLE advertisements, optionally filtering on advertised services or
-    other conditions, and collect a list of `BLEDevice` objects. These can subsequently be used to
+    other conditions, and collect a list of :class:`BLEDevice` objects. These can subsequently be used to
     connect to the corresponding BLE server.
 
     A BleakScanner can be used as an asynchronous context manager in which case it automatically
--- a/bleak/__init__.py
+++ b/bleak/__init__.py
@@ -62,7 +62,7 @@ class BleakScanner:
     other conditions, and collect a list of :class:`BLEDevice` objects. These can subsequently be used to
     connect to the corresponding BLE server.
 
-    A BleakScanner can be used as an asynchronous context manager in which case it automatically
+    A :class:`BleakScanner` can be used as an asynchronous context manager in which case it automatically
     starts and stops scanning.
 
     Args:
--- a/bleak/__init__.py
+++ b/bleak/__init__.py
@@ -287,7 +287,6 @@ class BleakClient:
             Callback that will be scheduled in the event loop when the client is
             disconnected. The callable must take one argument, which will be
             this client object.
-            The callback will only be called on unsolicited disconnect event.
         timeout:
             Timeout in seconds passed to the implicit ``discover`` call when
             ``address_or_ble_device`` is not a :class:`BLEDevice`. Defaults to 10.0.
@@ -419,7 +418,9 @@ class BleakClient:
         """
         Pair with the specified GATT server.
 
-        This method may not be available (or needed) for some backends.
+        This method is not available on macOS. Instead of manually initiating
+        paring, the user will be prompted to pair the device the first time
+        that a characteristic that requires authentication is read or written.
         This method may have backend-specific additional keyword arguments.
 
         Returns:
@@ -431,6 +432,11 @@ class BleakClient:
     async def unpair(self) -> bool:
         """
         Unpair from the specified GATT server.
+
+        Unpairing will also disconnect the device.
+
+        This method is only available on Windows and Linux and will raise an
+        exception on other platforms.
 
         Returns:
             Always returns ``True`` for backwards compatibility.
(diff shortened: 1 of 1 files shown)


jackjansen:
Hmm, I'm not sure why the checks failed (they're fine in my repo, which should be in an identical state), and I don't have the rights to re-run them, in case it is a transient error. Could you try rerunning the checks please?

bleak/__init__.py
```diff
@@ -58,6 +58,13 @@ class BleakScanner:
     """
     Interface for Bleak Bluetooth LE Scanners.
 
+    The scanner will listen for BLE advertisements, optionally filtering on advertised services or
+    other conditions, and collect a list of `BLEDevice` objects. These can subsequently be used to
```
dlech, line 62: Needs a prefix for the link to work.

```suggestion
    other conditions, and collect a list of :class:`BLEDevice` objects. These can subsequently be used to
```

bleak/__init__.py
```diff
@@ -58,6 +58,13 @@ class BleakScanner:
...
+    connect to the corresponding BLE server.
+
+    A BleakScanner can be used as an asynchronous context manager in which case it automatically
```
dlech, line 65:
```suggestion
    A :class:`BleakScanner` can be used as an asynchronous context manager in which case it automatically
```

bleak/__init__.py
```diff
@@ -204,7 +211,10 @@ async def get_discovered_devices(self) -> List[BLEDevice]:
     async def find_device_by_address(
         cls, device_identifier: str, timeout: float = 10.0, **kwargs
     ) -> Optional[BLEDevice]:
-        """A convenience method for obtaining a ``BLEDevice`` object specified by Bluetooth address or (macOS) UUID address.
+        """Obtain a ``BLEDevice`` for a BLE server specified by Bluetooth address or (macOS) UUID address.
+
+        This method uses the BLE scanner, and depending on the backend it may not work while a connection to a
+        BLE server is active, or it may cause that connection to be forcibly closed.
```
dlech, line 217: I don't recall issues like this ever having been reported. It is common to have problems when scanning and connecting at the same time since each requires a different type of scan. But after the peripheral is connected, I'm not aware of any problems with scanning not working or devices disconnecting. Have I missed something?

bleak/__init__.py
```diff
@@ -263,7 +277,13 @@ def apply_filter(d: BLEDevice, ad: AdvertisementData):
...
 
 class BleakClient:
-    """The Client Interface for Bleak Backend implementations to implement.
+    """The Client interface for connecting to a specific BLE GATT server and communicating with it.
+
+    A BleakClient can be used as an asynchronous context manager in which case it automatically
+    connects and disconnects.
+
+    How many BLE connections can be active simultaneously, and whether connections can be active while
+    scanning depends on the backend.
```
dlech, line 286: These depend on the Bluetooth adapter hardware, not the backend.

bleak/__init__.py
```diff
@@ -402,7 +423,10 @@ async def disconnect(self) -> bool:
 
     async def pair(self, *args, **kwargs) -> bool:
         """
-        Pair with the peripheral.
+        Pair with the specified GATT server.
+
+        This method may not be available (or needed) for some backends.
```
dlech, line 428: Let's be explicit about which backends don't support it. IIRC, only macOS doesn't support this and it will instead pair the first time a characteristic that requires authentication is accessed.

bleak/__init__.py
```diff
@@ -412,7 +436,7 @@ async def pair(self, *args, **kwargs) -> bool:
 
     async def unpair(self) -> bool:
         """
-        Unpair with the peripheral.
+        Unpair from the specified GATT server.
```
dlech, line 439: Some additional info to add: Currently this is only available on Windows. Unpairing also disconnects the device if it currently connected.

bleak/__init__.py
jackjansen: I haven't seen this issue with bleak, but for years I've never tried to do simultaneous scan and connects (because it used to never work, due to the hardware generally only having a single receiver, but maybe this is has completely changed by now). I will check whether this is indeed an issue.

jackjansen: Removed the phrase.

bleak/__init__.py
jackjansen: Fixed.

bleak/__init__.py
```diff
@@ -273,6 +287,7 @@ class BleakClient:
             Callback that will be scheduled in the event loop when the client is
             disconnected. The callable must take one argument, which will be
             this client object.
+            The callback will only be called on unsolicited disconnect event.
```
dlech, line 290: It is called on all disconnect events, not just unsolicited ones.

bleak/__init__.py
dlech:
```suggestion
        This method is not available on macOS. Instead of manually initiating
        paring, the user will be prompted to pair the device the first time
        that a characteristic that requires authentication is read or written.
```

bleak/__init__.py
dlech:
```suggestion
        Unpair from the specified GATT server.

        Unpairing will also disconnect the device.

        This method is only available on Windows and Linux and will raise an
        exception on other platforms.
```

bleak/__init__.py
dlech:
```suggestion
```

PR #1055, opened 2022-10-02
Use unpack_variants from dbus-fast

bdraco:
A faster `unpack_variants` is now provided by dbus-fast

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -21,7 +21,8 @@ Added
 
 Changed
 -------
-Changed ``AdvertisementData`` to a named tuple.
+* Changed ``AdvertisementData`` to a named tuple.
+* A faster ``unpack_variants`` is now provided by dbus-fast
 
 Fixed
 -----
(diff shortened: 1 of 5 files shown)


bdraco:
> I'm not familiar with this one, what does it mean?
> 
Subject say it all, but clearly not in this case 🤣 

> Can you please add an explanation of why we are making the change to the commit message?

Adjusted
> 
> Also, it looks like all of the dependencies were updated in poetry.lock, not just the ones related to this PR (can be fixed by reverting the changes to that file then run `poetry lock --no-update`).

Done

CHANGELOG.rst
```diff
@@ -10,6 +10,10 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 `Unreleased`_
 =============
 
+Changed
```
dlech, line 13: It looks like we have a duplicate changed section here.

bdraco: Conflict fixed correctly this time and repushed

dlech:
Thanks!

PR #1058, opened 2022-10-03
backends/bluezdbus/manager: replace startswith in hot path

dlech:
Receiving advertisements is a hot path in the code. We can improve performance by avoiding using startswith. In this case we conveniently already have the adapter path available as a D-Bus property, so we can use that instead.

Also drop comment while we are touching this since deep copy wouldn't actually make much of a difference since the dictionary value should be treated as immutable already.

### DIFF
--- a/bleak/backends/bluezdbus/manager.py
+++ b/bleak/backends/bluezdbus/manager.py
@@ -837,10 +837,9 @@ class BlueZManager:
         """
         for (callback, adapter_path) in self._advertisement_callbacks:
             # filter messages from other adapters
-            if not device_path.startswith(adapter_path):
+            if adapter_path != device["Adapter"]:
                 continue
 
-            # TODO: this should be deep copy, not shallow
             callback(device_path, device.copy())
 
 


dlech:
Addresses https://github.com/hbldh/bleak/pull/1047#issuecomment-1264148263

cc @bdraco

bdraco:
Thanks.  Will test later today. Day job is super busy today

bdraco:
no_startswith

bdraco:
Doing a baseline profile against develop now

bdraco:
Its definitely a nice improvement

All of these are taken from an Odroid N2+ so the performance is outsized compared to what some users are using (an RPI 3b). When comparing the N2+ is roughly 4x faster.  For the RPI3b users when we rolled out the first version the advertisement processing can bring the cpu to 100%+.  With all the optimizations in the stack over the last 2 months we are down to about 14%

bdraco (approved):

dlech:
> With all the optimizations in the stack over the last 2 months we are down to about 14%

Amazing!

PR #1059, opened 2022-10-03
backends/device: deprecate BLEDevice.rssi and .metadata

### LINKED ISSUE #1025: deprecate BLEDevice.metadata and BLEDevice.rssi
The [BLEDevice.metadata](https://bleak.readthedocs.io/en/latest/api/index.html?highlight=metadata#bleak.backends.device.BLEDevice.metadata) attribute was added in the early days of bleak as a hack to get advertisement data from the simple BleakScanner discovery methods without having to add a callback. The `rssi` attribute was added later for the same reason.

However, the BLEDevice and the advertisement data really should be treated as separate objects. (In fact, it #1022 we are discussing using a BLEDevice where no advertisements are involved at all.)

To still be able to to get the last advertisement data from methods like `BleakScanner.find_by_name()`, we could add an optional parameter to these methods to change the return value. If the parameter is true, then it would return a tuple of `BLEDevice` and `AdvertisementData` instead of just a `BLEDevice`.

dlech:
This make a clean separation between the BLEDevice which is the OS handle for a "known" BLE device and the advertising data for that device. There are now alternatives in place for getting advertising data paired with the BLEDevice so we can deprecate these attributes.

Fixes #1025.


bdraco:
This one is going to take a while to consume since we use `.rssi` all over the place.

If possible would you mind deferring to 0.20 so we can use 0.19 and have a full month to change all the `.rssi` calls in every library?

dlech:
Happy to defer and/or change to [DeprecationWarning](https://docs.python.org/3/library/exceptions.html#DeprecationWarning) to make the warnings less verbose.

bdraco:
Ideally deferral would be best since any type of message will trigger issue reports.   Once .19 has the RSSI we can rip everything using it and upgrade to .20 and nobody will ever see any warnings.

PR #1062, opened 2022-10-04
backends/winrt/client: retry getting services if services changed event occurs

### LINKED ISSUE #1061: access denied error in windows backend due to GattServicesChanged event
I had a very similar problem while writing data to a Nordic UART service. I had many access denies errors when I started with bleak version 0.13 and after updating to 0.15 it change more to "The object is already closed" or "A method was called at an unexpected time" (don't know the exact error messages as my system reported them in German)
The problem for me was that the device kept updating the service list during service discovery. I added a gatt_services_changed listener to the WinRT client and started over with service discovery if it is was currently running or started a new discovery before accessing the services list the next time.
The documentation of the GattServicesChanged event (https://docs.microsoft.com/en-us/uwp/api/windows.devices.bluetooth.bluetoothledevice.gattserviceschanged) says:

> This event is raised when the remote device changes its services, or an unpaired device is disconnecting. All services are cleared because unpaired device services can't be cached between connections.
The object parameter in this event is null for every event that is raised.
In your handler for this event, do the following in order to get the services available. Call [BluetoothLEDevice.GetGattServicesAsync](https://docs.microsoft.com/en-us/uwp/api/windows.devices.bluetooth.bluetoothledevice.getgattservicesasync) on the [BluetoothLEDevice](https://docs.microsoft.com/en-us/uwp/api/windows.devices.bluetooth.bluetoothledevice) that's passed to your handler. When calling GetGattServicesAsync, use the [Cached](https://docs.microsoft.com/en-us/uwp/api/windows.devices.bluetooth.bluetoothcachemode) option. This can be used to refresh the list of services and re-initialize the services on the device if they are gone.

For my device I get the event 11 times during the first service discovery and if I reload them after I got the event everything works as expected.
I can open a PR in the next days with a suggestion how to deal with this issue.

_Originally posted by @jochenjagers in https://github.com/hbldh/bleak/issues/849#issuecomment-1245549149_

dlech:
We can get an access denied error for services returned before a services changed event. This handles services changed events that are received while attempting to get services. Hopefully, this is the only time this event is received in most cases.

Fixes #1061.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -29,6 +29,7 @@ Fixed
 * On BlueZ, support creating additional instances running on a different event
   loops (i.e. multiple pytest-asyncio cases)
 * Fixed unhandled exception in ``max_pdu_size_changed_handler`` in WinRT backend. Fixes #1039.
+* Fixed stale services in WinRT backend causing ``WinError -2147483629``. Fixes #1061.
 
 `0.18.1`_ (2022-09-25)
 ======================
--- a/bleak/backends/winrt/client.py
+++ b/bleak/backends/winrt/client.py
@@ -173,6 +173,7 @@ class BleakClientWinRT(BaseBleakClient):
         else:
             self._device_info = None
         self._requester: Optional[BluetoothLEDevice] = None
+        self._services_changed_events: List[asyncio.Event] = []
         self._session_active_events: List[asyncio.Event] = []
         self._session_closed_events: List[asyncio.Event] = []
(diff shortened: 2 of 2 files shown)


dlech:
Hi @jochenjagers, can you please test and see if this fixes your issue?

jochenjagers:
Hi @dlech,
I'm currently on vacation and will test this as soon as I'm back. Probably in the week from 17. to 21. October.

PR #1100, opened 2022-10-31
Pairing agent

### LINKED ISSUE #827: programmatically accept pairing request in python without dialog interaction
* bleak version: 0.14.3
* Python version: 3.7.3
* Operating System: Win10 (and Linux)

### Description

Looking for a way to accept pairing request during integration tests without dialog action in windows. also want to understand what would be needed to do the same in a linux environment

### What I Did

used BleakClient.pair(). works if I wait and confirm pairing in dialog. but I'm looking for a way to do that by code.

dlech:
This is a work-in-progress attempt at implementing to implement programmatic pairing on Linux and Windows (alternative to the stalled #640).

I'm fairly happy with the Linux/BlueZ implementation at this point. But the Windows implementation is quite flaky need work.

Fixes #827

Issues to be resolved before merging:
- [ ] Clarify new exception types. Currently there are two new exceptions to indicate that pairing failed based on error returned by BlueZ. It was expected that these should correspond to the org.bluez.Error.Canceled and org.bluez.Error.Rejected responses sent by the agent. However, these don't seem to get passed through the Bluetooth stack and both come out the other end as a generic org.bluez.Error.AuthenticationFailed error. org.bluez.Error.AuthenticationCanceled seems to only happen when the device disconnects before pairing is complete.
- [ ] Additional pairing methods need to be implemented in BlueZ backed. Currently, only "RequestPasskey" is implemented, which corresponds to `BaseBleakAgentCallbacks.request_pin()`.
- [ ] Figure out how to properly do cancellation on Windows. It seems very easy to get the Bluetooth stack on Windows in a bad state if you end the Python process without allowing the pairing to complete. I had to reboot my computer many times while working on this to restore the Windows Bluetooth stack to a good state.
- [ ] Figure out how to get the `pair_async()` method to return when `accept()` is not called. The pair method seems to block forever unless the `accept()` method of the [DevicePairingRequestedEventArgs](https://learn.microsoft.com/en-us/uwp/api/windows.devices.enumeration.devicepairingrequestedeventargs.accept?view=winrt-22621#windows-devices-enumeration-devicepairingrequestedeventargs-accept) is called. But the user might not want to accept the pairing, so how do we handle this case? Also, if a pin is requested and you reply with a non-numeric string, the `pair_async()` method never returns instead of the expected invalid pin error.
- [ ] Settle on pin type, int vs str. BlueZ uses UInt32 for the pin while Windows uses a string. It is currently implemented here using str. But (as noted in the previous item) Windows seems to have problems with non-numeric pins, so it might be better to use int.
- [ ] Figure out how to do cross-platform async reading from stdin for the example. I found a method that works nicely on *nix, but it doesn't work on Windows. Everything I have tried on Windows so far causes a crash or locks up stdin.

### DIFF
--- a/bleak/backends/bluezdbus/agent.py
+++ b/bleak/backends/bluezdbus/agent.py
@@ -1,0 +1,171 @@
+"""
+Agent
+-----
+
+This module contains types associated with the BlueZ D-Bus `agent api
+<https://github.com/bluez/bluez/blob/master/doc/agent-api.txt>`.
+"""
+
+import asyncio
+import contextlib
+import logging
+import os
+from typing import Set, no_type_check
+
+from dbus_fast import DBusError, Message
+from dbus_fast.aio import MessageBus
+from dbus_fast.service import ServiceInterface, method
+
+from bleak.backends.device import BLEDevice
+
+from ...agent import BaseBleakAgentCallbacks
+from . import defs
+from .manager import get_global_bluez_manager
+from .utils import assert_reply
+
+logger = logging.getLogger(__name__)
+
+
+class Agent(ServiceInterface):
+    """
+    Implementation of the org.bluez.Agent1 D-Bus interface.
+    """
+
+    def __init__(self, callbacks: BaseBleakAgentCallbacks):
+        """
+        Args:
+        """
+        super().__init__(defs.AGENT_INTERFACE)
+        self._callbacks = callbacks
+        self._tasks: Set[asyncio.Task] = set()
+
+    async def _create_ble_device(self, device_path: str) -> BLEDevice:
+        manager = await get_global_bluez_manager()
+        props = manager.get_device_props(device_path)
+        return BLEDevice(
+            props["Address"], props["Alias"], {"path": device_path, "props": props}
+        )
+
+    @method()
+    def Release(self):
+        logger.debug("Release")
+
+    # REVISIT: mypy is broke, so we have to add redundant @no_type_check
+    # https://github.com/python/mypy/issues/6583
+
+    @method()
+    @no_type_check
+    async def RequestPinCode(self, device: "o") -> "s":  # noqa: F821
+        logger.debug("RequestPinCode %s", device)
+        raise NotImplementedError
+
+    @method()
+    @no_type_check
+    async def DisplayPinCode(self, device: "o", pincode: "s"):  # noqa: F821
+        logger.debug("DisplayPinCode %s %s", device, pincode)
+        raise NotImplementedError
+
+    @method()
+    @no_type_check
+    async def RequestPasskey(self, device: "o") -> "u":  # noqa: F821
+        logger.debug("RequestPasskey %s", device)
+
+        ble_device = await self._create_ble_device(device)
+
+        task = asyncio.create_task(self._callbacks.request_pin(ble_device))
+        self._tasks.add(task)
+
+        try:
+            pin = await task
+        except asyncio.CancelledError:
+            raise DBusError("org.bluez.Error.Canceled", "task canceled")
+        finally:
+            self._tasks.remove(task)
+
+        if not pin:
+            raise DBusError("org.bluez.Error.Rejected", "user rejected")
+
+        return int(pin)
+
+    @method()
+    @no_type_check
+    async def DisplayPasskey(
+        self, device: "o", passkey: "u", entered: "q"  # noqa: F821
+    ):
+        passkey = f"{passkey:06}"
+        logger.debug("DisplayPasskey %s %s %d", device, passkey, entered)
+        raise NotImplementedError
+
+    @method()
+    @no_type_check
+    async def RequestConfirmation(self, device: "o", passkey: "u"):  # noqa: F821
+        passkey = f"{passkey:06}"
+        logger.debug("RequestConfirmation %s %s", device, passkey)
+        raise NotImplementedError
+
+    @method()
+    @no_type_check
+    async def RequestAuthorization(self, device: "o"):  # noqa: F821
+        logger.debug("RequestAuthorization %s", device)
+        raise NotImplementedError
+
+    @method()
+    @no_type_check
+    async def AuthorizeService(self, device: "o", uuid: "s"):  # noqa: F821
+        logger.debug("AuthorizeService %s", device, uuid)
+        raise NotImplementedError
+
+    @method()
+    @no_type_check
+    def Cancel(self):  # noqa: F821
+        logger.debug("Cancel")
+        for t in self._tasks:
+            t.cancel()
+
+
+@contextlib.asynccontextmanager
+async def bluez_agent(bus: MessageBus, callbacks: BaseBleakAgentCallbacks):
+    agent = Agent(callbacks)
+
+    # REVISIT: implement passing capability if needed
+    # "DisplayOnly", "DisplayYesNo", "KeyboardOnly", "NoInputNoOutput", "KeyboardDisplay"
+    capability = ""
+
+    # this should be a unique path to allow multiple python interpreters
+    # running bleak and multiple agents at the same time
+    agent_path = f"/org/bleak/agent/{os.getpid()}/{id(agent)}"
+
+    bus.export(agent_path, agent)
+
+    try:
+        reply = await bus.call(
+            Message(
+                destination=defs.BLUEZ_SERVICE,
+                path="/org/bluez",
+                interface=defs.AGENT_MANAGER_INTERFACE,
+                member="RegisterAgent",
+                signature="os",
+                body=[agent_path, capability],
+            )
+        )
+
+        assert_reply(reply)
+
+        try:
+            yield
+        finally:
+            reply = await bus.call(
+                Message(
+                    destination=defs.BLUEZ_SERVICE,
+                    path="/org/bluez",
+                    interface=defs.AGENT_MANAGER_INTERFACE,
+                    member="UnregisterAgent",
+                    signature="o",
+                    body=[agent_path],
+                )
+            )
+
+            assert_reply(reply)
+
+    finally:
+        bus.unexport(agent_path, agent)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,11 +10,17 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 `Unreleased`_
 =============
 
+Added
+-----
+* Added ``BleakClient.close()`` method.
+
+
 Changed
 -------
 * Dropped ``async-timeout`` dependency on Python >= 3.11.
 * Deprecated ``BLEDevice.rssi`` and ``BLEDevice.metadata``. Fixes #1025.
 * ``BLEDevice`` now uses ``__slots__`` to reduce memory usage.
+* BlueZ no longer closes D-Bus socket on disconnect of ``BleakClient``.
 
 
 `0.19.4`_ (2022-11-06)
@@ -41,6 +47,7 @@ Fixed
 * Fixed crash when getting services in WinRT backend in Python 3.11. Fixes #1112.
 * Fixed cache mode when retrying get services in WinRT backend. Merged #1102.
 * Fixed ``KeyError`` crash in BlueZ backend when removing non-existent property. Fixes #1107.
+
 
 `0.19.1`_ (2022-10-29)
 ======================
--- a/bleak/__init__.py
+++ b/bleak/__init__.py
@@ -38,6 +38,7 @@ if sys.version_info[:2] < (3, 8):
 else:
     from typing import Literal
 
+from .agent import BaseBleakAgentCallbacks
 from .backends.characteristic import BleakGATTCharacteristic
 from .backends.client import BaseBleakClient, get_platform_client_backend_type
 from .backends.device import BLEDevice
@@ -493,7 +494,9 @@ class BleakClient:
         """
         return await self._backend.disconnect()
 
-    async def pair(self, *args, **kwargs) -> bool:
+    async def pair(
+        self, callbacks: Optional[BaseBleakAgentCallbacks] = None, **kwargs
+    ) -> bool:
         """
         Pair with the specified GATT server.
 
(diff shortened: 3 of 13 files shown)


bojanpotocnik:
Using the `pairing_agent.py` example provided in this PR
```python
    # REVISIT: implement passing capability if needed
    # "DisplayOnly", "DisplayYesNo", "KeyboardOnly", "NoInputNoOutput", "KeyboardDisplay"
    capability = ""
```
```python
    print("pairing...")
    callbacks = AgentCallbacks()
        try:
            await client.pair(callbacks)
            print("success")
        except BleakPairingCancelledError:
            print("paring was canceled")
        except BleakPairingFailedError:
            print("pairing failed (bad pin?)")
```
connection to my test device fails with (error from Silabs BLE stack):
```plain
Bonding failed (reason=0x1205 (Pairing not supported))
```
If 
When the original code is modified, so that the pairing agent has proper capabilities
```
    # REVISIT: implement passing capability if needed
    # "DisplayOnly", "DisplayYesNo", "KeyboardOnly", "NoInputNoOutput", "KeyboardDisplay"
    capability = "KeyboardDisplay"
```    
and the agent is created before connection
```python
    print("pairing...")
    callbacks = AgentCallbacks()
    client = BleakClient(device)
    # Bus needs to be connected for RegisterAgent in bluez_agent() to succeed
    await client._backend._bus.connect()
    async with bluez_agent(bus=client._backend._bus, callbacks=callbacks):
        async with client:
            try:
                await client.pair()  # Do not pass callbacks to prevent registering another agent
                print("success")
            except BleakPairingCancelledError:
                print("paring was canceled")
            except BleakPairingFailedError:
                print("pairing failed (bad pin?)")
```
the example works.

I took this concept and changed API a bit - my proposed solution is here https://github.com/hbldh/bleak/compare/pairing-agent...bojanpotocnik:bleak:pairing-agent

bojanpotocnik:
I've attached [bleak_pr1100_wireshark_captures.zip](https://github.com/hbldh/bleak/files/10030072/bleak_pr1100_wireshark_captures.zip) containing `original.pcapng` and `bojan_fork.pcapng`, captured using [nRF Sniffer for Bluetooth LE](https://www.nordicsemi.com/Products/Development-tools/nrf-sniffer-for-bluetooth-le).

Interestingly, in both cases Master (my PC) sends `LL_FEATURE_REQ` with the same feature set:
```plain
Feature Set: 0x00000000000059ff
    .... ...1 = LE Encryption: True
    .... ..1. = Connection Parameters Request Procedure: True
    .... .1.. = Extended Reject Indication: True
    .... 1... = Slave Initiated Features Exchange: True
    ...1 .... = LE Ping: True
    ..1. .... = LE Data Packet Length Extension: True
    .1.. .... = LL Privacy: True
    1... .... = Extended Scanner Filter Policies: True
    .... ...1 = LE 2M PHY: True
    .... ..0. = Stable Modulation Index - Transmitter: False
    .... .0.. = Stable Modulation Index - Receiver: False
    .... 1... = LE Coded PHY: True
    ...1 .... = LE Extended Advertising: True
    ..0. .... = LE Periodic Advertising: False
    .1.. .... = Channel Selection Algorithm #2: True
    0... .... = LE Power Class 1: False
    .... ...0 = Minimum Number of Used Channels Procedure: False
    0000 000. = Reserved: 0
    Reserved: 0000000000
```
Shortly after, Slave requests pairing with
```plain
Bluetooth Security Manager Protocol
    Opcode: Security Request (0x0b)
    AuthReq: 0x0d, Secure Connection Flag, MITM Flag, Bonding Flags: Bonding
        000. .... = Reserved: 0x0
        ...0 .... = Keypress Flag: False
        .... 1... = Secure Connection Flag: True
        .... .1.. = MITM Flag: True
        .... ..01 = Bonding Flags: Bonding (0x1)
```
to which Master responds with
```plain
Bluetooth Security Manager Protocol
    Opcode: Pairing Failed (0x05)
    Reason: Pairing Not Supported (0x05)
```
in `original.pcapng`, or
```plain
Bluetooth Security Manager Protocol
    Opcode: Pairing Request (0x01)
    IO Capability: Keyboard, Display (0x04)
    OOB Data Flags: OOB Auth. Data Not Present (0x00)
    AuthReq: 0x0d, Secure Connection Flag, MITM Flag, Bonding Flags: Bonding
        000. .... = Reserved: 0x0
        ...0 .... = Keypress Flag: False
        .... 1... = Secure Connection Flag: True
        .... .1.. = MITM Flag: True
        .... ..01 = Bonding Flags: Bonding (0x1)
    Max Encryption Key Size: 16
    Initiator Key Distribution: 0x0d, Link Key, Signature Key (CSRK), Encryption Key (LTK)
    Responder Key Distribution: 0x0f, Link Key, Signature Key (CSRK), Id Key (IRK), Encryption Key (LTK)
```
in `bojan_fork.pcapng`.

You're correct about IO capabilities - the behaviour in both cases is the same (that is, the pairing works with my fork even with `capability = ""`).

dlech:
Thanks for the analysis. I had not considered the case where the remote device initiates pairing.

Given that Windows only support custom pairing when the central initiates the pairing, it would be nice if we could only support that case to keep things consistent across platforms. Would it work to do something like this?

```python
client = BleakClient(device, ...)
await client.pair(...)
try:
     ...
finally:
    await client.disconnect()
```

There are probably some modifications needed to get the services but in BlueZ pairing will result in connecting the device. I'll have to look to see how it would work on Windows.

bojanpotocnik:
> I had not considered the case where the remote device initiates pairing.

Technically, pairing is still initiated by the central device (Pairing Request), the peripheral just requests the higher security level (the thing which triggers pairing in in Core Bluetooth).

The only difference here is that usually, peripheral devices allow service discovery without pairing, but send Security Request when central tries to read the _value_ of certain protected characteristic. We usually know that such device will require pairing to read the value, so we initiate pairing before reading the value (after service discovery), so the Security Request is never received.
My device requires higher security level for whole GATT table, meaning it requires pairing even for service discovery. As this is usually done automatically, initial service discovery will fail and peripheral will send Security Request, expecting central to initiate pairing and try again.

I am not sure if I understand what is different in solution proposed in your last comment. However, I just tested with
```python
    print("pairing...")
    callbacks = AgentCallbacks()
    async with BleakClient(device, pairing_callbacks=callbacks) as client:
        await asyncio.sleep(10)
```
and 
```python
class AgentCallbacks(BaseBleakAgentCallbacks):
    async def request_pin(self, device: BLEDevice) -> str:
```
was still invoked, so effectively you're correct about "the remote device initiates pairing".

I understand your argument about keeping things consistent across platforms, however I am fond of the idea of supporting wider range of devices, if it only cost changing API in a way that pairing callbacks are provided to the `BleakClient` constructor instead of `pair()` method and registering the pairing agent earlier.
I will try to get a chance to test what happens with my device on Windows in the following days.

dlech:
> My device requires higher security level for whole GATT table, meaning it requires pairing even for service discovery.

This seems to go against the requirement of the Bluetooth specification. For example, in the Core spec v5.3, Part G, Section 3.1, it states:

> A service declaration is an Attribute with the _Attribute Type_ set to the UUID for «Primary Service» or «Secondary Service». ... The _Attribute Permissions_ shall be read-only and shall not require authentication or authorization.

There is similar language for other attributes related to service discovery.

> however I am fond of the idea of supporting wider range of devices, if it only cost changing API

While I generally agree with this sentiment, I'm not fond of trying to support devices that don't follow the specifications.

> The behaviour for all existing devices shall stay the same - or am I missing something?

In the case of BlueZ, it means that the agent is registered as long as the device is connected instead of just during the pairing operation. It also complicates managing the lifetime of the agent, such as ensuring that the agent is unregistered when the device is disconnected. These sorts of issues are a common source of bugs and cost quite a bit of my time.

Since this only works with the BlueZ backend and is only needed devices like this that don't follow the specifications, I would suggest this workaround using the current proposed API:

```python
    callbacks = AgentCallbacks()
    client = BleakClient(device)
    await client._bus.connect()
    async with bluez_agent(client._bus, callbacks), client:
        ...
```

bojanpotocnik:
I double checked the peripheral device and done a bit more testing.

There are 2 separate things this peripheral device is using to ensure sufficient security level, both of them are per BT spec.

You are correct that _Attribute Permissions_ do not require authentication or authorization, but reading the value does - this peripheral device has _Authenticated + Bonded + Encrypted_ selected for all entries in table shows under [Set up Attribute Permissions in GATT](https://docs.silabs.com/bluetooth/latest/general/security/using-bluetooth-security-features-in-silicon-labs-bluetooth-sdk#set-up-attribute-permissions-in-gatt). So although it is indeed impossible to protect reading attribute access properties and permission levels, everything else can be protected (and most central devices try to automatically read GATT characteristic descriptors).

Additionally, there is another feature, described in Core Spec 5.2 | Vol 3, Part H, 2.4.6 Slave Security Request:
> The slave device may request security by transmitting a Security Request command to the master. When a master device receives a Security Request command it may encrypt the link, initiate the pairing procedure, or reject the request.
...
If pairing or encryption mode is not supported or cannot be initiated at the time when the slave’s Security Request command is received, then the master shall respond with a Pairing Failed command with the reason set to “Pairing Not Supported.”

This device utilizes [Increase Security Level](https://docs.silabs.com/bluetooth/latest/general/security/using-bluetooth-security-features-in-silicon-labs-bluetooth-sdk#increase-security-level) feature immediately after connection, to ensure that unauthorized central cannot be connected to this peripheral - otherwise malicious central device could connect without trying to read any GATT value (therefore not initiating pairing) and stay connected, effectively blocking other central devices to connect.

I tested with Windows 10 yesterday and I was greeted with Windows notification that the device would like to pair (system pairing agent), while Bleak was waiting for pairing process to be completed. I will continue testing in the evening - I suspect adding `DevicePairingKinds.PROVIDE_PIN` (and implementing the usage of `BaseBleakAgentCallbacks`) instead of `DevicePairingKinds.CONFIRM_ONLY` will be enough, otherwise I will try with `custom_pairing.add_pairing_requested(handler)` before initiating connection.

bleak/backends/bluezdbus/agent.py
```diff
@@ -0,0 +1,171 @@
...
+        """
+        super().__init__(defs.AGENT_INTERFACE)
+        self._callbacks = callbacks
+        self._tasks: Set[asyncio.Task] = set()
+
+    async def _create_ble_device(self, device_path: str) -> BLEDevice:
+        manager = await get_global_bluez_manager()
+        props = manager.get_device_props(device_path)
+        return BLEDevice(
+            props["Address"], props["Alias"], {"path": device_path, "props": props}
```
bojanpotocnik, line 46:
```suggestion
            props["Address"], props["Alias"], {"path": device_path, "props": props}, props.get("RSSI", -127)
```

bojanpotocnik:
During testing on Windows, I discovered that my problem can be solved with completely different and much simpler approach, as described in https://github.com/hbldh/bleak/pull/1133

bojanpotocnik:
@dlech do you intent to merge 9472a10f2324caee532c4cfdc7e9381212931e35 separatelly/earlier?

bojanpotocnik:
Though I've not been able to closely follow all the updates and improvements made to `bleak` in the past months, I believe some "_Issues to be resolved before merging_" listed in the original post might already have been taken care of (albeit possibly in a different way). 
@dlech, would you be able to update the original post to indicate what tasks still need to be completed, based on your vision of how this should be properly implemented? This would help me in revisiting this PR when I can find some spare time.

abmantis:
Could it be an option to try to move this forward in smaller increments? For example, if Windows is still problematic, an initial version could support only bluez. Also, additional pairing methods could be added later. 
That could also allow other people to contribute and implement those missing parts, since the groundwork would already be available :)

someburner:
@dlech 

this seems to work for me on linux, mostly.

but on ubuntu 22 (bluez 5.64), pairing of a device im working with results in immediate disconnect before pin code handler is able to run. log:

```
2023-11-25 00:47:54,960 - received D-Bus signal: org.freedesktop.DBus.Properties.PropertiesChanged (/org/bluez/hci0/dev_FB_FB_FB_FF_08_96): ['org.bluez.Device1', {'UUIDs': <dbus_fast.signature.Variant ('as', ['00001800-0000-1000-8000-00805f9b34fb', '00001801-0000-1000-8000-00805f9b34fb', '0000180a-0000-1000-8000-00805f9b34fb', '0000fef5-0000-1000-8000-00805f9b34fb', '18424398-7cbc-11e9-8f9e-2a86e4085a59'])>, 'ServicesResolved': <dbus_fast.signature.Variant ('b', True)>}, []]
2023-11-25 00:47:54,961 - received D-Bus signal: org.freedesktop.DBus.Properties.PropertiesChanged (/org/bluez/hci0/dev_55_F9_E7_C6_D4_05): ['org.bluez.Device1', {'RSSI': <dbus_fast.signature.Variant ('n', -96)>, 'TxPower': <dbus_fast.signature.Variant ('n', 11)>}, []]
2023-11-25 00:47:54,961 - received D-Bus signal: org.freedesktop.DBus.Properties.PropertiesChanged (/org/bluez/hci0/dev_5D_A1_70_38_ED_EA): ['org.bluez.Device1', {'RSSI': <dbus_fast.signature.Variant ('n', -85)>}, []]
2023-11-25 00:47:54,962 - received D-Bus signal: org.freedesktop.DBus.Properties.PropertiesChanged (/org/bluez/hci0/dev_FB_FB_FB_FF_08_96): ['org.bluez.Device1', {'Trusted': <dbus_fast.signature.Variant ('b', True)>}, []]
2023-11-25 00:47:54,962 - Pairing to BLE device @ FB:FB:FB:FF:08:96
2023-11-25 00:47:55,285 - received D-Bus signal: org.freedesktop.DBus.Properties.PropertiesChanged (/org/bluez/hci0/dev_FB_FB_FB_FF_08_96): ['org.bluez.Device1', {'Modalias': <dbus_fast.signature.Variant ('s', bluetooth:v00D2p0580d0100)>}, []]
2023-11-25 00:47:56,367 - RequestPasskey /org/bluez/hci0/dev_FB_FB_FB_FF_08_96
pairing failed (bad pin?)
2023-11-25 00:47:56,370 - Disconnecting (/org/bluez/hci0/dev_FB_FB_FB_FF_08_96)
2023-11-25 00:47:56,495 - received D-Bus signal: org.freedesktop.DBus.Properties.PropertiesChanged (/org/bluez/hci0/dev_FB_FB_FB_FF_08_96): ['org.bluez.Device1', {'ServicesResolved': <dbus_fast.signature.Variant ('b', False)>}, []]
2023-11-25 00:47:56,496 - received D-Bus signal: org.freedesktop.DBus.Properties.PropertiesChanged (/org/bluez/hci0/dev_FB_FB_FB_FF_08_96): ['org.bluez.Device1', {'Connected': <dbus_fast.signature.Variant ('b', False)>}, []]
2023-11-25 00:47:56,496 - Device disconnected (/org/bluez/hci0/dev_FB_FB_FB_FF_08_96)
2023-11-25 00:47:56,496 - _cleanup_all(/org/bluez/hci0/dev_FB_FB_FB_FF_08_96)
```

removing `ble_device = await self._create_ble_device(device)` and removing the device arg from callback causes it to work as expected. Could just be something wonky with my laptop / ble driver, but I have a pretty common intel ax200.

As an aside- I think it would be great to be able to provide a pin in `client.pair(pin=1234)`. Then if the device is unpaired, create a basic "predefined pin agent" that simply returns that pin. Maybe just me but I think that would handle a lot of use cases.

patman15:
I see more and more devices requiring pairing, e.g. JK BMS, see https://github.com/patman15/BMS_BLE-HA/issues/724, to be operational. If I can support somehow to move this, please let me know!

PR #1109, opened 2022-11-02
Avoid setting negotiate_unix_fd with new BlueZ to Dbus reduce latency

bdraco:
I'm not sure if this would be as reasonable replacement for `_acquire_mtu` but since it hasn't worked and always raises `StopIteration` for me I'm submitting this as a draft for feedback.

1. Avoids another round trip during auth
2. Avoids the slower unmarshaller path

### DIFF
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -495,7 +495,7 @@ class BleakClientBlueZDBus(BaseBleakClient):
         # We would rather not start notifications if we don't have to.
         if BlueZFeatures.has_mtu_property:
             self._mtu_size = (
-                min(
+                max(
                     c.max_write_without_response_size
                     for c in self.services.characteristics.values()
                 )
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -126,7 +126,8 @@ class BleakClientBlueZDBus(BaseBleakClient):
         # BlueZ quirk where notifications are automatically enabled on reconnect.
         self._bus = await MessageBus(
             bus_type=BusType.SYSTEM,
-            negotiate_unix_fd=not BlueZFeatures.has_mtu_property,
+            negotiate_unix_fd=not BlueZFeatures.has_mtu_property
+            and not BlueZFeatures.write_without_response_workaround_needed,
         ).connect()
 
         def on_connected_changed(connected: bool) -> None:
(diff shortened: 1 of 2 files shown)


bleak/backends/bluezdbus/client.py
```diff
@@ -125,7 +125,8 @@ async def connect(self, dangerous_use_bleak_cache: bool = False, **kwargs) -> bo
         # Each BLE connection session needs a new D-Bus connection to avoid a
         # BlueZ quirk where notifications are automatically enabled on reconnect.
         self._bus = await MessageBus(
-            bus_type=BusType.SYSTEM, negotiate_unix_fd=True
+            bus_type=BusType.SYSTEM,
+            negotiate_unix_fd=not BlueZFeatures.has_mtu_property,
```
dlech, line 129: `negotiate_unix_fd` is also needed when `BlueZFeatures.write_without_response_workaround_needed`

dlech: I've also been mulling over possibly using "AcquireNotify" to improve performance of notifications which would require passing file descriptors.

bleak/backends/bluezdbus/client.py
```diff
@@ -492,6 +493,16 @@ async def _acquire_mtu(self) -> None:
         """
         # This will try to get the "best" characteristic for getting the MTU.
         # We would rather not start notifications if we don't have to.
+        if BlueZFeatures.has_mtu_property:
```
dlech, line 497: I would rather not try to populate this for BlueZ since there isn't a way to get the real value. Users can just use the per-characteristic property instead.

bleak/backends/bluezdbus/client.py
bdraco: has_mtu_property is always higher but probably good to have an explicit check 👍

bdraco: 25219d00fab760dba0a8257ea6895288352e90a7

bdraco:
> What kind of performance gain do we get with this change?
> 
2-4ms per op.  Makes a difference for the HomeKit stuff since its doing quite a few read/writes to restore a connection from the encryption resume.

> Maybe negotiating fd descriptors is something we could make as a manual option (disabled by default or enabled if the user needs the MTU hack or performance benefits of the "Acquire" D-Bus APIs).

bleak/backends/bluezdbus/client.py
bdraco: Since we are effectively picking an arbitrary characteristic already, I figured it didn't make much difference.

dlech: That was just a hack that was better than nothing from before we had a better option. I would rather not try to continue to support it since there is a better option now.

bdraco:
Thanks for the feedback 👍

PR #1117, opened 2022-11-06
Reduce memory usage of BLEDevice with __slots__

bdraco:
We have thousands of these floating around in memory so it helps a bit for the RPI3b users.

Not done for `AdvertisementData` since `NamedTuple` already uses `__slots__` under the hood


dlech:
Sounds good, thanks.

bdraco:
thanks

PR #1135, opened 2022-11-22
Instantiate BaseBleakClient.services collection only after service discovery

bojanpotocnik:
Instead of using `_services_resolved: bool` and direct access to `services: BleakGATTServiceCollection` collection, this PR changes it to property with additional check whether services were actually discovered - they shall be before `connect()` returns.

### DIFF
--- a/bleak/__init__.py
+++ b/bleak/__init__.py
@@ -576,7 +576,13 @@ class BleakClient:
         Gets the collection of GATT services available on the device.
 
         The returned value is only valid as long as the device is connected.
+
+        Raises:
+            BleakError: if service discovery has not been performed yet during this connection.
         """
+        if not self._backend.services:
+            raise BleakError("Service Discovery has not been performed yet")
+
         return self._backend.services
 
     # I/O methods
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -309,8 +309,7 @@ class BleakClientBlueZDBus(BaseBleakClient):
             self._bus = None
 
             # Reset all stored services.
-            self.services = BleakGATTServiceCollection()
-            self._services_resolved = False
+            self.services = None
 
     async def disconnect(self) -> bool:
         """Disconnect from the specified GATT server.
@@ -587,7 +586,7 @@ class BleakClientBlueZDBus(BaseBleakClient):
         if not self.is_connected:
             raise BleakError("Not connected")
 
-        if self._services_resolved:
+        if self.services is not None:
             return self.services
 
         manager = await get_global_bluez_manager()
@@ -595,7 +594,6 @@ class BleakClientBlueZDBus(BaseBleakClient):
         self.services = await manager.get_services(
             self._device_path, dangerous_use_bleak_cache
         )
-        self._services_resolved = True
 
         return self.services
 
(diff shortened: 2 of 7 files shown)


bleak/backends/client.py
```diff
@@ -156,6 +154,21 @@ async def get_services(self, **kwargs) -> BleakGATTServiceCollection:
         """
         raise NotImplementedError()
 
+    @property
+    def services(self) -> BleakGATTServiceCollection:
```
dlech, line 158: Can we just add this logic to the `BleakClient.services` property instead? 

We have some users that have made custom backends, so changing services to a property here would be a breaking change for them.

bleak/backends/corebluetooth/client.py
```diff
@@ -190,20 +186,22 @@ async def get_services(self, **kwargs) -> BleakGATTServiceCollection:
...
-        if self._services is not None:
-            return self.services
+        if self._services:
+            return self._services
+
+        services = BleakGATTServiceCollection()
 
         logger.debug("Retrieving services...")
-        services = await self._delegate.discover_services()
+        cbt_services = await self._delegate.discover_services()
```
dlech, line 195: CB is the common prefix for CoreBluetooth, so `cb_services` seems like the obvious name here.

bleak/backends/bluezdbus/client.py
```diff
@@ -587,17 +586,16 @@ async def get_services(
         if not self.is_connected:
             raise BleakError("Not connected")
 
-        if self._services_resolved:
-            return self.services
+        if self._services:
```
dlech, line 589: Lets explicitly check `is not None` here (in case we ever add `__bool__` or otherwise change `BleakGATTServiceCollection` where an empty collection could be evaluated as `False`).

bleak/backends/client.py
```diff
@@ -42,9 +42,7 @@ def __init__(self, address_or_ble_device: Union[BLEDevice, str], **kwargs):
         else:
             self.address = address_or_ble_device
 
-        self.services = BleakGATTServiceCollection()
-
-        self._services_resolved = False
+        self._services: Optional[BleakGATTServiceCollection] = None
```
bojanpotocnik, line 45: > We have some users that have made custom backends, so changing services to a property here would be a breaking change for them. ~ @dlech

Do you think having `self.services` as `Optional` would be the same breaking change for them? In such case this PR is kinda useless - while testing paring I just saw that we have a redundant `_services_resolved` and I thought of removing it.

dlech: No, I don't think it is breaking. If `self.services` is never `None`, it just won't trigger the new exception.

bleak/__init__.py
```diff
@@ -576,7 +576,13 @@ def services(self) -> BleakGATTServiceCollection:
         Gets the collection of GATT services available on the device.
 
         The returned value is only valid as long as the device is connected.
+
+        Raises:
+            BleakError: if :meth:`get_services` has not been called yet after establishing the connection.
```
dlech, line 581: `get_services` is not a method on `BleakClient` so this link won't work (and really it is just an implementation detail of the backend), so we could probably just make a more generic comment here.

dlech (approved):
Looks good, thanks!

Issue #1136, opened 2022-11-22
Failure to Add Advertisement Patterns Monitor not caught


bojanpotocnik:
* bleak version: bleak-0.19.5
* Python version: Python 3.8.10
* Operating System: Ubuntu 20.04.5 LTS
* BlueZ version (`bluetoothctl -v`) in case of Linux: 5.66

### Description

I updated BlueZ and tried to scan using passive mode, using the following code:
```python
import asyncio

from bleak import BleakScanner, BLEDevice, AdvertisementData
from bleak.assigned_numbers import AdvertisementDataType
from bleak.backends.bluezdbus.advertisement_monitor import OrPattern
from bleak.backends.bluezdbus.scanner import BlueZScannerArgs, BlueZDiscoveryFilters


async def main():
    def scan_callback(bd: BLEDevice, ad: AdvertisementData):
        print("scan_callback", bd, ad)

    filters = BlueZDiscoveryFilters(Transport="le", DuplicateData=True)
    # or_patterns ar required for BlueZ passive scanning, even if empty
    or_patterns = [
        OrPattern(0, AdvertisementDataType.FLAGS, b"\x06"),
    ]

    async with BleakScanner(detection_callback=scan_callback,
                            scanning_mode="passive",
                            bluez=BlueZScannerArgs(filters=filters, or_patterns=or_patterns)):
        await asyncio.sleep(60)


asyncio.run(main(), debug=True)
```

Output of `sudo btmon --mgmt` when program reached the following line https://github.com/hbldh/bleak/blob/b8f22bc156b8d661e453f9b6a33e98155a85dc50/bleak/backends/bluezdbus/manager.py#L476
was
```plain
= bluetoothd: Path /org/bleak/1768114/139848428638416 reserved for Adv Monitor app :1.68046
```
but after execution of that line, this was the output
```plain
= bluetoothd: Path /org/bleak/1768114/139848428638416 reserved for Adv Monitor app :1.68046
@ MGMT Command: Add Advertisement Patterns Monitor (0x0052) plen 35
        Number of patterns: 1
          Pattern 1:
            AD type: 1
            Offset: 0
            Length: 1
            Value : 06
@ MGMT Event: Command Status (0x0002) plen 3
      Add Advertisement Patterns Monitor (0x0052)
        Status: Unknown Command (0x01)
= bluetoothd: Failed to Add Adv Patterns Monitor with status 0x01
= bluetoothd: Merged_pattern not found when removing monitor
```
[...]

PR #1140, opened 2022-11-23
bluez: Verify that Advertisement Monitor has been registered

bojanpotocnik:
Solves https://github.com/hbldh/bleak/issues/1136

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,7 +17,8 @@ Added
 * Added optional command line argument to use debug log level to all applicable examples.
 * Make sure the disconnect monitor task is properly cancelled on the BlueZ client.
 * Added ``BleakNoPassiveScanError`` exception.
-* Make sure that BlueZ Advertisement Monitor is actually registered when passive scanning. Solves #1136.
+* Added check to verify that BlueZ Advertisement Monitor was actually registered. Solves #1136.
+* Added ``passive_scan.py`` example.
 
 Changed
 -------
--- a/examples/passive_scan.py
+++ b/examples/passive_scan.py
@@ -1,0 +1,128 @@
+"""
+Scanner using passive scanning mode
+--------------
+
+Example similar to detection_callback.py, but using passive scanning
+
+Updated on 2022-11-24 by bojanpotocnik <info@bojanpotocnik.com>
+
+"""
+import argparse
+import asyncio
+import logging
+from typing import Optional, List, Dict, Any
+
+import bleak
+from bleak import AdvertisementData, BLEDevice, BleakScanner
+
+logger = logging.getLogger(__name__)
+
+
+def _get_os_specific_scanning_params(
+    uuids: Optional[List[str]],
+    rssi: Optional[int] = None,
+    macos_use_bdaddr: bool = False,
+) -> Dict[str, Any]:
+    def get_bluez_dbus_scanning_params() -> Dict[str, Any]:
+        from bleak.assigned_numbers import AdvertisementDataType
+        from bleak.backends.bluezdbus.advertisement_monitor import OrPattern
+        from bleak.backends.bluezdbus.scanner import (
+            BlueZScannerArgs,
+            BlueZDiscoveryFilters,
+        )
+
+        filters = BlueZDiscoveryFilters(
+            # UUIDs= Added below, because it cannot be None
+            # RSSI= Added below, because it cannot be None
+            Transport="le",
+            DuplicateData=True,
+        )
+        if uuids:
+            filters["UUIDs"] = uuids
+        if rssi:
+            filters["RSSI"] = rssi
+
+        # or_patterns ar required for BlueZ passive scanning
+        or_patterns = [
+            # General Discoverable (peripherals)
+            OrPattern(0, AdvertisementDataType.FLAGS, b"\x02"),
+            # BR/EDR Not Supported (BLE peripherals)
+            OrPattern(0, AdvertisementDataType.FLAGS, b"\x04"),
+            # General Discoverable, BR/EDR Not Supported (BLE peripherals)
+            OrPattern(0, AdvertisementDataType.FLAGS, b"\x06"),
+            # General Discoverable, LE and BR/EDR Capable (Controller), LE and BR/EDR Capable (Host) (computers, phones)
+            OrPattern(0, AdvertisementDataType.FLAGS, b"\x1A"),
+        ]
+
+        return {"bluez": BlueZScannerArgs(filters=filters, or_patterns=or_patterns)}
+
+    def get_core_bluetooth_scanning_params() -> Dict[str, Any]:
+        from bleak.backends.corebluetooth.scanner import CBScannerArgs
+
+        return {"cb": CBScannerArgs(use_bdaddr=macos_use_bdaddr)}
+
+    return {
+        "BleakScannerBlueZDBus": get_bluez_dbus_scanning_params,
+        "BleakScannerCoreBluetooth": get_core_bluetooth_scanning_params,
+        # "BleakScannerP4Android": get_p4android_scanning_params,
+        # "BleakScannerWinRT": get_winrt_scanning_params,
+    }.get(bleak.get_platform_scanner_backend_type().__name__, lambda: {})()
+
+
+async def scan(args: argparse.Namespace, passive_mode: bool):
+    def scan_callback(device: BLEDevice, adv_data: AdvertisementData):
+        logger.info("%s: %r", device.address, adv_data)
+
+    async with BleakScanner(
+        detection_callback=scan_callback,
+        **_get_os_specific_scanning_params(
+            uuids=args.services, macos_use_bdaddr=args.macos_use_bdaddr
+        ),
+        scanning_mode="passive" if passive_mode else "active",
+    ):
+        await asyncio.sleep(60)
+
+
+async def main(args: argparse.Namespace):
+    try:
+        await scan(args, passive_mode=True)
+    except bleak.exc.BleakNoPassiveScanError as e:
+        if args.fallback:
+            logger.warning(
+                f"Passive scanning not possible, using active scanning ({e})"
+            )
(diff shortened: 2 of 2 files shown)


bleak/backends/bluezdbus/manager.py
```diff
@@ -471,10 +471,46 @@ async def passive_scan(
...
+                        path=adapter_path,
+                        interface=defs.PROPERTIES_INTERFACE,
+                        member="Get",
+                        signature="ss",
+                        body=[
+                            defs.ADVERTISEMENT_MONITOR_MANAGER_INTERFACE,
+                            "SupportedFeatures",
+                        ],
+                    )
+                )
```
bojanpotocnik, line 505: Another option would be to execute
```python
                reply = await self._bus.call(
                    Message(
                        path=monitor_path,
                        member="Get",
                        signature="ss",
                        body=[defs.ADVERTISEMENT_MONITOR_INTERFACE, "Type"],
                    )
                )
```
however this fails with `['Method "Get" with signature "ss" on interface "org.freedesktop.DBus.Properties" doesn\'t exist\n']`. Considering that even `await asyncio.sleep(0.1)` was sufficient here, I left this call which always succeeds.

dlech (commented):
This is an interesting idea. Since Release being called could be for other reasons besides this one particular case, maybe it is better to just stick with the `os.uname()` check and fail early?

bleak/backends/bluezdbus/manager.py
bojanpotocnik: Tested, and it is not enough. I think that `await asyncio.sleep(0)` does yield, but it does not assure that the D-Bus event/operation queue is completely processed, or processed at all. Just for curiosity, `asyncio.sleep(0.001)` "always" does the job, `0.0001` does the job in about half of the tries, `0.00001` is never enough.

bojanpotocnik:
> Since Release being called could be for other reasons besides this one particular case ...

My first thought - before further analysing the `BlueZ` source code - was that even though the `Release` is called for other reasons, it shall for sure not be called before this function returns.

Checking the source https://git.kernel.org/pub/scm/bluetooth/bluez.git/tree/src/adv_monitor.c#n520, I see that D-Bus `Release` method is invoked only from `monitor_release()`, which is called only from `monitor_destroy()`, which is called from 5 functions, all of them related to removal of the monitor as a result of deliberate action, error, or invalid parameters.

> ... maybe it is better to just stick with the os.uname() check and fail early?

I like the `os.uname()` idea, but would prefer to solve this as BlueZ-ish as possible - maybe someone can use a custom kernel or kernel patched to have or have not support for this feature... Most probably not, but still - this was a fun/frustrating challenge for today.

bojanpotocnik:
Done some more testing, using kernel <5.10 and >5.10.

In case of kernel <5.10, `Release()` is invoked during the call of `SupportedFeatures()`, as expected.

In case of kernel >5.10, `Activate()` is called some time after this function already returns - e.g. if using:
```python
async with BleakScanner(..., scanning_mode="passive"):
    await asyncio.sleep(0.0001)
```
no `AdvertisementMonitor` method is invoked, but when this delay is increased, `Activate()` is invoked. Interestingly, `Release()` is not invoked at all, even if I add some sleep after the context manager.

I added example file: On one hand, I like to show how one can use async iterator to print detected devices, but on the other hand, it gets complicated. I can remove or simplify it if you prefer, I'll do PR for BleakScanner supporting async iteration anyway.

dlech:
The scope of this PR is creeping. Can we please stick to just one change at a time?


Regarding checking if `Release` has been called or not, I've thought about it a bit more and it seems like we should make this cancel any async task in the async context manager and raise an exception similar to how `asyncio.timeout` works. This would avoid any race condition of when `Release` is called and we wouldn't have to rely on the hack of getting a property to get the right timing and would also cover any other reason for `Release` being called at any time.



> but would prefer to solve this as BlueZ-ish as possible

This seems like it would require a change to BlueZ, such as exposing a list of supported management APIs. Since Release can be called for multiple reasons, it doesn't seem like a reliable solution for detecting support of passive scanning (but it still does seem worth handling the call to Release since we have seen that not handling can lead to waiting for advertisements that will never come).

bojanpotocnik:
I'll wait for https://github.com/hbldh/bleak/pull/1146 and https://github.com/hbldh/bleak/pull/1147 to be merged and then rebase using features from #1146.

bleak/backends/bluezdbus/manager.py
```diff
@@ -471,10 +471,47 @@ async def passive_scan(
 
                 assert_reply(reply)
 
-                # It is important to export after registering, otherwise BlueZ
-                # won't use the monitor
+                # It is important to export AFTER registering, otherwise BlueZ won't use the monitor
                 self._bus.export(monitor_path, monitor)
 
+                # During export, MGMT command "Add Advertisement Patterns Monitor (0x0052)" is executed.
```
bdraco, line 477: Would it work to detect `"org.bluez.AdvertisementMonitorManager1"` in the adapter details for this case?

https://github.com/Bluetooth-Devices/bluetooth-adapters/blob/bcadadd67150bd4648a0002dc151adc46d302732/src/bluetooth_adapters/systems/linux.py#L104

bojanpotocnik: Thank you for the suggestion, but unfortunately this would not solve this particular issue.

 The issue here is that not only that `org.bluez.AdvertisementMonitorManager1` is present in managed objects, even calling `org.bluez.AdvertisementMonitorManager1.RegisterMonitor()` succeeds. Only "later" kernel responds with `Unknown Command (0x01)` to underlying MGMT command `Add Advertisement Patterns Monitor (0x0052)` - this response is processed by the BlueZ code and registered `AdvertisementMonitor1` is destroyed, but the error is not propagated to the D-Bus caller (bleak) in any way.

bleak/backends/bluezdbus/manager.py
```diff
@@ -472,8 +472,26 @@ async def passive_scan(
...
+            # using active scanning).
+            # That is why discovery_stopped_callback is not added to self._discovery_stopped_callbacks
+            # at this point, as org.bluez.Adapter1.Discovering status is not relevant for passive
+            # scanning.
+
+            # If advertisement monitor is released before the scanning is stopped, it means that the
+            # kernel does not support passive scanning and error was returned when trying to execute
+            # MGMT command "Add Adv Patterns Monitor" (see https://github.com/hbldh/bleak/issues/1136).
+            def advertisement_monitor_released() -> None:
+                # FIXME: This is not actually raised
```
bojanpotocnik, line 487: I've again hit a roadblock and would be happy for some ideas on how to solve it. So, there are many reasons why active discovery could be (prematurely) stopped (`Discovering` set to `False`), but only a specific case when Advertising Monitor is released.
I want to use this fact to detect that the passive scanning is not supported, instead of just informing the user that something stopped the discovery (e.g. using HCI injection or mgmt commands) and it should be restarted (which would not help in this case).
Trying to achieve that `BleakNoPassiveScanError` would be raised (preferably in the `start()` method), I noticed that I came really close to the previous solution, where I "processed the D-Bus events" after exporting the Advertising Manager :thinking:

bojanpotocnik:
Based on the BlueZ tests, it seems like `Release()` being invoked for failed registration of the monitor is expected behaviour.

https://git.kernel.org/pub/scm/bluetooth/bluez.git/tree/test/example-adv-monitor#n328
```python
    # Register the app root path to expose advertisement monitors.
    # Release() should get called on monitor0 - incorrect monitor type.
    # Activate() should get called on monitor1.
    ret = app.register_app()
    if not ret:
        print('RegisterMonitor failed.')
        mainloop.quit()
        exit(-1)
```

bojanpotocnik:
After further checking BlueZ sources and tests, I used the fact that either `Activate` or `Release` will be invoked when the advertisement monitor is processed and came up with the current implementation.
One additional benefit of this implementation is that when `BleakScanner.start()` returns, we are sure that the passive scanning has actually been started (or failed).

bleak/backends/bluezdbus/manager.py
```diff
@@ -472,12 +477,28 @@ async def passive_scan(
...
+
+            # If advertisement monitor is released before the scanning is stopped, it means that the
+            # kernel does not support passive scanning and error was returned when trying to execute
+            # MGMT command "Add Adv Patterns Monitor" (see https://github.com/hbldh/bleak/issues/1136).
+            # Otherwise, monitor will be activated and start to receive advertisement packets.
+            monitor_processed = asyncio.Queue()
+
             try:
-                monitor = AdvertisementMonitor(filters, discovery_stopped_callback)
+                monitor = AdvertisementMonitor(filters, monitor_processed.put_nowait)
```
bojanpotocnik, line 495: We can also keep the `discovery_stopped_callback` functionality, if required, by adding on intermediate callback which will call `put_nowait` and then `discovery_stopped_callback`.

bojanpotocnik:
As the latest solution does not really use features from #1146, except adding a comment explaining why callback is not added to `_discovery_stopped_callbacks` when passive scanning
```python
BlueZManager -> async passive_scan() -> async with self._bus_lock:
            self._device_removed_callbacks.append(device_removed_callback_and_state)

            # Once a monitoring job is activated by BlueZ, the client can expect to get notified
            # on the targeted advertisements no matter if there is an ongoing discovery session
            # (started/stopped with org.bluez.Adapter1.StartDiscovery/StopDiscovery, as done when
            # using active scanning).
            # That is why discovery_stopped_callback is not added to self._discovery_stopped_callbacks
            # at this point, as org.bluez.Adapter1.Discovering status is not relevant for passive
            # scanning.

            # If advertisement monitor is released before the scanning is stopped, it means that the
```
I've moved it to [bojanpotocnik:1136_verify_adding_adv_monitor_rebased_1146](https://github.com/bojanpotocnik/bleak/tree/1136_verify_adding_adv_monitor_rebased_1146) and rebased this one on develop.

PR #1145, opened 2022-11-25
Specify scan mode as boolean instead of string literal

bojanpotocnik:
Split from https://github.com/hbldh/bleak/pull/1140

### DIFF
--- a/bleak/__init__.py
+++ b/bleak/__init__.py
@@ -87,10 +87,10 @@ class BleakScanner:
             Optional list of service UUIDs to filter on. Only advertisements
             containing this advertising data will be received. Required on
             macOS >= 12.0, < 12.3 (unless you create an app with ``py2app``).
-        scanning_mode:
-            Set to ``"passive"`` to avoid the ``"active"`` scanning mode.
-            Passive scanning is not supported on macOS! Will raise
-            :class:`BleakError` if set to ``"passive"`` on macOS.
+        passive:
+            Use passive instead of active scanning mode. Passive scanning
+            is not supported on macOS! Will raise :class:`BleakError` if
+            set to ``True`` on macOS.
         bluez:
             Dictionary of arguments specific to the BlueZ backend.
         cb:
@@ -114,7 +114,7 @@ class BleakScanner:
         self,
         detection_callback: Optional[AdvertisementDataCallback] = None,
         service_uuids: Optional[List[str]] = None,
-        scanning_mode: Literal["active", "passive"] = "active",
+        passive: bool = False,
         *,
         bluez: BlueZScannerArgs = {},
         cb: CBScannerArgs = {},
@@ -125,10 +125,18 @@ class BleakScanner:
             get_platform_scanner_backend_type() if backend is None else backend
         )
 
+        # Backward compatibility with scanning_mode: Literal["active", "passive"] parameter
+        scanning_mode = kwargs.get("scanning_mode")
+        if scanning_mode is not None:
+            _logger.debug(
+                f"Overwriting parameter passive={passive} with legacy scanning_mode={scanning_mode}"
+            )
+            passive = scanning_mode == "passive"
+
         self._backend = PlatformBleakScanner(
             detection_callback,
             service_uuids,
-            scanning_mode,
+            passive,
             bluez=bluez,
             cb=cb,
             **kwargs,
--- a/bleak/backends/bluezdbus/scanner.py
+++ b/bleak/backends/bluezdbus/scanner.py
@@ -6,9 +6,9 @@ from warnings import warn
 from dbus_fast import Variant
 
 if sys.version_info[:2] < (3, 8):
-    from typing_extensions import Literal, TypedDict
+    from typing_extensions import TypedDict
 else:
-    from typing import Literal, TypedDict
+    from typing import TypedDict
 
 from ...exc import BleakError
 from ..scanner import AdvertisementData, AdvertisementDataCallback, BaseBleakScanner
@@ -109,8 +109,8 @@ class BleakScannerBlueZDBus(BaseBleakScanner):
             Optional list of service UUIDs to filter on. Only advertisements
             containing this advertising data will be received. Specifying this
             also enables scanning while the screen is off on Android.
-        scanning_mode:
-            Set to ``"passive"`` to avoid the ``"active"`` scanning mode.
+        passive:
+            Use passive instead of active scanning mode.
         **bluez:
             Dictionary of arguments specific to the BlueZ backend.
         **adapter (str):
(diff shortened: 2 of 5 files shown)


dlech:
What is the motivation for these changes?

For me, a more compelling example would be something along the lines of...

```python
# pseudocode

scan_mode = "passive"

if platform_does_not_support_passive_scanning():
    scan_mode = "active"
    print("WARNING! Passive scanning is not supported on this platform.")
    print("Don't leave this running for a long time, otherwise it will drain all of your batteries!")

async with BleakScanner(scan_mode):
    wait_forever()
```

Treating active scanning as a fallback for passive scanning doesn't seem like something that should be encouraged.

bojanpotocnik:
I see your point - buried in my use-case(s), I did not think of this library being used as running scans for a long time on a battery powered devices.

- I added enum instead of string literal because, well, they're made for such use cases, and I'm influenced by the C/++ background. It's true that we have a `typing.Literal`, but it is still more convenient when using code completion.
- I added example because I spent some time searching for proper or_patterns and thought such example might be helpful.
- I added additional exception to detect this error in more correct way than catching `BleakError` and checking `if "passive" in str(e): # try active`

I like your idea of using separate function, or maybe `BleakScanner` static method to check for passive scanning support. However, I am going through the BlueZ source code for 2 days now and I'm afraid we cannot check that for BlueZ prior to starting the scanning. `SupportedMonitorTypes` and `SupportedFeatures` of `AdvertisementMonitorManager` do not perform any MGMT commands and their result depends solely on the BlueZ version, irrespective of the Linux kernel version.

If such check would be possible, then `TRY_PASSIVE` could for sure be removed from the enum. After that, the enum holds only 2 options... In that case I'm wondering why not just use `passive_scan: bool` as a parameter?

dlech:
> In that case I'm wondering why not just use `passive_scan: bool` as a parameter?

The answer is the same reason we didn't use an enum. It was already added to Bleak long ago as a string and we have just left it that way because it is working even if not ideal. 



> However, I am going through the BlueZ source code for 2 days now and I'm afraid we cannot check that for BlueZ prior to starting the scanning.

Just brainstorming here... What if we do both? The platform support check function for BlueZ could return True that BlueZ, in general, has support for passive scanning. But at runtime, raise exceptions for the cases where it won't work as we currently do (BlueZ too old, experimental features not enabled, kernel too old).

For users that need more robust checking for features at runtime but before actually scanning, we could add some aysnc helper functions specifically for BlueZ using the existing `BlueZFeatures` class. (As a side note, to support sandboxed apps as in #798, I would like to eventually git rid of the other BlueZ version checks we are currently using this class for - we could keep the checks in the `BlueZFeatures` class, but not use them in other Bleak code). 

Instead of the kernel version check, we could call a subprocess with `btmgmt revision` to test if the kernel supports a feature such as advertisement monitors.

bojanpotocnik:
I thought of implementing the check function as `BleakScanner` member, as it seemed the most fitting:

```python
class BleakScanner:
    ...

    @property
    def supports_passive_scanning(self) -> bool:
        return self._backend.supports_passive_scanning

    # Or
    async def supports_passive_scanning(self) -> bool:
        """Check whether passive scanning is possible on this platform"""
        return await self._backend.supports_passive_scanning()

class BaseBleakScanner(abc.ABC):
    ...

    # @staticmethod
    @abc.abstractmethod
    async def supports_passive_scanning(self) -> bool:
        """Check whether passive scanning is possible on this platform"""
        raise NotImplementedError()
```
which would work well, because if/when https://github.com/bluez/bluez/issues/434 is fixed, something like this could be used to check if passive scanning is supported on the system:
```python
class BleakScannerBlueZDBus(BaseBleakScanner):
    ...

    async def supports_passive_scanning(self) -> bool:
        reply = await self._bus.call(
            Message(
                path=self._adv_monitor_path,
                member="Get",
                signature="ss",
                body=[defs.ADVERTISEMENT_MONITOR_MANAGER_INTERFACE, "SupportedFeatures"],
            )
        )
        return reply.message_type != MessageType.ERROR
```

However, downside of such approach is that `BleakScanner._backend` is chosen only after `BleakScanner` is instantiated, so one cannot use it in the context manager as
```python
async with BleakScanner(
    scan_callback,
    scanning_mode="passive" if await BleakScanner.supports_passive_scanning() else "active"
):
    ...
```

Here something like "try passive" would come handy... or we make a truly separate function, similar to `get_platform_scanner_backend_type()` - but I'm not sure why I don't like this "abstract function" approach (and another D-Bus connection needs to be made).

bojanpotocnik:
I've started implementing it in a way
```python
class BleakScanner:
    @staticmethod
    async def supports_passive_scanning(backend: Optional[Type[BaseBleakScanner]] = None) -> bool:
        """Check if passive scanning mode is possible in this platform"""
        return await (get_platform_scanner_backend_type() if backend is None else backend).supports_passive_scanning()

class BaseBleakScanner(abc.ABC):
    @staticmethod
    @abc.abstractmethod
    async def supports_passive_scanning() -> bool:
        """Check if passive scanning mode is possible in this platform"""
        raise NotImplementedError()

class BlueZManager:
    async def supports_passive_scanning(self, adapter_path: str) -> bool:
        reply = await self._bus.call(
            Message(
                path=adapter_path,
                member="Get",
                signature="ss",
                body=[
                    defs.ADVERTISEMENT_MONITOR_MANAGER_INTERFACE,
                    "SupportedFeatures",
                ],
            )
        )
        return reply.message_type != MessageType.ERROR

class BleakScannerBlueZDBus(BaseBleakScanner):
    @staticmethod
    async def supports_passive_scanning() -> bool:

        return await manager.supports_passive_scanning(manager.get_default_adapter())
```
but what about
```python
        if self._adapter:
            adapter_path = f"/org/bluez/{self._adapter}"
        else:
            adapter_path = manager.get_default_adapter()
```
in `BleakScannerBlueZDBus.supports_passive_scanning()`?
No matter how I try to turn it around, I come back to requiring `BleakScannerBlueZDBus` instantiated with the actual arguments which will later be used for scanning. So I would need to do something like
```python
scanner = BleakScanner(scan_callback, adapter=adapter)
if scanner.supports_passive_scanning():
    scanner.scanning_mode = "passive"  # This property does not exist
async with scanner:
    ...
```
Ok, let's say we are OK with this. However, because in case of BlueZ, missing kernel support can only be detected after scanning is started
```python
async with scanner:
    ...
```
we still need to try/except `BleakNoPassiveScanError` at this point, and this cannot be removed easily to support different (older) kernel and BlueZ versions (some of the future BlueZ will hopefully have this fixed - https://github.com/bluez/bluez/issues/434).

So (for now), if try/except must be there and handled anyway, it might just be the sole way of detecting the unsupported passive scanning.

bojanpotocnik:
One thing observed when testing example added in #1140, considering the usage of string literals in typing.

```python
async with BleakScanner(scan_callback, scanning_mode="passive" if passive_mode_supported else "active"):
    await asyncio.sleep(60)
```
causes PEP Check warning
```
Expected type 'Literal["active", "passive"]', got 'str' instead
```
which does not happen with enums.

bojanpotocnik:
This is now kept open just to avoid warning in the previous comment. If you think it is not worth the change, it can as well be closed.

PR #1153, opened 2022-11-28
Examples: added debug flag

lorenzofelletti:
Added a debug flag to all applicable examples that enables DEBUG level logging instead of default INFO level.
The flag can be called with either `-d` or `--debug`.

### DIFF
--- a/examples/async_callback_with_queue.py
+++ b/examples/async_callback_with_queue.py
@@ -122,7 +122,7 @@ if __name__ == "__main__":
 
     args = parser.parse_args()
 
-    log_level = "DEBUG" if args.debug else "INFO"
+    log_level = logging.DEBUG if args.debug else logging.INFO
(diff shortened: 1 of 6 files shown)


examples/async_callback_with_queue.py
```diff
@@ -118,6 +113,19 @@ async def main(args: argparse.Namespace):
...
+    parser.add_argument(
+        "-d",
+        "--debug",
+        action="store_true",
+        help="sets the logging level to debug",
+    )
+
     args = parser.parse_args()
 
+    log_level = "DEBUG" if args.debug else "INFO"
```
dlech, line 125: How about using the constants from the logging module instead of strings?

lorenzofelletti: yes, maybe it is better

dlech:
Thanks!

Issue #1154, opened 2022-11-28
Improving disconnected_callback readability/semantics


lorenzofelletti:
* bleak version:
* Python version:
* Operating System:
* BlueZ version (`bluetoothctl -v`) in case of Linux:

### Description

Is there a reason why async `disconnected_callback`s are not supported?

Following the provided example, one should use this callback to set some disconnected event, then wait for this event to be set to do things that are meant to do on disconnect.

In my opinion, this behavior can be improved by:

- providing a method/constructor property like `events_to_set_on_disconnection` accepting a list of events that will be set on disconnect, this way making more clear the behavior to a reader (so that just by reading the method call/constructor invocation one knows exactly what event(s) to look for, without having to read the callback definition first)
- providing support for `async` callbacks, so that one can put all its disconnection logic in a single place, the callback, instead of using the callback to set an event, and putting in another place a wait for this event.

PR #1155, opened 2022-11-29
Added list of events to set on disconnection to corebluetooth

lorenzofelletti:
This is related to #1154. It adds a keyword argument to pass a list of events that are set on disconnection. It could be a way to make the intention of the developer clearer to a reader than passing a `disconnected_callback` setting the events.

I implemented a list because maybe someone would want more than one event to be set on disconnection for whatever reason, but I don't know if this is an overkill and switching to a single event is indeed a better alternative.

This commit was tested on a MacBook Air with M1 and macOS 12.6, and was working.
Similar changes may be easily implemented for other backends too.

### DIFF
--- a/bleak/backends/client.py
+++ b/bleak/backends/client.py
@@ -10,7 +10,7 @@ import asyncio
 import os
 import platform
 import uuid
-from typing import Callable, Optional, Type, Union
+from typing import Callable, List, Optional, Type, Union
 from warnings import warn
 
 from ..exc import BleakError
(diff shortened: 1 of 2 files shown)


dlech:
Thanks for the PR. This seems like an unnecessary feature to me since it is already possible to do anything, including triggering multiple asyncio events, using the existing `disconnected_callback`.

dlech:
> it cannot be `async`, thus it is not possible to perform asynchronous tasks within it

It is possible to do this by using `asyncio.create_task()` or `asyncio.get_running_loop().call_soon()` to start an async task from this callback.


> Thus, in my opinion, it should either be implemented an async callback, or removed the disconnected_callback, and instead implemented a list of events to set.

This would be a major breaking change.

PR #1159, opened 2022-12-01
bluezdbus/client: Make sure the disconnect monitor task is properly cancelled

arthur-proglove:
Sometimes the exception on connect is raised so early that the disconnect monitor task does not get the disconnect event set because the on_connected_changed callback is already unregistered and the task stays in pending state causing asyncio to throw an exception:
  Task was destroyed but it is pending!

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,7 @@ Added
 * Added optional hack to use Bluetooth address instead of UUID on macOS.
 * Added ``BleakScanner.find_device_by_name()`` class method.
 * Added optional command line argument to use debug log level to all applicable examples.
+* Make sure the disconnect monitor task is properly cancelled on the BlueZ client.
 
 Changed
 -------
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -161,6 +161,8 @@ class BleakClientBlueZDBus(BaseBleakClient):
         )
         self._remove_device_watcher = lambda: manager.remove_device_watcher(watcher)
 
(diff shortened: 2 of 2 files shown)


dlech (commented):
Thanks for the PR.

It would be nice if we could fix this without introducing additional object state variables since that is what triggered the bug in the first place.

Does something like this solve the problem?

```diff
diff --git a/bleak/backends/bluezdbus/client.py b/bleak/backends/bluezdbus/client.py
index 8bf88ff..b42bf02 100644
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -197,10 +197,12 @@ class BleakClientBlueZDBus(BaseBleakClient):
                 self._is_connected = True
 
                 # Create a task that runs until the device is disconnected.
-                self._disconnect_monitor_event = asyncio.Event()
+                self._disconnect_monitor_event = (
+                    local_disconnect_monitor_event
+                ) = asyncio.Event()
                 asyncio.ensure_future(
                     self._disconnect_monitor(
-                        self._bus, self._device_path, self._disconnect_monitor_event
+                        self._bus, self._device_path, local_disconnect_monitor_event
                     )
                 )
 
@@ -246,6 +248,9 @@ class BleakClientBlueZDBus(BaseBleakClient):
 
                 raise
         except BaseException:
+            # this effectively cancels the disconnect monitor in case the event
+            # was not triggered by a D-Bus callback
+            local_disconnect_monitor_event.set()
             self._cleanup_all()
             raise
 ```

arthur-proglove:
Seems to work as well ;-). I updated the PR.

PR #1160, opened 2022-12-01
Proposal for adding a callback for when peripherals modify services

lorenzofelletti:
This is a proposal for adding a callback that is called when a peripheral modifies its services.

This is useful in the case the peripheral invalidates one or more of its services and I want to trigger a disconnection if the invalidated services are mandatory for the functioning of the application.

I implemented it for the CoreBluetooth backend only as of now, but it should be easy to implement it for other backends as well if needed.

At the moment, the callback takes a list of invalidates services in input, that are the UUIDs of the `NSArray` of invalidated services. It should probably be implemented a utility function to help standardize the UUIDs (because now only a 4chars UUID is returned).
Maybe, a representation of the peripheral invalidating the services should be provided as input too to the callback, but I didn't find a way to convert the peripheral object to a platform independent representation of it.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,7 @@ Added
 -----
 * Added optional hack to use Bluetooth address instead of UUID on macOS.
 * Added ``BleakScanner.find_device_by_name()`` class method.
+* Added services_modified_callback to ``BleakClient``.
 
 Changed
 -------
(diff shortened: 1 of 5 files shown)


dlech:
This will be challenging to implement in a way that is consistent on all backends. For example, on Windows, we get the [GattServicesChanged](https://learn.microsoft.com/en-us/uwp/api/windows.devices.bluetooth.bluetoothledevice.gattserviceschanged?view=winrt-22621) event called for each service that is discovered.

PR #1175, opened 2022-12-19
Add return type None to some scanner methods

koenvervloesem:
This PR adds some type hints for the return type of some scanner methods, as they were reported by mypy in strict mode for my application using Bleak.


dlech:
Nice one, thanks!

PR #1182, opened 2022-12-23
bluezdbus: Allow hardcoding a UID with BLEAK_DBUS_AUTH_UID

### LINKED ISSUE #1181: Hardcoding DBus user UID
* bleak version: `develop`
* Python version: `3.10.6`
* Operating System: Linux
* BlueZ version (`bluetoothctl -v`) in case of Linux: `5.64`

### Description

I'd like to connect to the system DBus via a Docker container that uses a user namespace. This currently fails, as the DBus client implementation passes along the remapped UID (ex: `0`) instead of the actual UID on the host (ex: `1000000`). I've submitted a small [change](https://github.com/Bluetooth-Devices/dbus-fast/pull/189) against `dbus-fast` (available in v1.83.0) that allows a hardcoded UID to be specified. What would be the best way to expose this functionality within `bleak`?

### Logs

n/a

kiwiz:
Closes #1181


bdraco:
Don't forget the bump the minimum `dbus-fast` version

bdraco:
Also if you want to use it in Home Assistant its dbus-fast is hard pinned so you'll need to do a PR to bump that as well https://github.com/home-assistant/core/blob/dev/homeassistant/components/bluetooth/manifest.json#L14

bleak/backends/bluezdbus/utils.py
```diff
@@ -43,3 +45,17 @@ def bdaddr_from_device_path(device_path: str) -> str:
...
+    try:
+        uid = int(os.environ("BLEAK_DBUS_UID"))
+    except ValueError:
+        pass
+
+    auth = None
+    if uid:
+        auth = AuthExternal(uid=uid)
+
+    return auth
```
dlech, line 61: This would not allow `BLEAK_DBUS_UID=0`. We can allow that by using `None` instead of `0`.

```suggestion
def get_dbus_authenticator():
    uid = None
    try:
        uid = int(os.environ("BLEAK_DBUS_UID"))
    except ValueError:
        pass

    auth = None
    if uid is not None:
        auth = AuthExternal(uid=uid)

    return auth
```

docs/backends/linux.rst
```diff
@@ -41,6 +41,12 @@ individual Bleak objects should not be shared between event loops. Otherwise,
 RuntimeErrors similar to ``[...] got Future <Future pending> attached to a
 different loop`` will be thrown.
 
+User Namespaces
```
dlech, line 44: We have another similar request for anonymous d-bus authentication at https://github.com/hbldh/bleak/pull/796 that was never resolved. So perhaps it makes sense to make this section about "D-Bus authentication" in general instead of just user namespaces. Then we can add to it in the other PR.

Also, to start a pattern, perhaps we could make this environment variable `BLEAK_DBUS_AUTH_UUID` and use `BLEAK_DBUS_AUTH_ANON` in the other PR.

bleak/backends/bluezdbus/utils.py
kiwiz: Good catch.

docs/backends/linux.rst
kiwiz: Sounds reasonable, will do!

To clarify, do you mean `BLEAK_DBUS_AUTH_UID` (vs `UUID`)?

dlech: I meant that we should replace `BLEAK_DBUS_UUID` with `BLEAK_DBUS_AUTH_UUID` in this PR.

kiwiz: Yup, I get that - but do you explicitly want it to be [**U** **U** **I** **D**](https://en.wikipedia.org/wiki/Universally_unique_identifier)?

dlech: Ha, oops, didn't even notice that. Should be UID, of course!

kiwiz:
Tested with:
```
sudo docker run --rm -it --entrypoint /bin/bash -v /run/dbus:/run/dbus:ro python:3
# In container
pip install https://github.com/kiwiz/bleak/archive/refs/heads/develop.zip
BLEAK_DBUS_AUTH_UID=800000 python -c 'import bleak; import asyncio; asyncio.run(bleak.BleakScanner.discover())'
```

kiwiz:
My bad - `poetry.lock` should be up-to-date now.

dlech:
Can you run `poetry lock --no-update` instead so it only updates the one package?

kiwiz:
Hmm, that appears to generate the same file. Could this be due to an update to the `poetry.lock` format? I'm running poetry v1.3.1

dlech:
OK, I guess there is a new version of Poetry. Let's do a separate commit to update the lock file first.

kiwiz:
Split out `poetry.lock` format update into a separate commit.

poetry.lock
```diff
@@ -264,47 +264,44 @@ toml = ["tomli"]
...
+    {file = "dbus_fast-1.83.1-pp39-pypy39_pp73-manylinux_2_17_i686.manylinux_2_5_i686.manylinux1_i686.manylinux2014_i686.whl", hash = "sha256:bedbbb9692da11d0bc43a305655d82c18dbb876001133319573d40a0de38f883"},
+    {file = "dbus_fast-1.83.1-pp39-pypy39_pp73-manylinux_2_17_x86_64.manylinux2014_x86_64.whl", hash = "sha256:b95f7cdcf1773ab3e9817a0a476791d23faa8d0fc2cab3e6c7537c17898a50c0"},
+    {file = "dbus_fast-1.83.1.tar.gz", hash = "sha256:32269bee5d30b3ccc800eccac0b06bb233c514efd2871ebc649f11a987ee8d33"},
 ]
 
 [package.dependencies]
-async-timeout = ">=3.0.0"
-
-[package.extras]
-docs = ["Sphinx (>=5.1.1,<6.0.0)", "myst-parser (>=0.18.0,<0.19.0)", "sphinx-rtd-theme (>=1.0.0,<2.0.0)", "sphinxcontrib-asyncio (>=0.3.0,<0.4.0)", "sphinxcontrib-fulltoc (>=1.2.0,<2.0.0)"]
```
kiwiz, line 149: Not sure why this was removed by `poetry`.

kiwiz: Oh, I see. These were marked as deps of `dbus-fast` in `1.22.0`.

dlech:
Nice work, thanks!

Issue #1021, opened 2022-09-22
make BleakGATTService, BleakGATTCharacteristic, BleakGATTClient platform agnostic


dlech:
Currently, there is a backend-specific implementation for `BleakGATTService`, `BleakGATTCharacteristic` and `BleakGATTClient`. However, these should only contain static data, so all data can be passed into the constructor instead of overriding properties. The `obj` property is the only platform-dependent property an can be typed as `typing.Any`.

These types can then be moved out of the `backends` module (as suggested in https://github.com/hbldh/bleak/pull/904#issue-1320815388).

The non-public methods like `get_descriptor` and `add_descriptor` should be made private (by adding an underscore prefix) when we do this.

Also, new properties should be added (as suggested in https://github.com/hbldh/bleak/issues/759#issuecomment-1030940627) to navigate back up the tree. For example, the descriptor class should have a `characteristic` property to get the parent characteristic and the service class should have a `client` property to get the BleakClient. When this is done, we can deprecate the properties like `service_uuid` since we will be able to write `service.uuid` instead.

PR #1194, opened 2023-01-06
Add new backend: Silicon Labs BGAPI

karlp:
This supports the (very nice) bleak API for any device that uses the Silicon Labs BGAPI protocol.  This is mostly a Silabs EFR32 type device running "NCP" firmware, like so:
Block diagram of NCP

This is "cross platform" in an entirely different sense :)

I've implemented this in a couple of commits, to make it a little easier to review, but I'm happy to go back and squash/split further if desired.  It's largely a) scanner, b) client basics, c) client remainder, d) docs, but I've left a "quirk" commit standalone so it stands out what it's doing.  I only found the poetry reformatting late in the game, so I've left it standalone for now.

I'm certain there are holes or areas that I've not discovered/considered so far, but it's already ~equivalent to the bluez backend, only with less surprises and hidden quirks, wrt my own usecases :)


dlech:
This seems like a nice option to work around OS quirks for sure. 

I'm a bit hesitant about picking this up from a maintenance point of view. At a minimum, I would need somebody to send me one of these adapters so I could test it. If it doesn't add much time to the manual checks I do before releases, I will consider it. But if it just takes too much time and isn't used by many, I'm afraid it will end up broken and not well supported like the Android backend.

karlp:
I completely understand the maintenance question of an extra backend like this :)  I just felt it was at a stage that it was worth sharing and seeing what the feelings on this were.  It's definitely still got bugs, and I'll keep pushing and rebasing this as best I can while I work through them.  If nothing else, it can live in our fork as long as it needs.

I'm _more_ than happy to just ship you a couple of adapters though, if that's all it would take :)  
I've been using https://eu.mouser.com/ProductDetail/Silicon-Labs/XG24-EK2703A and https://eu.mouser.com/ProductDetail/Silicon-Labs/SLTB010A locally, as they're "easy" but you'd probably want some pre-programmed?  I'd be happy to ship you flashed versions if desired :)

docs/backends/bgapi.rst
```diff
@@ -0,0 +1,62 @@
...
+This backend cannot be automatically selected based on system type, so must be manually
+selected, using code such as:
+
+.. code-block:: python
+
+    async with bleak.BleakClient(
+            "11:aa:bb:cc:22:33",
+            backend=bleak.backends.bgapi.client.BleakClientBGAPI,
+            bgapi="/home/karlp/SimplicityStudio/SDKs/gecko-4.2.0/protocol/bluetooth/api/sl_bt.xapi",
+            adapter="/dev/ttyACM1",
```
dlech, line 33: Instead of requiring this, it would be nice to use environment variables to pass this information. Then any Bleak-based app could be run using that backend without having to modify any code.

dlech (commented):
Feel free to reach out via email about sending some adapters. I can't make a 100% promise that if you send them I will have time for this, but it does look fairly simple at first glance.

It would also be nice to do https://github.com/hbldh/bleak/issues/1021 first so we don't have to duplicate those classes here.

docs/backends/bgapi.rst
karlp: sounds good, this seemed like a reasonable method to start, but I can do that.

karlp:
#1021 does look good, I definitely felt some of those were laregely copy/pasted

docs/backends/bgapi.rst
karlp: You'd still expect the user to provide the backend explicitly though right? I could modify the selector right now to auto select the backend if an enviironmen variable pointing to the bgapi file exists? is that what you mean?

dlech: Yes, that way we can run all of the examples without having to modify the code.

karlp:
That seems perfectly reasonable.  The corporate backing behind this has dried up, and while I've personally got all the parts and still think it's worthwhile, I'm not sure how much time I would have going forwards with it.  I've flagged it locally though to make sure I push my last changes here, I did make follow ups to make it even better that don't appear to have made it up yet!

karlp:
Ok, this is the latest and greatest.  It was overhauled to use a singelton around the serial port, to allow a single python application to have multiple concurrent clients and scanners, and various field discovered bugs and quirks of the silabs stack squashed.  The latter commits could perhaps be cleaned/squished more, which is why this wasn't pushed up earlier.

PR #1201, opened 2023-01-15
docs: Change code example for BleakClient.start_notify callback

x11x:
The type of the first argument to the callback changed from `int` to `BleakGATTCharacteristic` in 0.18.0. The code example was still type hinting it as an int.

I have added an import statement for `BleakGATTCharacteristic`, to make the code runnable, but I'm not sure if it should be omitted for clarity/readability/minimize confusion (since type hinting is optional). Feel free to delete that line if you prefer


dlech:
Thanks!

PR #1229, opened 2023-02-18
github: upgrade pip before poetry install

dlech:
This fixes failures due to invalid wheels on macos with older versions of Python.


dlech:
Doesn't fix the problem.

Issue #1258, opened 2023-03-18
The disconnect monitor task can get garbage collected before it finishes


bdraco:
https://github.com/hbldh/bleak/blob/1b93859a8b5cd4ef886c7e4084977a7c973a92f4/bleak/backends/bluezdbus/client.py#L246

nothing is holding a reference to the task and asyncio tasks are tracked in weakset which means the task can get garbage collected before it finishes 

https://docs.python.org/3/library/asyncio-task.html#id3

From the asyncio tasks section:
> Important Save a reference to the result of this function, to avoid a task disappearing mid-execution. The event loop only keeps weak references to tasks. A task that isn’t referenced elsewhere may get garbage collected at any time, even before it’s done. For reliable “fire-and-forget” background tasks, gather them in a collection:

bdraco:
Thank you

PR #1259, opened 2023-03-18
bleak: fix leaking of ensure_future()

dlech:
As noted in the Python docs, a reference to the return value of `asyncio.ensure_future()` must be held in order to prevent it from being garbage collected before the task completes.

This applies the recommended fix from the docs of holding the reference in a global set and then discarding the reference when the task completes.

Also change `asyncio.ensure_future()` to `asyncio.create_task()` while we are touching this since the minimum supported Python version is now 3.7.

Fixes: https://github.com/hbldh/bleak/issues/1258

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,12 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 `Unreleased`_
 =============
 
+Fixed
+-----
+* Fixed possible garbage collection of running async callback from ``BleakClient.start_notify()``.
+* Fixed possible garbage collection of running async callback from ``BleakScanner(detection_callback=)``.
+* Fixed possible garbage collection of disconnect monitor in BlueZ backend. Fixed #1258.
+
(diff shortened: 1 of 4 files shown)


dlech:
cc: @bdraco

bdraco:
It might be safer to bind it to the scanner or client object as `self._background_tasks` so if the task never finishes for some reason and the object is destroyed the task will get garbaged collected eventually.

If we are sure the task will always eventually finish and release its references to the object than it probably doesn't matter.

dlech:
> That will only work if its ok for the task to get GCed when the scanner/client goes out of scope (I think it is?)

I don't think we ever want to have a task get GCed before completing, otherwise it could just stop in the middle of a coroutine without ever cancelling is which could lead to undefined/unexpected behavior. 

This type of fire-and-forget programming is a bit of an anti-pattern IMHO, so if people want to use it, I'm inclined to let them work out the details of how to cancel the task to avoid leaks if needed.

bdraco (approved):

Issue #1330, opened 2023-06-09
Scanner periodically crashes on Windows


vmacari:
* bleak version:
    * bleak==0.20.2       
     * bleak-winrt==1.2.0 
* Python version: Python 3.11.3 
* Operating System: Windows 10
* BlueZ version (`bluetoothctl -v`) in case of Linux: ? 

### Description
    Scanning for devices

### What I Did

`self.devices = await BleakScanner.discover()`

### Logs
```

handle: <Handle BleakScannerWinRT._stopped_handler(<_bleak_winrt...002092A07A7F0>, <_bleak_winrt...002092A0EA710>)>
Traceback (most recent call last):
  File "...asyncio\events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
  File "...\site-packages\bleak\backends\winrt\scanner.py", line 213, in _stopped_handler
    len(self.seen_devices), self.watcher.status
                            ^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'status'
```

vmacari:
Added a pull request https://github.com/hbldh/bleak/pull/1331

PR #1331, opened 2023-06-09
Fix windows scanner crash

vmacari:
https://github.com/hbldh/bleak/issues/1330

### DIFF
--- a/bleak/backends/winrt/scanner.py
+++ b/bleak/backends/winrt/scanner.py
@@ -210,7 +210,7 @@ class BleakScannerWinRT(BaseBleakScanner):
     def _stopped_handler(self, sender, e):
         logger.debug(
             "{0} devices found. Watcher status: {1}.".format(
-                len(self.seen_devices), "?" if self.watcher is None else self.watcher.status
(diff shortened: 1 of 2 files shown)


bleak/backends/winrt/scanner.py
```diff
@@ -210,7 +210,7 @@ def _received_handler(
     def _stopped_handler(self, sender, e):
         logger.debug(
             "{0} devices found. Watcher status: {1}.".format(
-                len(self.seen_devices), self.watcher.status
+                len(self.seen_devices), "?" if self.watcher is None else self.watcher.status
```
dlech, line 213: How about using `sender` instead?

```suggestion
                len(self.seen_devices), sender.status
```

dlech:
Can you also add a changelog entry?

bleak/backends/winrt/scanner.py
vmacari: Sure, that's an option too!

vmacari (commented):
Good point, thank you!

PR #1361, opened 2023-07-07
BleakScanner: Add async iterator scanning capability

bojanpotocnik:
> Enable using `BleakScanner()` as an async iterator (async for), combining detection_callback and filtering functionality. `BleakScanner().devices()` method has the same functionality with option to provide timeout.

I noticed that almost every time I use `bleak`, one of the first things is to create scanner method which combines `discover()` and `find_device_by_filter()` functionality "in concurrent fashion", without the need of using callbacks and queue. This enables reporting scanned devices immediately while giving all the filtering flexibility, even when scanning for multiple devices with different filters.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -22,7 +22,7 @@ Changed
 * Scanner backends modified to allow multiple advertisement callbacks. Merged #1367.
 * Changed default handling of the ``response`` argument in ``BleakClient.write_gatt_char``.
   Fixes #909.
-* Added ``advertisement_data()`` async iterator method to ``BleakScanner``.
+* Added ``advertisement_data()`` async iterator method to ``BleakScanner``. Merged #1361.
 * Added ``scan_iterator.py`` example.
 
 Fixed
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -22,6 +22,8 @@ Changed
 * Scanner backends modified to allow multiple advertisement callbacks. Merged #1367.
 * Changed default handling of the ``response`` argument in ``BleakClient.write_gatt_char``.
   Fixes #909.
+* Added ``advertisement_data()`` async iterator method to ``BleakScanner``.
+* Added ``scan_iterator.py`` example.
 
 Fixed
 -----
--- a/bleak/__init__.py
+++ b/bleak/__init__.py
@@ -227,7 +227,7 @@ class BleakScanner:
         Returns:
             An async iterator that yields tuples (:class:`BLEDevice`, :class:`AdvertisementData`).
 
-        .. versionadded:: 0.20.1
+        .. versionadded:: 0.21
         """
         devices = asyncio.Queue()
 
--- a/bleak/__init__.py
+++ b/bleak/__init__.py
@@ -16,6 +16,7 @@ import sys
 import uuid
 from typing import (
     TYPE_CHECKING,
+    AsyncGenerator,
     Awaitable,
     Callable,
     Dict,
@@ -214,6 +215,31 @@ class BleakScanner:
         )
         self._backend.set_scanning_filter(**kwargs)
 
+    async def advertisement_data(
+        self,
+    ) -> AsyncGenerator[Tuple[BLEDevice, AdvertisementData], None]:
+        """
+        Yields devices and associated advertising data packets as they are discovered.
+
+        .. note::
+            Ensure that scanning is started before calling this method.
+
+        Returns:
+            An async iterator that yields tuples (:class:`BLEDevice`, :class:`AdvertisementData`).
+
+        .. versionadded:: 0.20.1
+        """
+        devices = asyncio.Queue()
+
+        unregister_callback = self._backend.register_detection_callback(
+            lambda bd, ad: devices.put_nowait((bd, ad))
+        )
+        try:
+            while True:
+                yield await devices.get()
+        finally:
+            unregister_callback()
+
     @overload
     @classmethod
     async def discover(
@@ -370,16 +396,12 @@ class BleakScanner:
             the timeout.
 
         """
-        found_device_queue: asyncio.Queue[BLEDevice] = asyncio.Queue()
-
-        def apply_filter(d: BLEDevice, ad: AdvertisementData):
-            if filterfunc(d, ad):
-                found_device_queue.put_nowait(d)
-
-        async with cls(detection_callback=apply_filter, **kwargs):
+        async with cls(**kwargs) as scanner:
             try:
                 async with async_timeout(timeout):
-                    return await found_device_queue.get()
+                    async for bd, ad in scanner.advertisement_data():
+                        if filterfunc(bd, ad):
+                            return bd
             except asyncio.TimeoutError:
                 return None
 
--- a/docs/api/scanner.rst
+++ b/docs/api/scanner.rst
@@ -62,12 +62,18 @@ following methods:
 Getting discovered devices and advertisement data
 -------------------------------------------------
 
-If you aren't using the "easy" class methods, there are two ways to get the
+If you aren't using the "easy" class methods, there are three ways to get the
 discovered devices and advertisement data.
 
 For event-driven programming, you can provide a ``detection_callback`` callback
 to the :class:`BleakScanner` constructor. This will be called back each time
 and advertisement is received.
+
+Alternatively, you can utilize the asynchronous iterator to iterate over
+advertisements as they are received. The method below returns an async iterator
+that yields the same tuples as otherwise provided to ``detection_callback``.
+
+.. automethod:: bleak.BleakScanner.advertisement_data
 
 Otherwise, you can use one of the properties below after scanning has stopped.
 
--- a/examples/scan_iterator.py
+++ b/examples/scan_iterator.py
@@ -1,0 +1,37 @@
+"""
+Scan/Discovery Async Iterator
+--------------
+
+Example showing how to scan for BLE devices using async iterator instead of callback function
+
+Created on 2023-07-07 by bojanpotocnik <info@bojanpotocnik.com>
+
+"""
+import asyncio
+
+from bleak import BleakScanner
+
+
+async def main():
+    async with BleakScanner() as scanner:
+        print("Scanning...")
+
+        n = 5
+        print(f"\n{n} advertisement packets:")
+        async for bd, ad in scanner.advertisement_data():
+            print(f" {n}. {bd!r} with {ad!r}")
+            n -= 1
+            if n == 0:
+                break
+
+        n = 10
+        print(f"\nFind device with name longer than {n} characters...")
+        async for bd, ad in scanner.advertisement_data():
+            found = len(bd.name or "") > n or len(ad.local_name or "") > n
+            print(f" Found{' it' if found else ''} {bd!r} with {ad!r}")
+            if found:
+                break
+
+
+if __name__ == "__main__":
+    asyncio.run(main())


bleak/__init__.py
```diff
@@ -204,6 +206,53 @@ def set_scanning_filter(self, **kwargs):
         )
         self._backend.set_scanning_filter(**kwargs)
 
+    def __aiter__(self) -> AsyncIterator[Tuple[BLEDevice, AdvertisementData]]:
+        self._backend.new_devices = asyncio.Queue()
+        return self
+
+    async def __anext__(self) -> Tuple[BLEDevice, AdvertisementData]:
+        return await self._backend.new_devices.get()
```
bojanpotocnik, line 214: Not sure if implementing this is even required... It saved 10 chars of typing (`.devices()`), but cannot have the timeout or return type functionality. Additionally, `async for .. in scanner:` might be less intuitive than `async for .. in scanner.devices():`

bleak/__init__.py
```diff
@@ -204,6 +206,55 @@ def set_scanning_filter(self, **kwargs):
...
+    def devices(
+        self, timeout: Optional[float] = None, return_adv: Literal[False] = False
+    ) -> AsyncIterator[BLEDevice]:
+        ...
+
+    @overload
+    def devices(
+        self, timeout: Optional[float] = None, return_adv: Literal[True] = True
+    ) -> AsyncIterator[Tuple[BLEDevice, AdvertisementData]]:
+        ...
```
bojanpotocnik, line 226: Omitting `async` in typing overloads is deliberate (lost some time on this :facepalm:), see https://stackoverflow.com/questions/68905848/how-to-correctly-specify-type-hints-with-asyncgenerator-and-asynccontextmanager/68911014#68911014 and https://github.com/python/mypy/issues/5385#issuecomment-407281656

bleak/__init__.py
dlech: I would rather not make the `BleakScanner` class itself iterable.

bleak/__init__.py
```diff
@@ -204,6 +206,55 @@ def set_scanning_filter(self, **kwargs):
...
+
+    async def devices(self, timeout: Optional[float] = None, return_adv: bool = False):
+        """
+        Continuously scan for devices and yield them.
```
dlech, line 230:
```suggestion
        Continuously scans for advertisement data and yields them.
```

bleak/__init__.py
```diff
@@ -204,6 +206,55 @@ def set_scanning_filter(self, **kwargs):
...
+
+        Args:
+            timeout:
```
dlech, line 233: I would rather leave out the timeout argument and let users wrap with `with asyncio.timeout(...):` if needed.

bleak/backends/scanner.py
```diff
@@ -141,6 +141,9 @@ def __init__(
         )
 
         self.seen_devices = {}
+        self.new_devices: Optional[
```
dlech, line 144: I would much rather not add new state here as managing state like this has been the main cause of bugs in Bleak.

bleak/__init__.py
```diff
@@ -204,6 +206,55 @@ def set_scanning_filter(self, **kwargs):
...
+                The maximum duration, in seconds, to scan for each individual device.
+                This is not a total scan duration. Each invocation of ``next()`` will
+                wait for up to ``timeout`` seconds for a new device to be discovered
+                before stopping the iteration.
+            return_adv:
```
dlech, line 238: I don't think we should make `return_adv` an option, it should always be enabled. The only reason we have it on other methods is for backwards compatibility. Since this is a new method, we don't need it.

bleak/__init__.py
```diff
@@ -204,6 +206,55 @@ def set_scanning_filter(self, **kwargs):
...
+    ) -> AsyncIterator[BLEDevice]:
+        ...
+
+    @overload
+    def devices(
+        self, timeout: Optional[float] = None, return_adv: Literal[True] = True
+    ) -> AsyncIterator[Tuple[BLEDevice, AdvertisementData]]:
+        ...
+
+    async def devices(self, timeout: Optional[float] = None, return_adv: bool = False):
```
dlech, line 228: I think this should be a static method similar to `discover()` (I like how `find_device_by_filter()` can now be implemented in terms of this iterator.).

dlech: There are some filters that can actually be sent to the Bluetooth hardware, so I think we need to include options like `service_uuids: List[str]` here. Also `scanning_mode` and the various backend-specific args.

This can be implemented as `**kwargs` and just pass everything to the `BleakScanner` constructor and have the documentation refer to that.

bleak/__init__.py
bojanpotocnik: Done, methods removed.

CHANGELOG.rst
```diff
@@ -13,11 +13,12 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 Changed
 -------
 - Improved error messages when failing to get services in WinRT backend.
+- Made ``BleakScanner`` an async iterator, added ``BleakScanner.devices()`` async iterator method.
 
 Fixed
 -----
 - Fix handling all access denied errors when enumerating characteristics on Windows. Fixes #1291.
-- Added support for 32bit UUIDs. Fixes #1314 
+- Added support for 32bit UUIDs. Fixes #1314
```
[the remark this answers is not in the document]
bojanpotocnik: This is a result of PyCharm automatically removing trailing spaces, but also a good catch – fixed.

bleak/__init__.py
bojanpotocnik: This was my first implementation, didn't know that this is for backwards compatibility - but I am glad that this is the case :+1:

bleak/backends/scanner.py
bojanpotocnik: This was done as simplest/most-effective workaround to avoid https://github.com/hbldh/bleak/pull/1361/files#diff-e4f68248febf58b848a3fd5bbdce9a669c236349f58b7d9005b704fd507ab8b6R145-R149 plus https://github.com/hbldh/bleak/pull/1361/files#diff-a1877ebc56efdeaa2019b7e6738a9eba73a8788f646396e9a6221ac36f2d9f83R218-R221.
I had 2 implementations before this: the one implemented now and "the most proper one" from the user standpoint, which saved user callback passed to `BleakScanner(detection_callback)` to internal variable, registered own `BleakScanner._detection_calback()` in `self._backend = PlatformBleakScanner(self._detection_calback, ...)` and then called the user callback, if provided, from this internal callback after adding the device to the queue. However, this meant that I fully replicated the `BaseBleakScanner`'s `detection_callback` logic (https://github.com/hbldh/bleak/pull/1361/files#diff-e4f68248febf58b848a3fd5bbdce9a669c236349f58b7d9005b704fd507ab8b6R150-R184), which I did not like.

The solution now bases on the fact that: if a user is using the scanner as async iterator, it shall not require the `detection_callback` functionality, as coroutines are meant to replace this ("callback hell") exactly.

bleak/__init__.py
bojanpotocnik: I thought about this before opening a PR. I think there are two ways to put it:

1. Have class methods which instantiate class instance and use it for scanning:
```python
# Pseudo code
    @classmethod
    async def discover(cls, **kwargs) -> BLEDevice:
        with cls(**kwargs) as scanner:
            for device in scanner.scan():
                yield device

async for device in BleakScanner.discover(my_filters):
    print(device)
```

2. Have instance methods and instantiate the scanner first
```python
# Pseudo code
    async def discover(self, **kwargs) -> BLEDevice:
        for device in self.scan():
            yield device

async for device in BleakScanner(my_filters).discover():
    print(device)
```

I know we cannot change from method 1 for compatibility. However, now we need a copy of docstring and typing for `**kwargs` and other scanning parameters at `BleakScan`'s `__init__()`, `set_scanning_filter()` (soon not anymore), `discover()`, `find_device_by_address()`, `find_device_by_name()`, `find_device_by_filter()`, and `scan()` or `watch_advertisements()`.

In method 2, we only require them at one place, the `__init__()` (and the scanner instance can even be reused for some extra use case).
And because I "hate" having just arbitrary `**kwargs` instead of all properly typed and documented, I made the main scan iterator method instance method, delegating the task of properly documenting all parameters to the `__init__()` method. On top of this, I would call `__init__()` in the static method anyway, just that I would need to pass all arguments.

I have now, as example, for review implemented both.
I kept the original method named `devices()` and added a `note` in the docstring - this method does not actually do anything, it only returns the devices detected - almost exactly like the property `discovered_devices`, just in the real-time and in an async way.
The class method version is called `scan()`, because it actually instantiates the class, starts scanning, and then uses `devices()` to return the devices (method 1 vs 2 again). For `kwargs` I just copied the docstring from `find_device_by_filter()`, because they mean the same.

This is just a reference update and my thought, comments welcome.

bleak/__init__.py
```diff
@@ -204,6 +205,51 @@ def set_scanning_filter(self, **kwargs):
...
+
+        async def detection_callback(bd, ad):
+            await devices.put((bd, ad))
+
+        self._backend.register_detection_callback(detection_callback)
+
+        while True:
+            try:
+                yield await devices.get()
+            except GeneratorExit:
```
bojanpotocnik, line 233: Fix: I will change `except GeneratorExit:` to `finally:` on next force push

examples/scan_iterator.py
```diff
@@ -0,0 +1,38 @@
...
+    x = 5
+    print(f"Scanning for {x} devices...")
+    async for bd, ad in BleakScanner.scan():
+        print(f"{x}. {bd!r} with {ad!r}")
+        x -= 1
+        if x == 0:
+            break
+
+    # Ensure that asyncio loop has time to stop the scanner via the context manager exit
+    await asyncio.sleep(0.1)
```
dlech, line 25: Ugh. I forgot about the issue with unfinished generators hanging around without cleaning up. So it is probably best to just avoid the issue and leave out the `scan()` method.

dlech: The alternative is to always write like this instead:

```python
    print(f"Scanning for {x} devices...")
    dev_gen = BleakScanner.scan()

    try:
        async for bd, ad in dev_gen:
            print(f"{x}. {bd!r} with {ad!r}")
            x -= 1
            if x == 0:
                break
    finally:
        # Ensure that the scanner is stopped
        await dev_gen.aclose()
```

bleak/__init__.py
dlech: Technically, this is going to suffer the same leak when `devices()` is not iterated to completion as mentioned in my other comments. However, in this case it doesn't seem like the consequences are that bad since we usually leave the callback hanging anyway. The only problem would be if someone called `devices()` a second time without waiting causing it to raise the `BleakError` because of the lingering callback, but this can be worked around.

I suppose we could also consider just leaving out the `try/finally` and the error check.

bleak/__init__.py
```diff
@@ -148,12 +149,12 @@ def __init__(
...
-    async def __aenter__(self):
-        await self._backend.start()
+    async def __aenter__(self) -> BleakScanner:
+        await self.start()
         return self
 
-    async def __aexit__(self, exc_type, exc_val, exc_tb):
-        await self._backend.stop()
+    async def __aexit__(self, exc_type, exc_val, exc_tb) -> None:
+        await self.stop()
```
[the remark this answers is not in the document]
bojanpotocnik: Change remove. Should I open a PR or do you have any pending changes/commits that this could fit into?

examples/scan_iterator.py
bojanpotocnik: I understand – thank you. This then diminishes the whole point of having a class method, as it is even more work than using instance method :smile:

I now solved it by changing the order of examples – this solves the issue. I can still remove the `scan()` method if desired (would also fit my long comment above https://github.com/hbldh/bleak/pull/1361#discussion_r1258890610 )

bleak/__init__.py
bojanpotocnik: > Fix: I will change `except GeneratorExit:` to `finally:` on next force push

Note to self: just changing it raises https://rules.sonarsource.com/python/RSPEC-1143/ . However, I solved it by moving the `while True`.

> ...  this is going to suffer the same leak when devices() is not iterated to completion ...

As for the callback, I would argue that if this method registers it, it shall also unregister it for sure. Lingering callback would throw error anyway because `devices` queue may/would not be allocated anymore.

For the queue, I have thought about the queue cleaning before adding it (even in the first implementation), but somewhere found something in lines of
> When you set a variable to None, the reference to the object is lost. Python's garbage collector will eventually deallocate the object when there are no more references to it - including items left in the queue, assuming there are no other references to these objects.

However, as I cannot find this again, neither any actual source of it, I have added explicit cleanup. Still not sure whether the explicit cleanup is required or `devices = None` would do :thinking: 
Also note the added comment to commit message about why the queue is not cleaned using 
```python
try:
    while True:
        devices.get_nowait()
except asyncio.QueueEmpty:
    pass
```

bleak/__init__.py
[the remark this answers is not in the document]
dlech: Yes, feel free to make a separate PR for it. Always happy to keep improving typing. Just doing small fixes like this as we notice them is easier to review rather than trying to tackle a whole bunch at one time.

bleak/backends/scanner.py
```diff
@@ -142,6 +142,11 @@ def __init__(
 
         self.seen_devices = {}
 
+    @property
+    def detection_callback(self) -> Optional[AdvertisementDataCallback]:
+        """Get the callback provided to :meth:`__init__` or registered with :meth:`register_detection_callback`"""
+        return self._callback
+
```
dlech, line 149: Instead of adding this property and doing the check for existing callback in the new `devices()` method, could we just add the check for `if self._callback` in `register_detection_callback()`?

examples/scan_iterator.py
dlech: > This then diminishes the whole point of having a class method

I very much agree and think we should leave it out.

> I now solved it by changing the order of examples

I think this just hides the problem rather than solving it.

bleak/__init__.py
```diff
@@ -204,6 +205,52 @@ def set_scanning_filter(self, **kwargs):
         )
         self._backend.set_scanning_filter(**kwargs)
 
+    async def devices(self) -> AsyncIterator[Tuple[BLEDevice, AdvertisementData]]:
+        """
+        Yields devices and their advertising data as they are discovered.
+
+        .. note::
+            Ensure that scanning is started before calling this method.
```
dlech, line 213: I think it is also important to note that there must not be any `await` between starting scanning and calling `devices()` otherwise advertisements can be missed. (This pitfall was my main reason for preferring the proposed `scan()` over `devices()` until I realized the non-deterministic cleanup issues with `scan()`).

It would be very helpful to include an example snippet here in the documentation that shows proper usage.

dlech:
We also need to expand the "Getting discovered devices and advertisement data" section in `scanner.rst` so that the new method is included in the documentation.

examples/scan_iterator.py
dlech: For the future archeologists, here is perhaps a better way to do a classmethod async iterator with deterministic cleanup (but still not very nice).

```python
    print(f"Scanning for {x} devices...")
    
    async with contextlib.aclosing(BleakScanner.scan()) as dev_gen:
        async for bd, ad in dev_gen:
            print(f"{x}. {bd!r} with {ad!r}")
            x -= 1
            if x == 0:
                break
```

bleak/backends/scanner.py
bojanpotocnik: `BaseBleakScanner::register_detection_callback()` is not aware whether it is being called with user callback from `BleakScanner::__init__()` or `BleakScanner::devices()`, so it cannot perform any checks. Also the `register_detection_callback()`'s docstring says
> If another callback has already been registered, it will be replaced with ``callback``. ``None`` can be used to remove the current callback.

I cannot just overwrite the callback the user provided in `BleakScanner::__init__()` without their knowledge. So some check needs to be somewhere.
This is why the method of adding `queue` to `BaseBleakScanner` - in this case, any user callback and async iterator operation can work simultaneously, not exlusivelly.

bleak/__init__.py
```diff
@@ -204,6 +205,52 @@ def set_scanning_filter(self, **kwargs):
...
+            await devices.put((bd, ad))
+
+        self._backend.register_detection_callback(detection_callback)
+        try:
+            while True:
+                yield await devices.get()
+        finally:
+            self._backend.register_detection_callback(None)
+            while not devices.empty():
+                await devices.get()
```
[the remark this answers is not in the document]
dlech: correct

bleak/__init__.py
bojanpotocnik: Actually, I was aware of this, and it was kinda deliberate. I create the queue when the users starts iterating and destroy it afterward.
If I create the queue in the `__init__()` and add devices to it... What if the user never uses the async iterator, what if they just start the scanning and do nothing? When the queue fills up, should I remove the first found device and add the new one, or just stop adding new devices?

I actually don't see any drawback in skipping few advertisements... After all, what if the user started the scan a few moments later?

As I am mostly using the `BleakScanner()` as an instance instead of using the class methods, I wanted to allow user the flexibility: create the scanner; use the callback for a while, if you want; remove the callback; use it as an iterator; etc.. Unusual use case, but why not.

Now when you put it like that, I am thinking... But maybe the answer lies in the docstring themselves:
- `discovered_devices()`: Gets list of the devices that the scanner **has discovered during the scanning**.
I agree that this shall return all devices since `.start()`, and it does.
- `devices()`: Yields devices and their advertising data **as they are discovered**.
This just yields any device discovered [only] while you are using this method (iterating it).

Maybe I should really rename the `devices()` method. You suggested `watch_advertisements`, which seems kinda long, and everything else is based around the notion of _devices_. I would like to avoid `scan` (for reasons explained at the end of https://github.com/hbldh/bleak/pull/1361#discussion_r1258890610 ). `observe_devices` seems like we are observing their state or something...

bleak/backends/scanner.py
dlech: Thanks for you thoughts on this. It has given me an idea to perhaps change things in the backends to allow multiple callbacks to be registered at the same time. We don't have to do this as part of this pull request. So for now, how about we either silently overwrite the callback or just add a check for `if self._backend._callback` instead of adding this property?

bleak/__init__.py
dlech: > I actually don't see any drawback in skipping few advertisements... After all, what if the user started the scan a few moments later?

Agreed.

> Maybe I should really rename the `devices()` method.

I like having `advertisements` in the name since each iteration is triggered by new advertisement data rather than a new device. I think `observe` is a good name since it fits with BLE broadcast/observe terminology.

bojanpotocnik: Ok, I actually asked global buzzword for a help

> The distinction between discovered_devices() and devices() is indeed a bit subtle given their current names. Here are a few alternative names for the devices() method that might make this distinction clearer:
>- `live_devices()`: This suggests that the method provides a real-time feed of devices as they are discovered.
>- `streaming_devices()`: This also suggests a continuous feed of devices, and aligns with terms like "streaming data".
>- `dynamic_devices()`: This implies that the list of devices can change over time, which is the case for this method.
>- `discover_live_devices()`: This combines elements of a couple of the above suggestions.
>
> Remember, the key is to find a name that makes it clear this method is different from `discovered_devices()` in that it provides a live, ongoing feed of discovered devices.

> More suggestions please. You can also leave out `devices` word if you come up with any good alternative.

> Sure, here are a few more suggestions, dropping the devices word for some alternatives:
>- `monitor()`: This implies that the method is actively monitoring for new devices.
>- `discover_live()`: This suggests that the method is discovering devices in real-time.
>- `detect()`: This could imply a continuous process of detection.
>- `observe()`: This also suggests an ongoing, active process.
>- `watch()`: This is short, and implies a continuous process.
>- `continuous_discovery()`: A bit longer, but very descriptive.
>- `stream_discoveries()`: Suggests that this is an ongoing, streaming process.
>- `find_as_available()`: Implies that devices are yielded as soon as they're found.

I still can't decide. I kiiind of like `monitor` or `observe`, but the most fitting names are already taken :smile:

bleak/__init__.py
```diff
@@ -204,6 +205,37 @@ def set_scanning_filter(self, **kwargs):
...
+        self._backend.register_detection_callback(detection_callback)
+        try:
+            while True:
+                yield await devices.get()
+        finally:
+            # noinspection PyProtectedMember
+            if self._backend._callback not in (user_cb, detection_callback):
+                # This check can be removed when deprecated `register_detection_callback()` method is removed
+                raise BleakError("Detection callback has been changed while scanning")
+            self._backend.register_detection_callback(user_cb)
```
bojanpotocnik, line 237: This showcases why I did https://github.com/hbldh/bleak/pull/1366 - to avoid replicating the callback conversion logic as mentioned in https://github.com/hbldh/bleak/pull/1361#discussion_r1258874945 but still allow both functionalities to coexist (also the starting word https://github.com/hbldh/bleak/pull/1361/files#diff-48384236075c00b8fc59ec560483d0b5692082ebe327f27ac5f6aed55d73835eR72 is "_Additionally, ..._").

However, I don't really like this solution, and I can't really find a use case for using both methods concurrently. And yes, the check is done to handle if the user called `scanner.register_detection_callback()` while iterating over `scanner.observe()`, which they can.

bleak/__init__.py
bojanpotocnik: Renamed to `observe()`

bleak/__init__.py
bojanpotocnik: Explicit queue cleanup removed, as per https://github.com/hbldh/bleak/pull/1361#discussion_r1263876425

examples/scan_iterator.py
bojanpotocnik: This is added in the example https://github.com/hbldh/bleak/pull/1361/files#diff-547d9f54fbc8eafa2aa817c1d50d8e7cfd3958c4f5983d0923bf79a99b255222R37
I hope it is well enough separated to not confuse regular users while providing useful insight in async resource cleanup.

I am afraid nobody will find this comments here, or at least will took them too much time.

bleak/__init__.py
bojanpotocnik: Deleted class method as per https://github.com/hbldh/bleak/pull/1361#discussion_r1262777118

bleak/backends/scanner.py
bojanpotocnik: Done! In the meantime, I also pushed one example where both methods could be used simultaneously  https://github.com/hbldh/bleak/pull/1361#discussion_r1267361215

examples/scan_iterator.py
```diff
@@ -0,0 +1,106 @@
...
+            if not run_without_explicit_cleanup:
+                await observer.aclose()
+
+            observer = scanner.observe()
+            await iterate_devices(observer, 3)
+            if not run_without_explicit_cleanup:
+                await observer.aclose()
+
+        await iterate_devices(scanner.observe(), 6)
+        # This one can be cleaned implicitly, as we do not care about reusing the same scanner anymore
```
bojanpotocnik, line 95: Looking at this now, `run_without_explicit_cleanup` makes this example too messy. I will replace it with a simple comment and let the users experiment by themselves.

bojanpotocnik:
The last two comments are here so you can see the alternative implementations - I can add any of them back if desired.

bleak/__init__.py
dlech: I feel like since `observe()` is a verb, it makes it sound like this is starting scanning. How about `observed_changes()` instead? I haven't really encountered any many async generators yet, so I don't know if there is a common naming pattern of what feels the most Pythonic (the ones I have used were binding a C++ library, so they don't really count).

```python
async with BleakScanner() as scanner:
    async for dev, ad in scanner.observed_changes():
```

Another reason for this name is that after thinking about it some more this is more descriptive of how the OS Bluetooth stacks actually work. We don't necessarily receive an event for every single advertisement packet, but rather receive an event when something changes, either the RSSI or the value of one of the data. It would be helpful to include a variation of the previous sentence in the docstring.

bleak/__init__.py
```diff
@@ -204,6 +205,33 @@ def set_scanning_filter(self, **kwargs):
...
+            )
+
+        devices = asyncio.Queue()
+
+        self._backend.register_detection_callback(
+            lambda bd, ad: devices.put_nowait((bd, ad))
+        )
+        try:
+            while True:
+                yield await devices.get()
```
dlech, line 231: We should add a `TODO` comment here that if/when https://github.com/hbldh/bleak/pull/1146 gets merged, we need to exit the generator here instead of waiting forever.

examples/scan_iterator.py
```diff
@@ -0,0 +1,103 @@
...
+    print(f"\nScanning for a device with name longer than {n} characters...")
+    async with BleakScanner() as scanner:
+        async for bd, ad in scanner.observe():
+            found = len(bd.name or "") > n or len(ad.local_name or "") > n
+            print(f"Found{' it' if found else ''} {bd!r} with {ad!r}")
+            if found:
+                break
+
+
+async def deterministic_cleanup():
```
dlech, line 37: I would like to merge https://github.com/hbldh/bleak/pull/1367 first, then I think we can avoid the need for deterministic cleanup.

docs/api/scanner.rst
```diff
@@ -69,6 +69,12 @@ For event-driven programming, you can provide a ``detection_callback`` callback
 to the :class:`BleakScanner` constructor. This will be called back each time
 and advertisement is received.
 
+Alternatively, you can utilize the asynchronous iterator to iterate over
```
dlech, line 72: A few lines above, it says there are "two ways" , but now there are "three".

bleak/__init__.py
dlech: And actually we probably need something now, to handle the case, e.g. if someone manually stops the scanner while we are still iterating.

bleak/__init__.py
```diff
@@ -204,6 +205,33 @@ def set_scanning_filter(self, **kwargs):
...
+    async def observe(self) -> AsyncIterator[Tuple[BLEDevice, AdvertisementData]]:
+        """
+        Yields devices and their advertising data as they are discovered.
+
+        .. note::
+            Ensure that scanning is started before calling this method.
+
+        Returns:
+            An async iterator that yields tuples (:class:`BLEDevice`, :class:`AdvertisementData`).
+        """
```
dlech, line 217:
```suggestion

        .. versionadded:: 0.20.1
        """
```

examples/scan_iterator.py
bojanpotocnik: I will wait for https://github.com/hbldh/bleak/pull/1367 to merge - then we also avoid dilemmas like https://github.com/hbldh/bleak/pull/1361#discussion_r1267361215

bojanpotocnik: Removed that code from the example. I could not read all the discussions here in the past half year, but if there are users hitting this problem in some other use cases, this info about proper cleanup can be added to some other (or separate) example.

bleak/__init__.py
bojanpotocnik: This is solved by #1367, see https://github.com/hbldh/bleak/pull/1361#discussion_r1268395747

bleak/__init__.py
bojanpotocnik: > We don't necessarily receive an event for every single advertisement packet, but rather receive an event when something changes

Sometimes getting each advertisement packet is a desired behavior, but hard to achieve on certain OS (e.g. my tries with passive scanning and `AdvertisementMonitor` in the past). I would say that one wants to either _find a BLE device_, or _scan advertisement packets_ – filtering later to only get packets where something changes shall be an option/feature, not a limitation (however this is domain of OS BT stack, not `bleak`).

> I haven't really encountered any many async generators yet, so I don't know if there is a common naming pattern of what feels the most Pythonic

I haven't seen or found any naming convention on this topic. Looking at some existing methods, it seems that the name represents what the method _is doing_ internally (verb), if anything, or what it represents (noun):
- `enumerate()`, `zip()`, `map()`, `filter()`, `walk()`, `scandir()`
- `keys()`, `values()`, `items()`, `union()`, `lines()`, `nodes()`, `children()`, `elements()`, `records()`

I still somehow see this method having almost the same functionality as `discovered_devices_and_advertisement_data` property, only that the devices are yielded in real-time, not after the scan, and there are duplicates.

If we ignore all async timeout related overhead in the example below, they are indeed very similar:
```python
    async with BleakScanner() as scanner:
        await asyncio.sleep(1)
    print("Devices:", scanner.discovered_devices_and_advertisement_data)

    # The same, except it prints all advertisement packets instead of devices
    async with BleakScanner() as scanner:
        found = []
        try:
            async with async_timeout(1):
                async for bd_ad in scanner.observe():
                    found.append(bd_ad)
        except asyncio.TimeoutError:
            pass
    print("Adv. packets:", found)
```
With this logic, I could use `discovered_devices_and_advertisement_data` as a starting point:
1. Remove `discovered_`, because this method works in the real time instead of reporting the past.
**\>** `devices_and_advertisement_data`
2. The resulting name is too long, and it exists only for backward compatibility – if we could, we would have just `discovered_devices` which would return the same.
**\>** `devices`
3. `devices` represents the actual `BLEDevice`s, created from, and updated with `AdvertisementData` during scanning. This is not true for this method, as it returns each `AdvertisementData` packet, and `BLEDevice` associated with (created from) it.
**\>** `advertisements`?
**\>** `advertisement_data`?

bleak/__init__.py
bojanpotocnik: I will need to think about how to solve this neatly, if at all possible without adding some `stop` callbacks (`handle_early_stop`). Considering that the code will block on `async for _ in scanner.observe()`, the user cannot stop the scan from the same thread/coroutine, but of course it can be stopped in other ways, or as a result of external events. 

I wanted to add "How do we handle the unexpected stops now, if the user uses the callback?", but then I membered that this is what https://github.com/hbldh/bleak/pull/1146 is all about. Maybe we shall jump to that one first?

bleak/__init__.py
dlech: > 3\. `advertisement_data`

I think this is the best of all the ideas we have come up with so far.

bleak/__init__.py
dlech: I think for now we can leave this the way it is. After thinking about it some more, I'm not sure that stopping the generator when scanning stops is the most intuitive or desirable behavior. For example, I could see someone wanting to just set up this generator and leave it running forever in a long running app and connect the scanner start() and stop() methods to UI button events.

Instead, I'm thinking we might want a more general async context manager for capturing scanner stop events similar to how asyncio.timeout() works. This way users can selectively opt in to having a task interrupted when the scanner stops.

bleak/__init__.py
```diff
@@ -214,6 +215,31 @@ def set_scanning_filter(self, **kwargs):
...
+        Yields devices and associated advertising data packets as they are discovered.
+
+        .. note::
+            Ensure that scanning is started before calling this method.
+
+        Returns:
+            An async iterator that yields tuples (:class:`BLEDevice`, :class:`AdvertisementData`).
+
+        .. versionadded:: 0.20.1
```
bojanpotocnik, line 230: v0.20.2 is already released?

dlech: The next release will be v0.21, so the docs should use that version.

CHANGELOG.rst
```diff
@@ -22,6 +22,8 @@ Changed
 * Scanner backends modified to allow multiple advertisement callbacks. Merged #1367.
 * Changed default handling of the ``response`` argument in ``BleakClient.write_gatt_char``.
   Fixes #909.
+* Added ``advertisement_data()`` async iterator method to ``BleakScanner``.
```
dlech, line 25:
```suggestion
* Added ``advertisement_data()`` async iterator method to ``BleakScanner``. Merged #1361.
```

bleak/__init__.py
dlech, line 230:
```suggestion
        .. versionadded:: 0.21
```

dlech:
Thanks for all of the work on this! It was more complicated than I expected.

PR #1366, opened 2023-07-17
backends/scanner: Fix typing for callback

bojanpotocnik:
Function `register_detection_callback()` of `BaseBleakScanner` ensures that `_callback` is a synchronous function, so change its typing accordingly.

Additionally, passing `Awaitable` to `asyncio.create_task()` results in warning:
>  Unexpected type(s): (Awaitable[None] | None)
>  Possible type(s): (Generator[Any, None, Any] | Coroutine)

Every coroutine is an awaitable, but not every awaitable is a coroutine, so change the `AdvertisementDataFilter` type accordingly.

---

Not sure if this needs an entry in `CHANGELOG.rst` as it does not affect runtime functionality in any way?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,7 +17,8 @@ Changed
 Fixed
 -----
 - Fix handling all access denied errors when enumerating characteristics on Windows. Fixes #1291.
-- Added support for 32bit UUIDs. Fixes #1314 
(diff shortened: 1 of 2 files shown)


dlech:
Looks good, thanks. I think we should add a changelog entry since it is fixing something.

bojanpotocnik:
Done, ready to merge (if no further comments)

PR #1367, opened 2023-07-19
backends: allow multiple advertisement callbacks per scanner

dlech:
We are considering adding an async iterator that can be used to handle advertisement data as it is received. It will be trivial for users to create multiple parallel iterators, so instead of disallowing it, this prepares the way to allow multiple simultaneous subscribers.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,7 +13,7 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 Changed
 -------
 - Improved error messages when failing to get services in WinRT backend.
-- Scanner backends modified to allow multiple advertisement callbacks.
+- Scanner backends modified to allow multiple advertisement callbacks. Merged #1367.
 
 Fixed
 -----
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,7 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 Changed
 -------
 - Improved error messages when failing to get services in WinRT backend.
+- Scanner backends modified to allow multiple advertisement callbacks.
 
 Fixed
 -----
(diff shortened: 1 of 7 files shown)


bleak/backends/scanner.py
```diff
@@ -134,8 +134,15 @@ def __init__(
         service_uuids: Optional[List[str]],
     ):
         super(BaseBleakScanner, self).__init__()
-        self._callback: Optional[Callable[[BLEDevice, AdvertisementData], None]] = None
-        self.register_detection_callback(detection_callback)
+
+        self._ad_callbacks: List[Callable[[BLEDevice, AdvertisementData], None]] = []
+        """
+        List of callbacks to call when an advertisement is received
```
dlech, line 140:
```suggestion
        List of callbacks to call when an advertisement is received.
```

bleak/backends/scanner.py
```diff
@@ -176,7 +185,24 @@ def detection_callback(s, d):
...
+        self._ad_callbacks.append(detection_callback)
+
+        def remove():
+            self._ad_callbacks.remove(detection_callback)
+
+        return remove
+
+    def call_detection_callbacks(
+        self, device: BLEDevice, advertisement_data: AdvertisementData
+    ):
```
dlech, line 197:
```suggestion
    ) -> None:
```

bleak/backends/scanner.py
```diff
@@ -176,7 +185,24 @@ def detection_callback(s, d):
         else:
             detection_callback = callback
 
-        self._callback = detection_callback
+        self._ad_callbacks.append(detection_callback)
```
dlech, line 188: Might want to consider a dict of token: callback instead of list to ensure we don't get unbalanced removals (i.e. it should be safe to call the remove function more than once).

CHANGELOG.rst
```diff
@@ -13,6 +13,7 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 Changed
 -------
 - Improved error messages when failing to get services in WinRT backend.
+- Scanner backends modified to allow multiple advertisement callbacks.
```
dlech, line 16:
```suggestion
- Scanner backends modified to allow multiple advertisement callbacks. Merged #1367.
```

PR #1376, opened 2023-07-25
Defer formatting for expensive debug logging

bdraco:
- Switch usage of f-strings for log message formatting
- Guard expensive loggers with isEnableFor

https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/logging-fstring-interpolation.html

Note: there are more of these. I only fixed the obvious ones and the ones that showed up in the profile (DeviceFound is red hot in the profile)


dlech:
Looks good, thanks!

PR #1329, opened 2023-06-08
Stop waiting for a device to disconnect if its been removed

bdraco:
There can be a long wait when a connection fails if the device is removed from the bus because the connection callback wil never fire as the device gets removed after the `le-connection-abort-by-local`
```
2023-06-08 17:15:51.630 DEBUG (MainThread) [bleak.backends.bluezdbus.client] Connecting to BlueZ path /org/bluez/hci1/dev_78_9C_85_0A_94_40
2023-06-08 17:15:51.720 DEBUG (MainThread) [bleak.backends.bluezdbus.client] retry due to le-connection-abort-by-local
2023-06-08 17:15:51.724 DEBUG (MainThread) [bleak.backends.bluezdbus.manager] received D-Bus signal: org.freedesktop.DBus.ObjectManager.InterfacesRemoved (/): ['/org/bluez/hci1/dev_78_9C_85_0A_94_40', ['org.freedesktop.DBus.Properties', 'org.freedesktop.DBus.Introspectable', 'org.bluez.Device1']]
```
```
66    17:04:21 connect(204, {sa_family=AF_UNIX, sun_path="/var/run/dbus/system_bus_socket"}, 34) = 0
66    17:04:21 ioctl(204, FIONBIO, [1] <unfinished ...>
66    17:04:21 sendto(204, "\0", 1, 0, NULL, 0 <unfinished ...>
66    17:04:21 sendto(204, "AUTH EXTERNAL 30\r\n", 18, 0, NULL, 0 <unfinished ...>
66    17:04:21 recvfrom(204, "OK eeb9e653007b4524f7fe8b6eff03a22f\r\n", 1024, 0, NULL, NULL) = 37
66    17:04:21 sendto(204, "NEGOTIATE_UNIX_FD\r\n", 19, 0, NULL, 0 <unfinished ...>
66    17:04:21 recvfrom(204,  <unfinished ...>
66    17:04:21 sendto(204, "BEGIN\r\n", 7, 0, NULL, 0 <unfinished ...>
66    17:04:21 sendto(204, "l\1\0\1\0\0\0\0\1\0\0\0m\0\0\0\1\1o\0\25\0\0\0/org/freedesktop/DBus\0\0\0\2\1s\0\24\0\0\0org.freedesktop.DBus\0\0\0\0\3\1s\0\5\0\0\0Hello\0\0\0\6\1s\0\24\0\0\0org.freedesktop.DBus\0\0\0\0", 128, 0, NULL, 0) = 128
[...]

PR #1377, opened 2023-07-25
Raise BleakError if the device disappears before trying to resolve services

bdraco:
split from https://github.com/hbldh/bleak/pull/1329


dlech:
Superseded by https://github.com/hbldh/bleak/pull/1399

Issue #1364, opened 2023-07-14
Linux, client.connect() connects then disconnects, using bluetoothctl directly works


librick:
* bleak version: 0.20.2
* Python version: 3.11.4
* Operating System: Linux debian 6.3.0-1-amd64
* BlueZ version (`bluetoothctl -v`) in case of Linux: 5.66

### Description
Calling `client.connect()` given a BleakClient instance shows that the device as connected in BlueZ, but then it immediately disconnects again and Bleak throws a TimeoutError. I can connect (and stay connected) just fine using bluetoothctl.

Manually connecting with bluetoothctl works as expected. I can connect with `connect <my-mac-address>`. Running `devices Connected` several seconds later shows the connected device. The device does not disconnect and all is well. I see these two lines in the bluetoothctl output:
```
[CHG] Device 00:14:41:33:28:8D Connected: yes
[CHG] Device 00:14:41:33:28:8D ServicesResolved: yes
```
Connecting with Bleak does not work. First, inside bluetoothctl, I make sure the device is disconnected: `disconnect 00:14:41:33:28:8D` (it disconnects as expected). Then I run my Python code. Watching the logs inside `bluetoothctl`, I see the following _four lines_;
```
[CHG] Device 00:14:41:33:28:8D Connected: yes
[CHG] Device 00:14:41:33:28:8D ServicesResolved: yes
[CHG] Device 00:14:41:33:28:8D ServicesResolved: no
[CHG] Device 00:14:41:33:28:8D Connected: no
```
And the Python code crashes with a `TimeoutError`.
It's possible I'm just missing something obvious with asyncio, but it seems strange to me that the device would connect and then immediately disconnect, and only when using Bleak.
I've included my full code and exception logs below.

### What I Did
My python code:
```python3
import asyncio
from bleak import BleakScanner, BleakClient

DEVICE_NAME = "Razer Stereo"

# I wrapped BleakScanner because I'm only interested
# in finding a specific device (based on local_name for now);
# once a matching device is discovered, _init() resolves
class BleScanner:
    device = None
    async def _init(self):
        self.stop_event = asyncio.Event()
        cb = lambda x, y: self._callback(x, y) 
        async with BleakScanner(detection_callback=cb) as scanner:
            await self.stop_event.wait()

    def _callback(self, device, advertising_data):
        if advertising_data.local_name != None:
            print("[{}] discovered device, local_name: {}".format(device.address, advertising_data.local_name))
[...]

PR #1382, opened 2023-07-31
bluezdbus client handling of devices that might not provide a response to a successful connection

uutzinger:
Small change from `assert reply is not None` to `if reply is not None` in client.py of bluezdbus.

This fixes issues with devices that do not respond properly to a connection.

Guideline check:
1. It does not add user accessible functionality.
2. Updated 'CHANGELOG.rst' unreleased changed section
3. This affects handling of Linux BlueZ


dlech:
As discussed in https://github.com/hbldh/bleak/issues/1364#issuecomment-1638252883, I would like proof that this is not a BlueZ bug first before considering making such a change.

Also, in the bug report, I didn't see anyone hit an `AssertationError`, so I don't see how this change makes any difference. Do you have a reproducible test case that I could try?

Issue #1184, opened 2022-12-29
App crash at startup


brinata:
if I insert the following line in my Android App:

"import bleak"

the app crashes at start up in my mobile, but it works under windows 10.

On windows I'm using the following versions:

Pythpn 3.10.9
bleak            0.19.5
Kivy               2.1.0
buildozer      1.4.0

Inside the builddozer.spec file I haven't specify the version of the packages

Can Someone Help me ?

dlech:
Without details about the crash and steps to reproduce the problem, it will be difficult for anyone to provide any help.

dlech:
> ModuleNotFoundError: No module named 'async_timeout'

This is a 3rd-party module and likely needs to be added to some depedency list somewhere. Maybe here?

https://github.com/hbldh/bleak/blob/4a6d1c495634f38449180bef113bcdb06a3cf783/bleak/backends/p4android/recipes/bleak/__init__.py#L14

brinata:
Maybe, but the Android Log Stream write that there is the following import error:

 "/home/ronni/app/.buildozer/android/platform/build-arm64-v8a_armeabi-v7a/build/python-installs/myapp/arm64-v8a/bleak/__init__.py", line 31, in <module>
**12-30 11:14:38.496 16697 16734 I python  :  ModuleNotFoundError: No module named 'async_timeout**'


at the line 31 of the file __init__.py , there is the followinfg code:

 sys.version_info < (3, 11):
    from **async_timeout** import timeout as async_timeout
else:
    from asyncio import timeout as async_timeout

it seems that the script recognize a python version older that 3.11 and so it import the timeout library drom the external async_timeout package instead of from the asyncio .

I think that Android doesn't find the "async_timeout" lib to import. On the contrary, on Windows, where the this library is installed as dependecy of the bleak package, the python finds the lib to import.

do I have to declare something special inside my buildozer.spec file ?

dlech:
I don't know much about how buildozer works. The line of code I linked to above was my best guess.

brinata:
I solver the problem adding "async-timeout" in the requirements( the lib name was different from label "async_timeout" used in the import declaration ...)

Now the app starts, but when I try to scan for ble devices, i give back an exception. It desn'pt find the class "com.github.hbldh.bleak.PythonScanCallback



12-30 19:20:36.594  6669  6728 I python  :  Task exception was never retrieved
[...]

PR #1398, opened 2023-08-24
Update of bleak on android and the android kivy example

robgar2001:
Update of bleak on android  and the android kivy example

When trying to build my app with bleak, the android kivy example seemed out of date.

**Commmit: 51b277ee3fce9d7f6fa77579e3371b5fdcaa2045 :**
-  'async-timeout' was missing as a requirement in the buildozer.spec file. Fixes https://github.com/hbldh/bleak/issues/1184 ?
- 'BLUETOOTH_SCAN' and 'BLUETOOTH_CONNECT' permissions were missing in buildoze.spec file
- android.arch is deprecated, android.archs is preferred now
- updated armeabi-v7a to arm64-v8a as default architecture

**Commit:  17afc9553d2dbc7b41d4185dbb4c13fc25f7ab4e :**
- The bleak recipe was not working, this had two issues:

1. The recipe was quite project directory structure dependant. This works for the example, but not for your own projects and if docker mounts come into play (as is the case for me). To solve this, the recipe now has a url to the latest release of bleak. So now to use bleak in your own app, just copy the recipe and point to it and add the necessary requirements to buildozer.spec.

2.  The current develop branch of p4a installs recipes based on the setup.py file. See  [p4a PythonRecipe](https://github.com/kivy/python-for-android/blob/fa3e5bfe39286e4a293882a4db0728d9bf9f4968/pythonforandroid/recipe.py#L930C5-L957C54). Which is not available anymore in the current develop branch of bleak. As a temporary fix, fix_setup.py is copied into the bleak build directory. 

Looking forward to a response,

Kind regards,

robgar2001

### DIFF
--- a/bleak/backends/p4android/recipes/bleak/__init__.py
+++ b/bleak/backends/p4android/recipes/bleak/__init__.py
@@ -8,34 +8,34 @@ from os.path import join
 
 class BleakRecipe(PythonRecipe):
     version = None  # Must be none for p4a to correctly clone repo
-    fix_setup_py_version = 'bleak develop branch'
-    url = 'git+https://github.com/hbldh/bleak.git'
+    fix_setup_py_version = "bleak develop branch"
+    url = "git+https://github.com/hbldh/bleak.git"
     name = "bleak"
 
     depends = ["pyjnius"]
     call_hostpython_via_targetpython = False
 
-    fix_setup_filename = 'fix_setup.py'
+    fix_setup_filename = "fix_setup.py"
 
     def prepare_build_dir(self, arch):
         super().prepare_build_dir(arch)  # Unpack the url file to the get_build_dir
         build_dir = self.get_build_dir(arch)
 
-        setup_py_path = join(build_dir, 'setup.py')
+        setup_py_path = join(build_dir, "setup.py")
         if not os.path.exists(setup_py_path):
             # Perform the p4a temporary fix
             # At the moment, p4a recipe installing requires setup.py to be present
             # So, we create a setup.py file only for android
 
             fix_setup_py_path = join(self.get_recipe_dir(), self.fix_setup_filename)
-            with open(fix_setup_py_path, 'r') as f:
+            with open(fix_setup_py_path, "r") as f:
                 contents = f.read()
 
             # Write to the correct location and fill in the version number
-            with open(setup_py_path, 'w') as f:
-                f.write(contents.replace('[VERSION]', self.fix_setup_py_version))
+            with open(setup_py_path, "w") as f:
+                f.write(contents.replace("[VERSION]", self.fix_setup_py_version))
         else:
-            info('setup.py found in bleak directory, are you installing older version?')
+            info("setup.py found in bleak directory, are you installing older version?")
 
     def get_recipe_env(self, arch=None, with_flags_in_cc=True):
         env = super().get_recipe_env(arch, with_flags_in_cc)
@@ -48,7 +48,9 @@ class BleakRecipe(PythonRecipe):
 
         info("Copying java files")
         dest_dir = self.ctx.javaclass_dir
-        path = join(self.get_build_dir(arch.arch), "bleak", "backends", "p4android", "java", ".")
+        path = join(
+            self.get_build_dir(arch.arch), "bleak", "backends", "p4android", "java", "."
+        )
 
         shprint(sh.cp, "-a", path, dest_dir)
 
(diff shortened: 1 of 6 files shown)


bleak/backends/p4android/recipes/bleak/__init__.py
```diff
@@ -1,27 +1,40 @@
...
 from pythonforandroid.recipe import PythonRecipe
 from pythonforandroid.toolchain import shprint, info
 import sh
 from os.path import join
 
 
 class BleakRecipe(PythonRecipe):
-    version = None
-    url = None
+    version = '0.20.2'
```
dlech, line 10: Can we do something else so that we don't have to update this version on every release?

examples/kivy/buildozer.spec
```diff
@@ -225,7 +237,7 @@ android.accept_sdk_license = True
 #android.copy_libs = 1
 
 # (str) The Android arch to build for, choices: armeabi-v7a, arm64-v8a, x86, x86_64
-android.arch = armeabi-v7a
+android.archs = arm64-v8a
```
dlech, line 240: Can we have both Arm arches here?

Or if that doubles the build time, then add a note to the readme that this may need to be changed if targeting older devices.

dlech (commented):
Thanks for fixing things up here.

Part of the reason that maintenance on the android backend has fallen behind is that there is not an easy way to build and test the current development snapshot Since this PR is changing things to only build v0.20.2, then I expect the situation will only continue to get worse. So that feels like a step in the wrong direction. Is there any way we can make it easier to build and test local changes to the bleak source for development?

examples/kivy/buildozer.spec
robgar2001: Yes both arm arches are possible, aldough this significantly increases the build time. I have added a note to the readme.

robgar2001:
The commits that I added change the recipe to pull from the github develop branch if no local source path is provided.

After looking in the p4a code, I found that a runtime environment can be set to use local sources instead of pulling.
This is also stated in the readme.

Kind regards,

robgar2001

bleak/backends/p4android/recipes/bleak/fix_setup.py
```diff
@@ -0,0 +1,45 @@
...
+    description=DESCRIPTION,
+    author=AUTHOR,
+    author_email=EMAIL,
+    url=URL,
+    packages=find_packages(exclude=("tests", "examples", "docs", "BleakUWPBridge")),
+    entry_points={"console_scripts": ["bleak-lescan=bleak:cli"]},
+    test_suite="tests",
+    include_package_data=True,
+    license="MIT",
+    classifiers=[
```
dlech, line 25: Since this file is just the minimum needed to install bleak, it would be nice to leave out all of the non-essential parts, e.g. classifiers, entry_points, test_suite, url, author, description, ...

dlech (commented):
Thanks for the updates. I think this looks quite good now. Just one more suggestion (see inline comment)

And it looks like we need to run black to format the code.

bleak/backends/p4android/recipes/bleak/fix_setup.py
robgar2001: So these are the only ones left?

setup(
    name=NAME,
    version=VERSION,
    packages=find_packages(exclude=("tests", "examples", "docs", "BleakUWPBridge")),
)

Not sure about packages, aldough not strictly necessary, it might give problems/be less clean without it.

dlech: Yes, I think those are the required ones. the packages can probably be changed to `find_packages(include=["bleak*"])` rather than the long list of excludes.

robgar2001:
The bleak demo app installs perfectly.

Any further suggestions?

Kind regards,

robgar2001

bleak/backends/p4android/recipes/bleak/fix_setup.py
dlech: `BleakUWPBridge` was removed years ago, so is no longer applicable

dlech:
Looks good, just need to resolve the merge conflicts.

robgar2001:
Thank you for reviewing!

bleak/backends/p4android/recipes/bleak/fix_setup.py
robgar2001: I have changed in the latest commit

robgar2001:
BleakUWP has been removed 😃

dlech:
merged, thanks!

PR #1399, opened 2023-08-25
fix BlueZ wait condition disconnect race

dlech:
Alternative to #1329 and #1377.

@bdraco could you please take a look?

The issue addressed by #1377 seems more widespread, so I took it a bit farther to make sure it is fixed in all public functions (first commit).

Instead of synthesizing a `"Connected"` property change as proposed in #1329, here I have opted to just add a 3rd race condition on `"InterfacesRemoved"` when waiting for services to be resolved (second commit).

### DIFF
--- a/bleak/backends/bluezdbus/manager.py
+++ b/bleak/backends/bluezdbus/manager.py
@@ -170,6 +170,22 @@ class BlueZManager:
         self._condition_callbacks: Set[Callable] = set()
         self._services_cache: Dict[str, BleakGATTServiceCollection] = {}
 
+    def _check_adapter(self, adapter_path: str) -> None:
+        """
+        Raises:
+            BleakError: if adapter is not present in BlueZ
+        """
+        if adapter_path not in self._properties:
+            raise BleakError(f"adapter '{adapter_path.split('/')[-1]}' not found")
+
+    def _check_device(self, device_path: str) -> None:
+        """
+        Raises:
+            BleakError: if device is not present in BlueZ
+        """
+        if device_path not in self._properties:
+            raise BleakError(f"device '{device_path.split('/')[-1]}' not found")
+
     async def async_init(self):
         """
         Connects to the D-Bus message bus and begins monitoring signals.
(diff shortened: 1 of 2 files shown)


bleak/backends/bluezdbus/manager.py
```diff
@@ -668,22 +693,66 @@ async def _wait_for_services_discovery(self, device_path: str) -> None:
...
+
+            done, _ = await asyncio.wait(
+                {
+                    services_discovered_wait_task,
+                    device_disconnected_wait_task,
+                    device_removed_wait_task,
+                },
+                return_when=asyncio.FIRST_COMPLETED,
+            )
+
```
dlech, line 723: Based on https://github.com/hbldh/bleak/pull/1329#discussion_r1273575569, it looks like we may need to check `done` for exceptions.

bdraco: Yes thats possible since tasks by default are not eager (at least by default and not unless https://github.com/python/cpython/issues/97696) and won't run until the next event loop so `_wait_condition` can try to access a removed key on the next iteration of the event loop

dlech:
Updated:

- added helper functions and additional checks to handle lack of eager evaluation of `_wait_*` coros in first commit
- added checking of done tasks for exceptions in second commit
- added 3rd commit to handle potential issues with condition callbacks

dlech:
@bdraco let me know if you want to test before merging

bdraco:
Will give it a shot shortly

bleak/backends/bluezdbus/manager.py
```diff
@@ -167,9 +174,25 @@ def __init__(self):
...
+        if adapter_path not in self._properties:
+            raise BleakError(f"adapter '{adapter_path.split('/')[-1]}' not found")
+
+    def _check_device(self, device_path: str) -> None:
+        """
+        Raises:
+            BleakError: if device is not present in BlueZ
+        """
+        if device_path not in self._properties:
+            raise BleakError(f"adapter '{device_path.split('/')[-1]}' not found")
```
dlech, line 194:
```suggestion
            raise BleakError(f"device '{device_path.split('/')[-1]}' not found")
```

bdraco (approved):
Tested, could not replicate race after this PR.

Note: it took days of testing on the other one to be sure so I can't be 100% but this seems good.

bdraco:
Ran overnight frequent polling.  Didn't notice the safety timeout ever kicking in anymore.  Could be just luck but I think its solved

bdraco:
Running for almost 48 hours now and no more safety timeouts.  I think this is solved 👍

dlech:
Thanks for testing!

PR #1400, opened 2023-08-25
Optimize BlueZ device watchers and condition callbacks to avoid linear searches

bdraco:
~~needs #1399~~


bleak/backends/bluezdbus/manager.py
```diff
@@ -166,10 +173,26 @@ def __init__(self):
 
         self._advertisement_callbacks: List[CallbackAndState] = []
         self._device_removed_callbacks: List[DeviceRemovedCallbackAndState] = []
-        self._device_watchers: Set[DeviceWatcher] = set()
-        self._condition_callbacks: Set[Callable] = set()
+        self._device_watchers: Dict[str, list[DeviceWatcher]] = {}
```
dlech, line 176: Would set be better than list?

Also we need to support Python < 3.10, so...
```suggestion
        self._device_watchers: Dict[str, List[DeviceWatcher]] = {}
```

bdraco:
rebasing..

bleak/backends/bluezdbus/manager.py
```diff
@@ -166,10 +173,26 @@ def __init__(self):
...
+        self._condition_callbacks: Dict[str, Set[Callable[[Dict[str, Any], None]]]] = {}
```
dlech, line 177: Let's modify and continue to use `DevicePropertiesChangedCallback` here.

bleak/backends/bluezdbus/manager.py
```diff
@@ -582,7 +582,10 @@ def remove_device_watcher(self, watcher: DeviceWatcher) -> None:
             The device watcher token that was returned by
             :meth:`add_device_watcher`.
         """
-        self._device_watchers.remove(watcher)
+        device_path = watcher.device_path
+        self._device_watchers[device_path].remove(watcher)
+        if not self._device_watchers[device_path]:
+            del self._device_watchers[device_path]
```
dlech, line 588: Probably doesn't hurt to leave the list since devices are likely to be seen again (you're probably the only person that has 1000s of devices around :laughing:)

bleak/backends/bluezdbus/manager.py
```diff
@@ -804,21 +807,21 @@ async def _wait_condition(
...
+        device_callbacks.add(_wait_condition_callback)
 
         try:
             # can be canceled
             await event.wait()
         finally:
-            device_callbacks.remove(callback)
+            device_callbacks.remove(_wait_condition_callback)
+            if not device_callbacks:
+                del self._condition_callbacks[device_path]
```
dlech, line 824: Same here, seems simpler to just leave the set.

bleak/backends/bluezdbus/manager.py
```diff
@@ -946,36 +951,38 @@ def _parse_msg(self, message: Message):
...
 
                     # handle device condition watchers
-                    for condition_callback in self._condition_callbacks.get(
-                        message.path, ()
-                    ):
-                        condition_callback()
+                    condition_callbacks = self._condition_callbacks.get(device_path)
+                    if condition_callbacks:
+                        for condition_callback in condition_callbacks:
+                            condition_callback(unpacked_changed)
```
dlech, line 964: This does not take into account that properties can be invalidated (removed). IMHO, it was simpler the way it was to let the handler just get the property instead of trying to recreate the logic twice.

bleak/backends/bluezdbus/manager.py
```diff
@@ -804,21 +807,21 @@ async def _wait_condition(
...
         event = asyncio.Event()
 
-        def callback():
-            if (
-                self._properties[device_path][defs.DEVICE_INTERFACE][property_name]
-                == property_value
-            ):
+        def _wait_condition_callback(changed: Dict[str, Any]) -> None:
+            """Callback for when a property changes."""
+            if changed.get(property_name) == property_value:
```
dlech, line 812: I think there could be some potential issues here with `None == False`. As mentioned in another comment, might be simpler to just leave it the way it was without passing the `changed` arg.

bleak/backends/bluezdbus/manager.py
```diff
@@ -946,36 +951,38 @@ def _parse_msg(self, message: Message):
...
-                            # callbacks may remove the watcher, hence the copy() above
-                            if message.path == device_path:
-                                on_connected_changed(self_interface["Connected"])
+                        watchers = self._device_watchers.get(device_path)
+                        if watchers:
+                            # callbacks may remove the watcher, hence the view
+                            for watcher in watchers[:]:
+                                watcher.on_connected_changed(
+                                    self_interface["Connected"]
+                                )
```
dlech, line 974: This looks leftover from the previous PR and should not be needed.

bleak/backends/bluezdbus/manager.py
bdraco: 9c94230

bleak/backends/bluezdbus/manager.py
bdraco: I was a bit worried about devices with random mac addresses would build up and leak memory over time

bleak/backends/bluezdbus/manager.py
bdraco: https://github.com/hbldh/bleak/pull/1400/files#r1307604639

bleak/backends/bluezdbus/manager.py
bdraco: Just realized this is still needed put back in 9874de6cdd8f52656d6e4b1ef129cdb64f8484ee

dlech: My mistake, you are right.

bleak/backends/bluezdbus/manager.py
dlech: Ah, didn't consider that case.

bleak/backends/bluezdbus/manager.py
bdraco: made it named tuple in 5e40e914d0a20597fcf7e89df12f1457aeb8c6a4

bleak/backends/bluezdbus/manager.py
bdraco: I changed the incoming to pull from `self_interface.get` so `None` will still work

bdraco: I left it passing the value because I always worry about race conditions when we have to look it up again. Not a problem right now but could be in the future if something gets refactored

bdraco:
marking as draft while I clean up and retest

bdraco:
Screenshot 2023-08-28 at 12 07 23 PM
_parse_msg

bleak/backends/bluezdbus/manager.py
```diff
@@ -173,8 +193,8 @@ def __init__(self):
 
         self._advertisement_callbacks: List[CallbackAndState] = []
         self._device_removed_callbacks: List[DeviceRemovedCallbackAndState] = []
-        self._device_watchers: Set[DeviceWatcher] = set()
-        self._condition_callbacks: Dict[str, Set[DevicePropertiesChangedCallback]] = {}
+        self._device_watchers: Dict[str, Set[DeviceWatcher]] = {}
+        self._condition_callbacks: Dict[str, set[DeviceConditionCallback]] = {}
```
dlech, line 197:
```suggestion
        self._condition_callbacks: Dict[str, Set[DeviceConditionCallback]] = {}
```

pyproject.toml
```diff
@@ -44,6 +44,7 @@ asynctest = { version = "^0.13.0",  python = "<3.8" }
 pytest = "^7.0.0"
 pytest-asyncio = "^0.19.0"
 pytest-cov = "^3.0.0 "
+dbus-fast = { version = "^1.83.0", markers = "platform_system == 'Linux'" }
```
dlech, line 47: Can we avoid the duplicate requirement?

bleak/backends/bluezdbus/manager.py
```diff
@@ -946,36 +971,40 @@ def _parse_msg(self, message: Message):
 
                 if interface == defs.DEVICE_INTERFACE:
```
bdraco, line 972: This is the hot path since its where all the RSSI updates come in

bleak/backends/bluezdbus/manager.py
```diff
@@ -946,36 +971,40 @@ def _parse_msg(self, message: Message):
...
                     # handle advertisement watchers
+                    device_path = message_path
 
                     self._run_advertisement_callbacks(
-                        message.path, cast(Device1, self_interface), changed.keys()
+                        device_path, cast(Device1, self_interface), changed.keys()
                     )
```
bdraco, line 978: This could be indexed by device when the user has multiple bluetooth adapters it has to reject callbacks for the wrong device. But that should be another PR

poetry.lock
```diff
@@ -1,9 +1,10 @@
-# This file is automatically @generated by Poetry 1.5.1 and should not be changed by hand.
+# This file is automatically @generated by Poetry and should not be changed by hand.
```
dlech, line 1: Looks like this was done with an older version of poetry.

bleak/backends/bluezdbus/manager.py
bdraco: 🤦 So used to py3.11 only now

pyproject.toml
bdraco: Yes, its not actually needed since I changed the test. I forgot to revert the poetry change.

done in e2d7f2b48a1a3454a33e89c1194e6df8b98e046f

poetry.lock
bdraco: reverted in e2d7f2b48a1a3454a33e89c1194e6df8b98e046f

bdraco:
retesting on production one more time just to be sure

bdraco:
all good 👍

bdraco:
Thanks for the review. Will be very happy to get a new release so we can get https://github.com/esphome/aioesphomeapi/pull/528 sorted out

dlech:
Hopefully we can get a release by the end of the week. There is still a Windows issue I would like to address and also prepare for Python 3.12.

Issue #1412, opened 2023-09-04
linux TypeError: 'StructMeta' object is not iterable


wangguan1888:
* bleak version:0.21.0
* Python version:3.8.11
* Operating System:Linux Yocto
* BlueZ version (`bluetoothctl -v`) in case of Linux:5.49

### Description
Trying to run the example code results in TypeError: 'StructMeta' object is not iterable.

### What I Did
1.install bleak: pip3 install bleak
2.run the example code
```
import asyncio
from bleak import BleakScanner

async def main():
    devices = await BleakScanner.discover()
    for d in devices:
        print(d)

asyncio.run(main())
```

3.Got the following result.
```
Traceback (most recent call last):
  File "test1.py", line 9, in <module>
    asyncio.run(main())
  File "/usr/lib/python3.8/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/usr/lib/python3.8/asyncio/base_events.py", line 616, in run_until_complete
    return future.result()
  File "test1.py", line 5, in main
    devices = await BleakScanner.discover()
  File "/usr/lib/python3.8/site-packages/bleak/__init__.py", line 317, in discover
    async with cls(**kwargs) as scanner:
  File "/usr/lib/python3.8/site-packages/bleak/__init__.py", line 145, in __init__
    get_platform_scanner_backend_type() if backend is None else backend
  File "/usr/lib/python3.8/site-packages/bleak/backends/scanner.py", line 284, in get_platform_scanner_backend_type
    from bleak.backends.bluezdbus.scanner import BleakScannerBlueZDBus
  File "/usr/lib/python3.8/site-packages/bleak/backends/bluezdbus/scanner.py", line 5, in <module>
    from dbus_fast import Variant
  File "/usr/lib/python3.8/site-packages/dbus_fast/__init__.py", line 1, in <module>
    from . import aio, glib, introspection, message_bus, proxy_object, service
  File "/usr/lib/python3.8/site-packages/dbus_fast/glib/__init__.py", line 1, in <module>
    from .message_bus import MessageBus
  File "/usr/lib/python3.8/site-packages/dbus_fast/glib/message_bus.py", line 20, in <module>
    from .proxy_object import ProxyObject
  File "/usr/lib/python3.8/site-packages/dbus_fast/glib/proxy_object.py", line 15, in <module>
    from gi.repository import GLib
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
[...]

PR #1413, opened 2023-09-04
backends/bluezdbus: don't import dbus_next directly

dlech:
The top-level dbus_next module tries to import the GLib backend that is not used by Bleak. This wastes memory and can cause a crash if the available version of `gi` is not compatible.

Fixes: https://github.com/hbldh/bleak/issues/1412


dlech:
Did not fix the issue.

PR #1415, opened 2023-09-08
Updated `dbus-fast` dependency to v2.0.0.

dlech:
This updates to a new version of `dbus-fast` that doesn't try to import
`gi` anymore.

Fixes: https://github.com/hbldh/bleak/issues/1412

PR #1416, opened 2023-09-08
Loosen dbus-fast pin to allow for dbus-fast 2.0

bdraco:
Looks like a new release will be needed as well to unblock https://github.com/home-assistant/core/pull/99894


dlech:
Probably better than https://github.com/hbldh/bleak/pull/1415.

Also need to run `poetry lock --no-update` with latest poetry (v1.6.1).

PR #1439, opened 2023-10-13
Add BleakCharacteristicNotFoundError

emontnemery:
## Change proposal

Add `BleakCharacteristicNotFoundError` which is raised if a device does not support a characteristic

## Rationale

It's not uncommon for devices to change its set of supported characteristics, for example during an initial setup.
Without a specialized exception class it's tricky to detect when this happens.

### DIFF
--- a/bleak/exc.py
+++ b/bleak/exc.py
@@ -18,9 +18,7 @@ class BleakCharacteristicNotFoundError(BleakError):
 
     char_specifier: Union[int, str, uuid.UUID]
 
-    def __init__(
-        self, char_specifier: Union[int, str, uuid.UUID], *args: object
-    ) -> None:
+    def __init__(self, char_specifier: Union[int, str, uuid.UUID]) -> None:
         """
         Args:
             characteristic (str): handle or UUID of the characteristic which was not found
--- a/bleak/exc.py
+++ b/bleak/exc.py
@@ -1,11 +1,32 @@
 # -*- coding: utf-8 -*-
-from typing import Optional
+import uuid
+from typing import Optional, Union
 
 
 class BleakError(Exception):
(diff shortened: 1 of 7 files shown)


bleak/exc.py
```diff
@@ -8,6 +9,26 @@ class BleakError(Exception):
...
+    """
+    Exception which is raised if a device does not support a characteristic.
+
+    .. versionadded: 0.22.0
+    """
+
+    char_specifier: Union[int, str, uuid.UUID]
+
+    def __init__(
+        self, char_specifier: Union[int, str, uuid.UUID], *args: object
```
dlech, line 22: What is `*args` for? Can we leave it out?

emontnemery: It's there just to stick to the same signature as that of `Exception.__init__`, but I think it can be omitted.
@bdraco do you agree?

bdraco: Seems fine to omit 👍

bleak/exc.py
```diff
@@ -8,6 +9,24 @@ class BleakError(Exception):
...
+    """
+
+    char_specifier: Union[int, str, uuid.UUID]
+
+    def __init__(self, char_specifier: Union[int, str, uuid.UUID]) -> None:
+        """
+        Args:
+            characteristic (str): handle or UUID of the characteristic which was not found
+        """
+        super().__init__(f"Characteristic {char_specifier} was not found!", *args)
```
dlech, line 26:
```suggestion
        super().__init__(f"Characteristic {char_specifier} was not found!")
```

emontnemery: Oops 🤦

dlech:
merged, thanks!

PR #1452, opened 2023-11-13
Updated missing Bluetooth SIG characteristics and service UUIDs

jpstotz:
As the Bluetooth SIG now provides UUIDs in YAML format and I noticed some were missing I updated the bleak uuid dictionary:
https://bitbucket.org/bluetooth-SIG/public/src/main/assigned_numbers/uuids/characteristic_uuids.yaml
https://bitbucket.org/bluetooth-SIG/public/src/main/assigned_numbers/uuids/service_uuids.yaml

Just for the records, I used this script to identify missing entries: 

```
from bleak.uuids import uuid16_dict
import yaml  ## pyyaml

file_name = 'service_uuids.yaml'  # or 'characteristic_uuids.yaml'

with open(file_name, 'r', encoding='utf-8') as file:
    yaml_data = yaml.safe_load(file)

for c in yaml_data['uuids']:
    val = int(c['uuid'])
    name = c['name']
    if val not in uuid16_dict:
        print('    0x{:02X}: "{}",'.format(val, name))
```


dlech:
Thanks. This looks fine.

In the future, it could be nice to make a separate library that is just automatically generated from those yaml files that could be used by any Bluetooth library, not just Bleak.

koenvervloesem:
I made this project exactly for this purpose: https://github.com/koenvervloesem/bluetooth-numbers

Generated from Nordic Semiconductor's Bluetooth numbers database because the Bluetooth SIG's YAML files didn't exist when I started the project, but I'm planning to integrate those YAML files in a later version. Feel free to open a discussion or issue in my project for suggestions.

PR #1488, opened 2023-12-30
fix: incorrect mypy diagnosis of Unpack[ExtraArgs].

### LINKED ISSUE #1487: mypy / error: Missing named argument ... [call-arg] / for optional kwargs
* bleak version: 0.21.1
* Python version: 3.10, 3.12
* Operating System: Ubuntu 22.04, Windows 11
* BlueZ version (`bluetoothctl -v`) in case of Linux: NA

### Description

Mypy treats kwargs that are `Unpack`ed as required.  In fact, these kwargs were always intended as platform-specific optional arguments.  

Mypy will diagnose missing kwargs as errors.  In bleak, this occurs for any function that types kwargs as `Unpack[ExtraArgs]`.  These include the methods discover, find_device_by_address, find_device_by_name, and find_device_by_filter.

Where a more friendly solution may be to use `Generic` and type aliases, use of `Unpack` is a welcome improvement over kwargs, so I believe that it's simple enough to add total=False because ALL of the kwargs are optional.  Refer to https://peps.python.org/pep-0692/#required-and-non-required-keys.  If finer grained control is required, then I suggest Generic and Type Aliases so that the user can see the arguments in their editor instead of having to rely on documentation.

### What I Did

Ran mypy 1.7.1 on a file like:

```python

async def main() -> None:
    await BleakScanner.find_device_by_address("address")
```

`mypy .` will diagnose every missing kwarg as error:

```
smpclient\blah.py:5: error: Missing named argument "service_uuids" for "find_device_by_address" of "BleakScanner"  [call-arg]
smpclient\blah.py:5: error: Missing named argument "scanning_mode" for "find_device_by_address" of "BleakScanner"  [call-arg]
smpclient\blah.py:5: error: Missing named argument "bluez" for "find_device_by_address" of "BleakScanner"  [call-arg]
smpclient\blah.py:5: error: Missing named argument "cb" for "find_device_by_address" of "BleakScanner"  [call-arg]
smpclient\blah.py:5: error: Missing named argument "backend" for "find_device_by_address" of "BleakScanner"  [call-arg]
```

### Logs

N/A

JPHutchins:
Fixes #1487 in the simplest way, according to https://peps.python.org/pep-0692/#required-and-non-required-keys.  This has the effect of making the `BleakScanner` `ExtraArgs` optional instead of required.

For reference:

```python
    class ExtraArgs(TypedDict):
        """
        Keyword args from :class:`~bleak.BleakScanner` that can be passed to
        other convenience methods.
        """

        service_uuids: List[str]
        """
        Optional list of service UUIDs to filter on. Only advertisements
        containing this advertising data will be received. Required on
        macOS >= 12.0, < 12.3 (unless you create an app with ``py2app``).
        """
        scanning_mode: Literal["active", "passive"]
        """
        Set to ``"passive"`` to avoid the ``"active"`` scanning mode.
        Passive scanning is not supported on macOS! Will raise
        :class:`BleakError` if set to ``"passive"`` on macOS.
        """
        bluez: BlueZScannerArgs
        """
        Dictionary of arguments specific to the BlueZ backend.
        """
        cb: CBScannerArgs
        """
        Dictionary of arguments specific to the CoreBluetooth backend.
        """
        backend: Type[BaseBleakScanner]
        """
        Used to override the automatically selected backend (i.e. for a
            custom backend).
        """
```

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -22,6 +22,7 @@ Contributors
 * Edward Betts <edward@4angle.com>
 * Robbe Gaeremynck <robbe.gaeremynck@outlook.com>
 * David Johansen <davejohansen@gmail.com>
+* JP Hutchins <jphutchins@gmail.com>
 
 Sponsors
 --------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -22,6 +22,7 @@ Changed
 Fixed
 -----
 * Fixed BlueZ version in passive scanning error message. Fixes #1433.
+* Fixed mypy requiring ``Unpack[ExtraArgs]`` that were intended to be optional.  Fixes #1487.
 
 `0.21.1`_ (2023-09-08)
 ======================
--- a/bleak/__init__.py
+++ b/bleak/__init__.py
@@ -245,7 +245,7 @@ class BleakScanner:
         finally:
             unregister_callback()
 
-    class ExtraArgs(TypedDict):
+    class ExtraArgs(TypedDict, total=False):
         """
         Keyword args from :class:`~bleak.BleakScanner` that can be passed to
         other convenience methods.


dlech:
Thanks!

PR #1510, opened 2024-02-13
Fix definition order in typing and remove duplicate definition

Siecje:


dlech:
Nice, thanks. Can you run `black` on it to fix the linting error?

Issue #1507, opened 2024-02-12
KeyError: 'org.bluez.Device1'


zevv:
- raspberry pi 4B
- debian bookworm
- bleak 0.20.2-1 from python3-bleak package
- bluetoothctl: 5.66

I'm having a very hard time getting my code to run stable; i'm basically round-robin polling about 60 bluetooth devices (For details of my setup see #1500)

Apart from the occasional problem at the kernel/driver level, my code often runs into the error below:

```
ERROR:root:A message handler raised an exception: 'org.bluez.Device1'.
Traceback (most recent call last):
  File "src/dbus_fast/message_bus.py", line 811, in dbus_fast.message_bus.BaseMessageBus._process_message
  File "/usr/lib/python3/dist-packages/bleak/backends/bluezdbus/manager.py", line 854, in _parse_msg
    condition_callback()
  File "/usr/lib/python3/dist-packages/bleak/backends/bluezdbus/manager.py", line 709, in callback
    self._properties[device_path][defs.DEVICE_INTERFACE][property_name]
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
KeyError: 'org.bluez.Device1'
```

I can't seem to relate this to anything I'm doing as none of my code is in the traceback - what could be the cause of above message, and how do I avoid this situation?

Thank you

dlech:
Enabling debug logs should at least show us the order in which things are happening to try to see where the issue is. (It could be helpful to compare with the "good" debug logs from the other issue to see what the difference is.)

zevv:
Of course, I should have made an shared the debug log in the first place, will do this today. thank you.

zevv:
Log attached below; it's quite large but fortunately the problem popped up within the first few minutes.

It starts with a discovery and then finds the first devices. The program attempts to connect and poll these, but the first few ones fail: my guess is that people just carried in a box with 40 devices into the building but are not properly within range yet. These failed attempts are expected and should not matter. (Timeouts, 

A bit later (approx line 2475) some devices are properly connected and polled, all is well until line 7657 where the first error pops up: 

```
[...]

PR #1511, opened 2024-02-14
Improve error handling when a device is not present in BlueZ

Siecje:
Fixes https://github.com/hbldh/bleak/issues/1507

### DIFF
--- a/bleak/backends/bluezdbus/manager.py
+++ b/bleak/backends/bluezdbus/manager.py
@@ -961,6 +961,12 @@ class BlueZManager:
                         del self._descriptor_map[obj_path]
                     except KeyError:
                         pass
+
+            # Remove empty properties when all interfaces have been removed.
+            # This avoids wasting memory for people who have noisy devices
+            # with private addresses that change frequently.
+            if obj_path in self._properties and not self._properties[obj_path]:
+                del self._properties[obj_path]
         elif message.member == "PropertiesChanged":
             interface, changed, invalidated = message.body
             message_path = message.path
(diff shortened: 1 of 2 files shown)


bleak/backends/bluezdbus/manager.py
```diff
@@ -917,6 +936,9 @@ def _parse_msg(self, message: Message):
                 except KeyError:
                     pass
 
+                if obj_path in self._properties and not self._properties[obj_path]:
```
dlech, line 939: I think this should be after the for loop so that it only runs once after all interfaces have been removed. A comment would be nice too since it isn't super-obvious why we are doing this.

dlech: Also wouldn't hurt to split this one out into a separate commit since it isn't strictly related to the other changes.

bleak/backends/bluezdbus/manager.py
```diff
@@ -216,6 +216,28 @@ def _check_device(self, device_path: str) -> None:
...
+        self, device_path: str, interface: str, property_name: str
+    ) -> Any:
+        self._check_device(device_path)
+        device_properties = self._properties[device_path]
+
+        try:
+            interface_properties = device_properties[interface]
+        except KeyError:
+            raise BleakError(
+                f"Interface {interface} not found for device '{device_path.split('/')[-1]}'"
```
dlech, line 229: Might be helpful to keep the full path for cases where people have multiple Bluetooth adapters.

dlech:
This isn't a trivial change, so I would like to add a commit message explaining the reason why these changes are needed and also add a changelog entry for the fix. Thanks!

bleak/backends/bluezdbus/manager.py
Siecje: Okay I copied the `_check_device()` message.

dlech: Ha, OK. I'll leave it up to you.

PR #1512, opened 2024-02-14
Partial typehints for bleak/__init__.py

Siecje:


bleak/__init__.py
```diff
@@ -154,11 +155,16 @@ def __init__(
             **kwargs,
         )
 
-    async def __aenter__(self):
+    async def __aenter__(self) -> "BleakScanner":
```
dlech, line 158: We should be able to drop the quotes since we have `from __future__ import annotations` in this file already.

bleak/__init__.py
```diff
@@ -513,19 +519,21 @@ def __init__(
...
-            if services is None
-            else set(map(normalize_uuid_str, services)),
+            disconnected_callback=(
+                None
+                if disconnected_callback is None
+                else functools.partial(disconnected_callback, self)
+            ),
+            services=(
+                None if services is None else set(map(normalize_uuid_str, services))
+            ),
```
dlech, line 536: This looks unrelated. I'm guessing `black` did this?

bleak/__init__.py
```diff
@@ -553,19 +559,24 @@ def mtu_size(self) -> int:
...
         return f"{self.__class__.__name__}, {self.address}"
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         return f"<{self.__class__.__name__}, {self.address}, {type(self._backend)}>"
 
     # Async Context managers
 
-    async def __aenter__(self):
+    async def __aenter__(self) -> "BleakClient":
```
dlech, line 570: Can remove the quotes here too.

bleak/__init__.py
Siecje: Yes black.

dlech:
Merged, thanks!

PR #1514, opened 2024-02-15
More typehints

Siecje:


bleak/backends/service.py
```diff
@@ -21,10 +21,10 @@
 class BleakGATTService(abc.ABC):
     """Interface for the Bleak representation of a GATT Service."""
 
-    def __init__(self, obj):
+    def __init__(self, obj) -> None:
```
Siecje, line 24: What should `obj` be?

dlech: `typing.Any`

Siecje: I saw that but what are the possible types? Can we list them?

dlech: They are backend-specific so we can't because it would cause import errors. Also, this is considered internal API, so I don't want people expecting any specific stable type.

dlech:
Thanks!

PR #1516, opened 2024-02-20
Fix ruff suggestions

Siecje:


bleak/uuids.py
```diff
@@ -1,10 +1,9 @@
-# -*- coding: utf-8 -*-
+from __future__ import annotations
 
-from typing import Dict
 from uuid import UUID
 
 
-uuid16_dict: Dict[int, str] = {
+uuid16_dict: dict[int, str] = {
```
dlech, line 6: We still support Python 3.8 so can't make this change yet.

Siecje: This also adds `from __future__ import annotations`

dlech: Hmm... I guess that prevents runtime errors, but it still technically isn't valid. Since Python 3.8 is reaching end of life later this year, I'm tempted to defer this until then.

Siecje: `bleak/__init__.py` has `from __future__ import annotations`

https://github.com/hbldh/bleak/blob/5e294f4fcdc3effac147d43e29697373e3209901/bleak/__init__.py#L5

PR #1517, opened 2024-02-21
Additional typehints for bluezdbus code

Siecje:


bleak/backends/bluezdbus/utils.py
```diff
@@ -23,11 +24,11 @@ def assert_reply(reply: Message):
     assert reply.message_type == MessageType.METHOD_RETURN
 
 
-def validate_address(address):
+def validate_address(address: str) -> bool:
```
Siecje, line 27: This is not used. Can it be removed?

dlech: sounds reasonable

PR #1518, opened 2024-02-22
Additional typehints for corebluetooth

Siecje:


bleak/backends/corebluetooth/client.py
```diff
@@ -111,7 +111,7 @@ async def connect(self, **kwargs) -> bool:
                 self._peripheral
             )
 
-        def disconnect_callback():
+        def disconnect_callback() -> bool:
```
dlech, line 114: Shouldn't this be `None` instead of `bool`?

Siecje: Yes, good catch! Fixed.

Issue #828, opened 2022-05-11
Any way to start_notify if the UUID characteristic has  no Descriptor


ps2229:
* bleak version:0.14.3
* Python version:3.9
* Operating System:Android
* BlueZ version (`bluetoothctl -v`) in case of Linux:

### Description

Trying to use start_notify function but getting errors like no descriptor, and after checking the bluetooth module there is no descriptor in the UUID characteristic , is there any way to still use start_notify without the descriptor.Using bluno bettle ble module

### What I Did
 On some forums I found solutions like using 'gatt.setCharacteristicNotification' but really confused as to call it where in the code.? Any idea how to go about it.

hbldh:
This is not really a Bleak issue. If `start_notify` delivers error that no descriptor is present, then it seems like you really need to add these in your peripheral to enable notifications.

I have no idea what is needed to be done on your a
Arduino, but Bleak cannot alleviate problems with incorrectly implemented peripherals.

NewAge12:
Having the same issue with the error using bluno bettle ble  --->> bleak.exc.BleakError: Descriptor None was not found!  
It does not contain the CLIENT_CHARACTERISTIC_CONFIGURATION_UUID='"00002902-0000-1000-8000-00805f9b34fb" , So the start_notify is giving errors, any way to bypass this and read the values using start_notify.??

NewAge12:
d

Egidio92:
I've the same problem too

NewAge12:
In the p4android folder change the client.py file to this and it will work.
# -*- coding: utf-8 -*-
"""
BLE Client for python-for-android
"""
import asyncio
import logging
import uuid
import warnings
from typing import Callable, Optional, Union

from bleak.backends.device import BLEDevice
from bleak.backends.service import BleakGATTServiceCollection
from bleak.exc import BleakError
from bleak.backends.client import BaseBleakClient
from bleak.backends.p4android.service import BleakGATTServiceP4Android
from bleak.backends.p4android.characteristic import BleakGATTCharacteristicP4Android
from bleak.backends.p4android.descriptor import BleakGATTDescriptorP4Android

from android.broadcast import BroadcastReceiver
from jnius import java_method

from . import defs
from . import utils

logger = logging.getLogger(__name__)


class BleakClientP4Android(BaseBleakClient):
    """A python-for-android Bleak Client

    Args:
[...]

Issue #972, opened 2022-09-02
startNotify fails with message 0x0e (Connection Rejected Due To Security Reasons)


fsaris:
* bleak version: 0.15.1
* Python version: 3.10
* Operating System: Ubuntu
* BlueZ version (`bluetoothctl -v`) in case of Linux: 5.64

### Description

I'm porting some code to control lights from bluepy to bleak.

I got the connection and controlling the device to work. But when I want to start listening for notifications I get the notification through the callback but it fails after a short time with the message ` [org.bluez.Error.Failed] Operation failed with ATT error: 0x0e (Connection Rejected Due To Security Reasons)`.

 
### What I Did

A simplified version of the code:

```
        def disconnected_callback(client):
            _LOGGER.debug("Disconnected callback called!")

        self._client = BleakClient(self.mac, disconnected_callback=disconnected_callback)

        # Connect
        await self._client.connect()

        _LOGGER.debug("Connected to %s", self.mac)

        _LOGGER.debug("Sent pair message"
        await self._client.write_gatt_char(PAIR_CHAR_UUID, message)

        _LOGGER.debug("Read pair response")
        value = await self._client.read_gatt_char(PAIR_CHAR_UUID)

... process pair to some session info

        # Tell the device to start sending out notifications
        await self._client.write_gatt_char(STATUS_CHAR_UUID, b'\x01')


        def callback(sender: int, data: bytearray):
            _LOGGER.debug(f"callback: {sender}: {data}")
        
        _LOGGER.debug("Start notify")
        await self._client.start_notify(STATUS_CHAR_UUID, callback)
```

This results in:

```
DEBUG:awoxmeshlight:Connected to a4:c1:38:8a:ba:94
DEBUG:awoxmeshlight:Sent pair message
DEBUG:bleak.backends.bluezdbus.client:Write Characteristic 00010203-0405-0607-0809-0a0b0c0d1914 | /org/bluez/hci0/dev_A4_C1_38_8A_BA_94/service0010/char001a: bytearray(b'\x0c\r8\x84\x019g\xf2\r\x91\xc4\xa3\xb0,Z\xda$')
DEBUG:awoxmeshlight:Sent status message
DEBUG:bleak.backends.bluezdbus.client:Write Characteristic 00010203-0405-0607-0809-0a0b0c0d1911 | /org/bluez/hci0/dev_A4_C1_38_8A_BA_94/service0010/char0011: b'\x01'
DEBUG:awoxmeshlight:Read pair response
[...]

PR #1527, opened 2024-03-17
andorid start_notify no Descriptor

tsiens:
#828

### DIFF
--- a/bleak/backends/p4android/client.py
+++ b/bleak/backends/p4android/client.py
@@ -429,13 +429,26 @@ class BleakClientP4Android(BaseBleakClient):
 
     async def start_notify(
         self,
-        characteristic: BleakGATTCharacteristic,
+        char_specifier: Union[BleakGATTCharacteristicP4Android, int, str, uuid.UUID],
         callback: NotifyCallback,
         **kwargs,
     ) -> None:
         """
         Activate notifications/indications on a characteristic.
+
+        Args:
+            char_specifier (BleakGATTCharacteristicP4Android, int, str or UUID): The characteristic to deactivate
+                notification/indication on, specified by either integer handle, UUID or
+                directly by the BleakGATTCharacteristicP4Android object representing it.
+
         """
+        if not isinstance(char_specifier, BleakGATTCharacteristicP4Android):
+            characteristic = self.services.get_characteristic(char_specifier)
+        else:
+            characteristic = char_specifier
+        if not characteristic:
+            raise BleakCharacteristicNotFoundError(char_specifier)
+            
         self._subscriptions[characteristic.handle] = callback
 
         assert self.__gatt is not None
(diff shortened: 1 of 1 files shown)


tsiens:
What happened？Format and Lint？I don't understand

dlech:
> Format and Lint？I don't understand

We use `black` to automatically format the code. So the CI failure is because `black` wasn't run on the changes.

Also, I don't think this is the right way to fix the issue. It seems like the if statements are just silently ignoring the bug.

Can you explain more about what you are trying to fix and why this fixes it?

tsiens:
> > Format and Lint？I don't understand
> 
> We use `black` to automatically format the code. So the CI failure is because `black` wasn't run on the changes.
> 
> Also, I don't think this is the right way to fix the issue. It seems like the if statements are just silently ignoring the bug.
> 
> Can you explain more about what you are trying to fix and why this fixes it?

I am developing an APK that uses a ble device，
1、In function  “start_notify“，characteristic can be a string like the  function 'stop_notify'
2、I don't know why my ble notification_descriptor is empty，self.write_gatt_descriptor will report an error，i think it's necessary to be compatible with empty notification_descriptor

dlech:
> In function “start_notify“，characteristic can be a string like the function 'stop_notify'

This should already be handled in the `start_notify()` in `bleak/__init__.py`. Is it not working?

> I don't know why

I would like to understand what is actually going on before considering merging these change. E.g. does the same Python script work on other OSes but not on Android. Can we get logs showing what descriptors are actually reported by the device. Are we sure this isn't a defective device like those described in https://github.com/hbldh/bleak/issues/972?

tsiens:
> > In function “start_notify“，characteristic can be a string like the function 'stop_notify'
> 
> This should already be handled in the `start_notify()` in `bleak/__init__.py`. Is it not working?
> 
> > I don't know why
> 
> I would like to understand what is actually going on before considering merging these change. E.g. does the same Python script work on other OSes but not on Android. Can we get logs showing what descriptors are actually reported by the device. Are we sure this isn't a defective device like those described in #972?

my code is working properly on Windows，
in bleak/p4android/client.py start_nofity，The original code required input BleakGATTCharacteristic object，This is unreasonable，Because the program will input a string char uuid just like stop_nofity,So the char_specifier should be modified to char_specifier: Union[BleakGATTCharacteristic, int, str, uuid.UUID].

and bleak/__init__.py start_nofity also  char_specifier: Union[BleakGATTCharacteristic, int, str, uuid.UUID]

tsiens:
> > In function “start_notify“，characteristic can be a string like the function 'stop_notify'
> 
> This should already be handled in the `start_notify()` in `bleak/__init__.py`. Is it not working?
> 
> > I don't know why
> 
> I would like to understand what is actually going on before considering merging these change. E.g. does the same Python script work on other OSes but not on Android. Can we get logs showing what descriptors are actually reported by the device. Are we sure this isn't a defective device like those described in #972?

He doesn't work. I entered a string uuid into the start_nofity and reported an error.
After I modified the p4a/client.py, it was successful

Issue #1532, opened 2024-04-05
tx_power on Android uses wrong getter


dlech:
To be consistent with the other backend APIs, the Android backend should not use `ScanResult.getTxPower()` but `ScanRecord.getTxPowerLevel()` instead, if I'm not mistaken.

_Originally posted by @MarkusPiotrowski in https://github.com/hbldh/bleak/discussions/984#discussioncomment-9023455_

PR #1549, opened 2024-04-28
backends/p4android: fix using wrong tx_power value

dlech:
This was using the advertisement received signal strength indicator instead of the transmitted power level. This is now fixed.

Fixes: https://github.com/hbldh/bleak/issues/1532


MarkusPiotrowski:
Thank you for fixing this. Could you also update the docstring for `AdvertisementData` to make clear to the user that this is actually reading the transmitted power level?

PR #1566, opened 2024-05-07
backends.winrt: add allow_sta() utility function

dlech:
We had lots of users reporting issues with the WinRT backend hanging forever when trying to connect to a device. This was happening because some other imported library was initializing the the main thread to STA which caused some async callback to never be called.

To work around this, we added a check to make sure the main thread is set to MTA rather than STA. Unfortunately, in cases where there is a graphical user interface library being used AND that library is properly interated with asynco so that Bleak runs in the main thread rather than in a background thread, the thread type does need to be STA for the GUI to work. So in those very specific conditions, we need to not raise an exception.

We don't know of a way to detect this automatically, so we added a new utility function `allow_sta()` that the user can call to allow Bleak to run in an STA thread.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,10 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 `Unreleased`_
 =============
 
+### Added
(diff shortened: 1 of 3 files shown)


dlech:
Wasn't actually merged, see #1567.

PR #1571, opened 2024-05-10
backends/winrt: add address acquisition process when advertising data is None

kaz399:
In some situations, device.details has the scan response data only and advertising data is None.

Since the ble address cannot be obtained from device.details.adv this time, the ble address is obtained from the device.details.scan (scan response data).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,11 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 `Unreleased`_
 =============
 
+Changed
+-------
+* Retrieve the BLE address required by ``BleakClientWinRT`` from scan response if advertising is None (WinRT).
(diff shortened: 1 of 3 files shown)


dlech:
Thanks for the PR! 

Can you please add a changelog entry?

Also, I'm wondering if we could write this in a way that is a bit less verbose. Maybe something like this?

```python
data = address_or_ble_device.details
self._device_info = (data.adv or data.scan).bluetooth_address
```

And it would be good to fix the type hint at https://github.com/hbldh/bleak/blob/181467f7d7e09e0a7130962a0f7f8db4510ddc59/bleak/backends/winrt/scanner.py#L53 while we are at it. (Should be `Optional[BluetoothLEAdvertisementReceivedEventArgs]`.)

kaz399:
Thanks for the suggestion for a more concise description!
It is very clear and good.

I modify the type hints and update the changelog.

dlech:
Thanks for updating!

PR #1573, opened 2024-05-17
Update uuids.py

patman15:
Fixed the documentation for UUID format to match the BT standard and thus actual implementation.


dlech:
Thanks!

Issue #1576, opened 2024-05-21
Service UUID filter does not apply to dict of discovered devices.


dlech:
So, service_uuid filtering does work in Windows? 
I try to filter on:
``` service_uuids = ['0000fef3-0000-1000-8000-00805f9b34fb']```
like this:
```
...
service_uuids = ['0000fef3-0000-1000-8000-00805f9b34fb']
# Without filter
async with Scanner(service_uuids=None) as scanner:
    await asyncio.sleep(10)
    self.show_scan_result(
        scanner.discovered_devices_and_advertisement_data
    )
# With filter
async with Scanner(service_uuids=service_uuids) as scanner:
    await asyncio.sleep(10)
    self.show_scan_result(
        scanner.discovered_devices_and_advertisement_data
    )
...
```
(where `show_scan_results` is a method to display the results).
However, I can't see a difference in the devices beeing found, and some of them don't have the service uuid in their AdvertisementData. 
Tested with 0.21.1 and 0.22.1 on Windows 10.

Am I doing something wrong or do I misunderstand the concept of service UUID filtering?

_Originally posted by @MarkusPiotrowski in https://github.com/hbldh/bleak/issues/1534#issuecomment-2123497252_

dlech:
In https://github.com/hbldh/bleak/pull/1559, we missed that the filter should also apply to the list of discovered devices. Currently, it only applies to the detection callback.

PR #1577, opened 2024-05-21
backends: filter discovered devices

dlech:
This is a follow-up to a818521 ("backends/scanner: always filter by service_uuids") to also filter discovered devices by the service UUIDs.

This was overlooked in that change and on Windows actually caused a regression.

Fixes: https://github.com/hbldh/bleak/issues/1576

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,11 @@ Changed
 * Retrieve the BLE address required by ``BleakClientWinRT`` from scan response if advertising is None (WinRT).
 * Changed type hint for ``adv`` attribute of ``bleak.backends.winrt.scanner._RawAdvData``.
 
+Fixed
+-----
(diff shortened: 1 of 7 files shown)


dlech:
Hi @MarkusPiotrowski, could you please test?

MarkusPiotrowski:
Yes, this work now. Out of 8 devices catched during a scan without UUID filtering, only two were found when scanning with service_uuids filter, both containing the selected service UUID.

BTW, is it usual that a device only advertises one service, at max? Although `service_uuids` is a list in `AdvertisementData`, I have never seen a device with more than one service in this list.

dlech:
> BTW, is it usual that a device only advertises one service, at max?

Probably, given the small size of the advertising data. I suppose someone might want to scan for two different types of devices at the same time in some cases?

PR #1581, opened 2024-05-24
backends/winrt: don't throw exception for properly configured GUI apps

dlech:
In commit 4a653e6 ("backends/winrt: raise exception when trying to scan with STA") we added a check to raise an exception when trying to scan when PyWinRT set the aparatment model to STA. However, properly working GUI apps will have the apartment model set to STA but Bleak will still work because there is something pumping the Windows message loop.

We don't want to raise an exception in this case to avoid breaking working apps. We can improve the test by checking if the current thread is actually pumping the message loop by scheduling a callback via a the win32 SetTimeout function. If the callback is called, then we know that the message loop is being pumped. If not, then we probably are not going to get async callbacks from the WinRT APIs and we raise an exception in this case.

### DIFF
--- a/bleak/backends/winrt/util.py
+++ b/bleak/backends/winrt/util.py
@@ -1,8 +1,17 @@
+import asyncio
 import ctypes
+from ctypes import wintypes
 from enum import IntEnum
 from typing import Tuple
 
 from ...exc import BleakError
+
+
+def _check_result(result, func, args):
+    if not result:
+        raise ctypes.WinError()
+
+    return args
 
 
 def _check_hresult(result, func, args):
@@ -10,6 +19,29 @@ def _check_hresult(result, func, args):
         raise ctypes.WinError(result)
 
     return args
+
+
+# not defined in wintypes
+if ctypes.sizeof(ctypes.c_long) == ctypes.sizeof(ctypes.c_void_p):
+    _UINT_PTR = ctypes.c_ulong
+elif ctypes.sizeof(ctypes.c_longlong) == ctypes.sizeof(ctypes.c_void_p):
+    _UINT_PTR = ctypes.c_ulonglong
+
+# https://learn.microsoft.com/en-us/windows/win32/api/winuser/nc-winuser-timerproc
+_TIMERPROC = ctypes.WINFUNCTYPE(
+    None, wintypes.HWND, _UINT_PTR, wintypes.UINT, wintypes.DWORD
+)
+
+# https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-settimer
+_SetTimer = ctypes.windll.user32.SetTimer
+_SetTimer.restype = _UINT_PTR
+_SetTimer.argtypes = [wintypes.HWND, _UINT_PTR, wintypes.UINT, _TIMERPROC]
+_SetTimer.errcheck = _check_result
+
+# https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-killtimer
+_KillTimer = ctypes.windll.user32.KillTimer
+_KillTimer.restype = wintypes.BOOL
+_KillTimer.argtypes = [wintypes.HWND, wintypes.UINT]
 
 
 # https://learn.microsoft.com/en-us/windows/win32/api/combaseapi/nf-combaseapi-cogetapartmenttype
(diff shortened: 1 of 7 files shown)


dlech:
Hi @MarkusPiotrowski, could you please test this with your BeeWare app?

I've tested it with your tkinter example and it seems to work.

MarkusPiotrowski:
In my setup, it fails in line 138 of ...bleak\backends\winrt\util.py with AttributeError: module 'asyncio' has no attribute 'timeout'

The `asyncio.timeout()` context manager was added in Python 3.11 (I'm on 3.10). I don't know your dependency policy, but requiring 3.11 is possibly quite strict?

bleak/backends/winrt/util.py
```diff
@@ -1,17 +1,49 @@
...
 
     return args
 
 
+# not defined in wintypes
+if ctypes.sizeof(ctypes.c_long) == ctypes.sizeof(ctypes.c_void_p):
+    _UINT_PTR = ctypes.c_ulong
+elif ctypes.sizeof(ctypes.c_longlong) == ctypes.sizeof(ctypes.c_void_p):
+    _UINT_PTR = ctypes.c_ulonglong
+
```
dlech, line 29: This breaks doc builds because ctypes is automocked and both cases are false.

bleak/backends/winrt/util.py
```diff
@@ -60,28 +92,61 @@ def _get_apartment_type() -> Tuple[_AptType, _AptQualifierType]:
...
+async def assert_mta() -> None:
     """
     Asserts that the current apartment type is MTA.
 
     Raises:
         BleakError: If the current apartment type is not MTA.
 
     .. versionadded:: 0.22
+
+    .. versionchanged:: unreleased
```
dlech, line 104: Need to add description of what changed

bleak/backends/winrt/util.py
```diff
@@ -60,28 +92,61 @@ def _get_apartment_type() -> Tuple[_AptType, _AptQualifierType]:
...
+    # set a timer to see if we get a callback to ensure the windows event loop
+    # is running
+    timer = _SetTimer(None, 1, 0, callback)
+
+    try:
+        async with asyncio.timeout(1):
+            await event.wait()
+    except asyncio.TimeoutError:
+        raise BleakError(
+            "Thread is configured for Windows GUI but callbacks are not working. Suspect PyWin32 unwanted side effects."
```
dlech, line 142: Could only mention PyWin32 if one of it's modules is in `sys.modules`.

docs/troubleshooting.rst
```diff
@@ -202,7 +202,7 @@ thread then call ``allow_sta()`` before calling any other Bleak APis::
         pass
```
dlech, line 203: Should mention setting `sys.????` as workaround to prevent pywin32 from initializing the thread in the first place.

bleak/backends/winrt/util.py
```diff
@@ -60,28 +92,61 @@ def _get_apartment_type() -> Tuple[_AptType, _AptQualifierType]:
...
+        )
+    else:
+        # if the windows event loop is running, we assume it is going to keep
+        # running and we don't need to check again
+        setattr(assert_mta, "_allowed", True)
+    finally:
+        _KillTimer(None, timer)
 
 
 def allow_sta():
```
dlech, line 152: Need to change implementation of `uninitialize_sta()` now that `assert_mta()` is async

bleak/backends/winrt/util.py
```diff
@@ -60,28 +92,61 @@ def _get_apartment_type() -> Tuple[_AptType, _AptQualifierType]:
...
+    # have to keep a reference to the callback or it will be garbage collected
+    # before it is called
+    callback = _TIMERPROC(wait_event)
+
+    # set a timer to see if we get a callback to ensure the windows event loop
+    # is running
+    timer = _SetTimer(None, 1, 0, callback)
+
+    try:
+        async with asyncio.timeout(1):
```
dlech, line 138: Need async timeout compatible with older Python.

MarkusPiotrowski:
Yes, this is now working for me, both with `allow_sta()` set or not.

dlech:
Great, thanks for testing! I assume this is a good enough solution for you?

dlech:
Tested with Kivy as well.

PR #1604, opened 2024-06-24
Remove debug logging check

Siecje:
Unless this is hyper optimization this check done inside the log. https://github.com/python/cpython/blob/3.12/Lib/logging/__init__.py#L1517


dlech:
These were found to have a measurable difference in performance when profiling. (Mentioned in https://github.com/hbldh/bleak/pull/1376)

So I would need to see hard evidence to the contrary before making these changes.

dlech:
Only if profiling shows a measured improvement.

PR #1605, opened 2024-06-29
Provide a local definition of user32.setTimer, user32.KillTimer and ole32.CoGetApartmentType on Windows

bramd:
By overriding the type definition of SetTimer, we prevent None being passed as the callback parameter. This causes problems when integrating Bleak in larger Windows applications that call SetTimer in this way.

By defining the function type definition locally, we don't override the global definition so existing code that uses None as a callback keeps working.

   guidelines in that document.
    - OS X / macOS >= 10.11

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -23,6 +23,7 @@ Contributors
 * Robbe Gaeremynck <robbe.gaeremynck@outlook.com>
 * David Johansen <davejohansen@gmail.com>
 * JP Hutchins <jphutchins@gmail.com>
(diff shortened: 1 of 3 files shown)


dlech:
Thanks for the PR! I agree that since Bleak is a library, it shouldn't be modifying Python globals like this. We should fix the other functions as well.

bramd:
@dlech Good idea to do the other functions as well. I've updated the PR to do this and also made sure to run black this time.

dlech:
Thanks for updating!

Could you add CHANGELOG entry too?

PR #1632, opened 2024-08-21
Index BlueZ advertisement_callbacks by adapter

bdraco:
- Instead of doing a linear search of all the callbacks to find the one for the adapter, store them in a dict so the adapter path can be looked up
- Remove unused arg from _run_advertisement_callbacks

- [x] If the pull request adds functionality, the docs should be updated.
- [x] Modify the `CHANGELOG.rst`, describing your changes as is specified by the
   guidelines in that document.
- [x] The pull request should work for Python 3.8+ on the following platforms:
    - OS X / macOS >= 10.11
- [x] Squash all your commits on your PR branch, if the commits are not solving
- [x] Feel free to add your name as a contributor to the `AUTHORS.rst` file!

### DIFF
--- a/bleak/backends/bluezdbus/manager.py
+++ b/bleak/backends/bluezdbus/manager.py
@@ -10,6 +10,7 @@ import asyncio
 import contextlib
 import logging
 import os
+from collections import defaultdict
 from typing import (
     Any,
     Callable,
@@ -52,22 +53,6 @@ Args:
     arg0: The D-Bus object path of the device.
     arg1: The D-Bus properties of the device object.
 """
-
-
-class CallbackAndState(NamedTuple):
-    """
-    Encapsulates an :data:`AdvertisementCallback` and some state.
-    """
-
-    callback: AdvertisementCallback
-    """
-    The callback.
-    """
-
-    adapter_path: str
-    """
-    The D-Bus object path of the adapter associated with the callback.
-    """
 
 
 DevicePropertiesChangedCallback = Callable[[Optional[Any]], None]
@@ -193,7 +178,9 @@ class BlueZManager:
         # map of characteristic d-bus object paths to set of descriptor d-bus object paths
         self._descriptor_map: Dict[str, Set[str]] = {}
 
-        self._advertisement_callbacks: List[CallbackAndState] = []
+        self._advertisement_callbacks: defaultdict[str, List[AdvertisementCallback]] = (
+            defaultdict(list)
+        )
         self._device_removed_callbacks: List[DeviceRemovedCallbackAndState] = []
         self._device_watchers: Dict[str, Set[DeviceWatcher]] = {}
         self._condition_callbacks: Dict[str, Set[DeviceConditionCallback]] = {}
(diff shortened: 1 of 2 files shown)


bleak/backends/bluezdbus/manager.py
```diff
@@ -1035,22 +1029,16 @@ def _parse_msg(self, message: Message) -> None:
                                     message_path, new_value
                                 )
 
-    def _run_advertisement_callbacks(
-        self, device_path: str, device: Device1, changed: Iterable[str]
-    ) -> None:
+    def _run_advertisement_callbacks(self, device_path: str, device: Device1) -> None:
```
bdraco, line 1032: changed has been unused for a while

dlech: Can you split that out to a separate patch?

Siecje (approved):

bleak/backends/bluezdbus/manager.py
```diff
@@ -10,12 +10,12 @@
 import contextlib
 import logging
 import os
+from collections import defaultdict
```
Siecje, line 13: Move to line 10 to keep imports sorted?

dlech: CI lint job is passing, so I think this is already sorted correctly according to `isort`.

bleak/backends/bluezdbus/manager.py
bdraco: https://github.com/hbldh/bleak/pull/1633

bleak/backends/bluezdbus/manager.py
```diff
@@ -1035,22 +1029,16 @@ def _parse_msg(self, message: Message) -> None:
...
             device: The current D-Bus properties of the device.
-            changed: A list of properties that have changed since the last call.
         """
-        for callback, adapter_path in self._advertisement_callbacks:
-            # filter messages from other adapters
-            if adapter_path != device["Adapter"]:
-                continue
-
+        adapter_path = device["Adapter"]
+        for callback in self._advertisement_callbacks[adapter_path]:
```
bdraco, line 1041: Average [time complexity](https://wiki.python.org/moin/TimeComplexity) should now be: `O(callbacks)` instead of `O(adapters * callbacks)`

dlech: Do you have before and after flamegraphs or something like that showing the difference?

bdraco:
> https://github.com/hbldh/bleak/pull/1632#discussion_r1726006211

Running with 6 Bluetooth adapters its ~28% less time spent in `_run_advertisement_callbacks` processing ~8000 messages per minute as measured with cProfile.  

It can be roughly benchmarked with

```
import timeit
from collections import defaultdict


def callback():
    pass


advertisement_callbacks: defaultdict[str, list] = defaultdict(list)

test_adapter = "/org/bluez/hci9"

for i in range(10):
    advertisement_callbacks[f"/org/bluez/hci{i}"] = [callback]

advertisement_callbacks_orig = [
    (callback, adapter_path)
    for callback, adapter_path in advertisement_callbacks.items()
]


def new():
    for callback in advertisement_callbacks[test_adapter]:
        pass


def orig():
    for callback, adapter_path in advertisement_callbacks_orig:
        if adapter_path != test_adapter:
            continue


print(timeit.timeit(orig, number=1000000))
print(timeit.timeit(new, number=1000000))
```

orig
```
0.1742814159952104
```
new
```
0.042215333989588544
```

bleak/backends/bluezdbus/manager.py
bdraco: https://github.com/hbldh/bleak/pull/1632#issuecomment-2303402217

dlech:
Thanks!

bdraco:
Thanks

PR #1640, opened 2024-09-24
Add support for Python 3.13

cdce8p:
Update the pin for `pyobjc-core` to add support for Python 3.13. While at it, added `3.12` and `3.13-dev` to the list of test environments.

While testing, I saw that the action runs failed on my fork due to old versions. Updated those as well.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,10 +22,10 @@ classifiers = [
 "Issues" =  "https://github.com/hbldh/bleak/issues"
 
 [tool.poetry.dependencies]
-python = ">=3.8,<3.13"
+python = ">=3.8,<3.14"
 async-timeout = { version = ">= 3.0.0, < 5", python = "<3.11" }
 typing-extensions = { version = ">=4.7.0", python = "<3.12" }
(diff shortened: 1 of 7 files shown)


bdraco (approved):
Thanks

There are a lot of downstream projects waiting for 3.13 support

pyproject.toml
```diff
@@ -22,10 +22,10 @@ classifiers = [
 "Issues" =  "https://github.com/hbldh/bleak/issues"
 
 [tool.poetry.dependencies]
-python = ">=3.8,<3.13"
+python = ">=3.8,<3.14"
 async-timeout = { version = ">= 3.0.0, < 5", python = "<3.11" }
 typing-extensions = { version = ">=4.7.0", python = "<3.12" }
-pyobjc-core = { version = "^10.0", markers = "platform_system=='Darwin'" }
+pyobjc-core = { version = "^10.3", markers = "platform_system=='Darwin'" }
```
dlech, line 28: If we are bumping this one, we should bump all other pyobjc packages as well.

dlech:
Thanks!

cdce8p:
@dlech It would be awesome if you could do a new release. With that we could start updating downstream packages as well.

Issue #1621, opened 2024-07-22
Android permissions


vascolp:
### Hello again!

* bleak version:  0.22.2
* Python version: python 3.12.3
* Operating System: Android 14

### Description

I have an android app that works very well with bleak (good start :-)) 
The app does not require location data. It only uses BLE to scan and connect to LEGO Hubs.
But bleak forces permissions `ACCESS_FINE_LOCATION, ACCESS_COARSE_LOCATION, ACCESS_BACKGROUND_LOCATION`.

Location permitions are a problem if you intend to do an android app that can be used by children.
Android offers the hability to "strongly assert that my app does not use location" (check point 5 [here](https://developer.android.com/develop/connectivity/bluetooth/bt-permissions)).
According to the same page, I am also convinced that in some situations `ACCESS_BACKGROUND_LOCATION` is also not needed. My case in particular.

### Testing

I tried all this by making a copy of bleak and just commenting the call to the `request_permissions()` (and following `await`) and using these buildozer permissions (check the special case for `BLUETOOTH_SCAN`):
```
android.permissions = android.permission.INTERNET,android.permission.BLUETOOTH_ADMIN,(name=android.permission.BLUETOOTH_SCAN;usesPermissionFlags=neverForLocation),android.permission.BLUETOOTH_CONNECT
```
**And everything works fine!!** 

Is there a chance that you could find a way to omit the call to `request_permissions()` if the user wants that? Some parameter valid only for android, or something like that?

I would appreciate it very much!

Thanks in advance!
VascoLP

vascolp:
Ok, here are some notes about android permission versions, that might be of interest to someone.

1. Did not investigate versions prior to android 10, but probably it works like android 10.
2. On android versions 10 and 11, (APIs 29 and 30) bleak will need `ACCESS_COARSE_LOCATION` and `ACCESS_FINE_LOCATION`. In some cases (when there is a background service) you might also need `ACCESS_BACKGROUND_LOCATION`.
3. On android 12 and above (API 31 and above), _if your app does not use locations_, you can use the ability to "strongly assert that my app does not use location" (check it [here](https://developer.android.com/develop/connectivity/bluetooth/bt-permissions)), so the `*LOCATION` permissions are not needed. 
4. If you want to give support for android 10 to 14 in a kivy application, and avoid location permissions where possible, you have to give some permissions based on API levels. With the following example, an APP will ask for location permissions in android up to 11, but not on android 12 or above.
Here is a configuration example:
[...]

PR #1642, opened 2024-09-27
add new permissions from android API 31

tomaszduda23:

   guidelines in that document.
    - OS X / macOS >= 10.11


dlech:
We've been discussing this in https://github.com/hbldh/bleak/issues/1621.

My preference would be to move this to a separate function in addition to adding the require permissions for newer Android versions.

tomaszduda23:
> We've been discussing this in #1621.
> 
> My preference would be to move this to a separate function in addition to adding the require permissions for newer Android versions.

done

bleak/backends/p4android/scanner.py
```diff
@@ -61,6 +61,41 @@ def __init__(
...
+                        BleakError("User denied access to " + str(permissions))
+                    )
+                )
+        if not kwargs.get("permissions"):
+            permissions = [
+                Permission.ACCESS_FINE_LOCATION,
+                Permission.ACCESS_COARSE_LOCATION,
+                Permission.ACCESS_BACKGROUND_LOCATION,
+            ]
+            VERSION = autoclass('android.os.Build$VERSION')
```
dlech, line 84: To be consistent with other constants in the android backend, I would expect this to be defined in the `defs` module.

bleak/backends/scanner.py
```diff
@@ -307,6 +307,15 @@ def set_scanning_filter(self, **kwargs) -> None:
         """
         raise NotImplementedError()
 
+    @abc.abstractmethod
+    async def request_permissions(self, **kwargs) -> None:
```
dlech, line 311: Since this only applies to Android, I don't think we should make this a common function to all platforms.

Also, it is not specifically related to the `BleakScanner`, so should not be a method on that object.

Making it a standalone function in `p2andoid.utils` seems like the logical place to put it.

examples/kivy/main.py
```diff
@@ -41,7 +41,9 @@ async def example(self):
         while self.running:
             try:
                 self.line("scanning")
-                scanned_devices = await bleak.BleakScanner.discover(1)
+                scanner = bleak.BleakScanner()
+                await scanner.request_permissions()
```
dlech, line 45: This should only be called once at the start of the program. Also, it needs to be guarded with an if or in a try/except so that the program can run on platforms other than android.

examples/kivy/main.py
```diff
@@ -41,7 +41,9 @@ async def example(self):
...
+                scanned_devices = scanner.discover(1)
```
dlech, line 46: `discover()` is not an instance method, so we should not be creating a `BleakScanner()` instance.

bleak/backends/p4android/scanner.py
```diff
@@ -61,6 +61,41 @@ def __init__(
...
+                loop.call_soon_threadsafe(
+                    permission_acknowledged.set_result, grantResults
+                )
+            else:
+                loop.call_soon_threadsafe(
+                    permission_acknowledged.set_exception(
+                        BleakError("User denied access to " + str(permissions))
+                    )
+                )
+        if not kwargs.get("permissions"):
```
dlech, line 78: Instead of kwargs, I would just make a named arg that defaults the default permissions.

dlech:
Also, we will need a changelog entry explaining that this is a breaking change.

bleak/backends/p4android/scanner.py
tomaszduda23: done

bleak/backends/scanner.py
tomaszduda23: done

examples/kivy/main.py
tomaszduda23: done

bleak/backends/p4android/scanner.py
tomaszduda23: done

examples/kivy/main.py
tomaszduda23: done

tomaszduda23:
updated according to review notes

tomaszduda23:
github should have started itself after pushing new version. It takes way to much time to fix all lint.

dlech:
Don't worry about the linting, I can fix it when I get back to this.

PR #1644, opened 2024-09-28
add auto_connect for android

tomaszduda23:

   guidelines in that document.
    - OS X / macOS >= 10.11

### DIFF
--- a/bleak/backends/p4android/client.py
+++ b/bleak/backends/p4android/client.py
@@ -49,6 +49,7 @@ class BleakClientP4Android(BaseBleakClient):
         self.__adapter = kwargs.get("adapter", kwargs.get("device", None))
         self.__gatt = None
         self.__mtu = 23
+        self.__auto_connect = kwargs.get("auto_connect", False)
 
     def __del__(self):
         if self.__gatt is not None:
@@ -84,7 +85,7 @@ class BleakClientP4Android(BaseBleakClient):
             dispatchApi=self.__device.connectGatt,
             dispatchParams=(
                 defs.context,
-                False,
+                self.__auto_connect,
                 self.__callbacks.java,
                 defs.BluetoothDevice.TRANSPORT_LE,
             ),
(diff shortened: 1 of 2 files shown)


dlech:
Can you describe the intended use for this feature? Typically, with Bleak, we always scan and then connect.

tomaszduda23:
There are restriction on scanning in android. It works each time when device is awake. When screen is off it fails in active and passive mode. Without this connect fails randomly.

dlech:
I'm not sure I fully understand. You are saying that the restriction is with scanning, but the change is in `BleakClient`, not `BleakScanner`.

tomaszduda23:
I'm sorry form mixing this up. Those are two separate things.
1. If you know device address already it is possible to connect without scanning.
a) E.g. you can have stored last device which was used in settings. Than you can connect without scanning.
b) Or you can scan when screen is on and connect from time to time when screan is off. When using auto_connect = True android blocks until device is available.
2. BleakScanner.discovery fails if scan is done from service when screen is off. It is not related to this PR. I just mention this for reference. That might be caused 
https://developer.android.com/reference/android/bluetooth/le/ScanSettings#SCAN_MODE_LOW_POWER

dlech:
> 1. When using auto_connect = True android blocks until device is available.

This is how all other OSes work, so it sounds like we want `auto_connect = True` all of the time (with a timeout). This way, Bleak works the same on all platforms.

tomaszduda23:
It seems that both are needed depending on remote device configuration https://stackoverflow.com/questions/40156699/which-correct-flag-of-autoconnect-in-connectgatt-of-ble. If remote device do not advertise it would never be available I guess. Doc says. 
```
autoConnect
boolean: Whether to directly connect to the remote device (false)
or to automatically connect as soon as the remote device becomes available (true).
```
My understanding is that true waits for advertisement from remote. false try to immediately connect to device which was "bonded" already. I can be wrong though.

tomaszduda23:
It seems to go back to https://github.com/bluez/bluez/commit/0f40779ee29133a35f9d9f420bd1cdda426787f6#diff-f936cc254469fe4d4a2cbca5892ebaeebe3e03fd1e7bfdc86d8a998477b20bbcR1946

tomaszduda23:
any chance to get it merged?

PR #1663, opened 2024-10-10
Escape demonstration bytes in BleakClient docstring.

MrSmoer:
The bytes inside the example weren't properly escaped. Some IDE's threw an error, readthedocs showed this as empty.

### DIFF
--- a/bleak/__init__.py
+++ b/bleak/__init__.py
@@ -726,7 +726,7 @@ class BleakClient:
         data: Buffer,
         response: bool = None,
     ) -> None:
-        """
+        r"""
         Perform a write operation on the specified GATT characteristic.
 
         There are two possible kinds of writes. *Write with response* (sometimes


dlech:
I think the bug is in the documentation renderer. 

`b"\x00\x01\x02\x03"` is perfectly valid Python code. The intention it to have a byte array with 4 bytes, namely 0, 1, 2, 3. With your suggested change, it would be a byte array with 16 bytes.

```python
>>> b"\x00\x01\x02\x03" == bytes([0, 1, 2, 3])
True
>>> b"\\x00\\x01\\x02\\x03" == bytes([0, 1, 2, 3])
False
>>> b"\\x00\\x01\\x02\\x03" == bytes([92, 120, 48, 48, 92, 120, 48, 49, 92, 120, 48, 50, 92, 120, 48, 51])
True
```

MrSmoer:
This is not supposed to be python code, this is supposed to be a string containing python code (it is wrapped with  `"""` which creates a multiline string).
In the context of the docstring you want it to be python code when it is displayed, eg. for displaying as `\x01` you want it to be made of: a backslash character, an x character, a zero character, a one character,
If you want to create a backslash character to display as a backslash inside a string in python, you need to escape it by putting another backslash. 
Before, the `\x` inside the string tells python to make the next character be the the next two hexdigits as ascii code.
Ascii code of 0-3 is in order: Null character, Start of Heading, Start of Text, End of Text. So this creates a string containing those four control characters, which don't properly display as anything - not on the readthedocs, not in my IDE's type-hint and not in my Terminal.

```
>>> print("""await client.write_gatt_char(MY_CHAR_UUID, b"\\x00\\x01\\x02\\x03", response=True)""")
await client.write_gatt_char(MY_CHAR_UUID, b"\x00\x01\x02\x03", response=True)
>>> print("""await client.write_gatt_char(MY_CHAR_UUID, b"\x00\x01\x02\x03", response=True)""")
await client.write_gatt_char(MY_CHAR_UUID, b"", response=True)
>>> "\x61" # 0x61 is 97 in decimal which decodes to 'a' in ascii 
'a'
>>> 

```

MrSmoer:
Besides, thank you for creating this library! 
It just saved me after fighting with another python BLE library for hours. Someone told me about bleak and it just worked.

dlech:
> This is not supposed to be python code, this is supposed to be a string containing python code

Ah, of course! My mistake.

dlech:
Instead of adding escapes, could we change it to a raw string instead (`r"""`)? Then people could still copy-paste the example from the source code.

MrSmoer:
Sure, an r-string is definitely better

dlech:
Thanks!

PR #1670, opened 2024-10-18
Add troubleshooting step for "Not connected" problems on Raspberry Pi.

denravonska:
After a lot of debugging we have discovered that disabling wifi fixes a _lot_ of sporadic "Not connected" errors we have been seeing in CI. This has been seen improving connectivity on a Raspberry Pi 4 and a Rock Pi 3C.

### DIFF
--- a/docs/troubleshooting.rst
+++ b/docs/troubleshooting.rst
@@ -119,6 +119,33 @@ crash with an ``ImportError`` similar to::
 To fix the error, change the name of the script to something other than ``bleak.py``.
 
 ----------
+Linux Bugs
+----------
+
+Occasional "Not connected" errors on Raspberry Pi
+=================================================
+
+If you are using a Raspberry Pi and are seeing occasional "Not connected" errors,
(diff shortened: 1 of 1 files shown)


docs/troubleshooting.rst
```diff
@@ -118,6 +118,33 @@ crash with an ``ImportError`` similar to::
...
 To fix the error, change the name of the script to something other than ``bleak.py``.
 
+----------
+Linux Bugs
+----------
+
+Occasional "Not connected" errors on Raspberry Pi
+=================================================
+
+If you are using a Raspberry Pi and are seeing occasional "Not connected" errors,
```
dlech, line 128: I assume this is using the built-in WiFi and Bluetooth on the newer RPis? Probably would be helpful to clarify that in the text.

denravonska: If what I'm suspecting is correct you're right, it shouldn't affect USB connected radios as they don't use the same chip. I'll see if it can be rephrased.

dlech:
Looks good, thanks!

Issue #1060, opened 2022-10-03
add BleakAdapter class


dlech:
This is a feature request to add a new `BleakAdapter` class to provide access to the system Bluetooth adapter.

This class should have a property for getting the adapter state (powered on, powered off, not present, not authorized) and allow registering a callback for when the adapter state changes. This could fix #320.

It could also have a method to get already connected Bluetooth devices. This could fix #367.

Additionally, it could provide an eaiser way to specify which adapter to use on Linux (instead of passing the adapter arg to each BleakScanner and BleakClient).

Creation of this object will necessarily be async due to the OS APIs that would need to be used, so this should have an async `default_adapter()` class (or static) method to get the adapter (currently all OSes other than Linux only allow one Bluetooth adapter) instead of an `__init__` method.

JPHutchins:
This seems to be a requirement in the case that Windows 10 has bonded with the device, I will be investigating.

bdraco:
Feel free to lift any code that might be useful from https://github.com/bluetooth-devices/bluetooth-adapters

JPHutchins:
Another resource for future reference https://github.com/IanSavchenko/BleLab.  I've confirmed that I can (only) access bonded devices with this tool.

dlech:
Partial progress on this is being done in #1966.

PR #1681, opened 2024-11-01
Added support for Google´s Bumble Bluetooth Controller stack

vChavezB:
The backend supports direct use with [Bumble](https://google.github.io/bumble/). The HCI Controller is managed by the Bumble stack and the transport layer can be defined by the user (e.g. VHCI, Serial, TCP, android-netsim).

Use cases for this backend are:

1. Bluetooth Functional tests without Hardware. Example of Bluetooth stacks that
   support virtualization are Android Emulator and Zephyr RTOS.
2. Virtual connection between different HCI Controllers that are not in the same
    radio network (virtual or physical).
    
    
To enable this backend the env. variable `BLEAK_BUMBLE` must be set.

At the moment this is an experimental backend and would like to get some feedback from users and the devs of bleak :)


vChavezB:
Fixed some linting and doc issues.

JPHutchins:
It looks like you may not have updated the lock file.  Please run `poetry lock --no-update`, commit, and push. thanks!

bleak/backends/bumble/client.py
```diff
@@ -0,0 +1,293 @@
...
+
+        """
+        if not self.is_connected:
+            raise BleakError("Not connected")
+
+        if self.services is not None:
+            return self.services
+
+        new_services = BleakGATTServiceCollection()
+        self._peer = Peer(self._connection)
```
JPHutchins, line 128: This is a type error; self._peer already has static type None - is it something OR nothing or is it always guaranteed to be something?  Should it be a mutable variable that is global to the class or should it be an immutable variable local to the function it's used in?

bleak/backends/bumble/client.py
```diff
@@ -0,0 +1,293 @@
...
+            char_specifier (BleakGATTCharacteristic, int, str or UUID): The characteristic to read from,
+                specified by either integer handle, UUID or directly by the
+                BleakGATTCharacteristic object representing it.
+
+        Returns:
+            (bytearray) The read data.
+
+        """
+        if not isinstance(char_specifier, BleakGATTCharacteristic):
+            characteristic = self.services.get_characteristic(char_specifier)
```
JPHutchins, line 164: Type error - services can be None.

bleak/backends/bumble/client.py
```diff
@@ -0,0 +1,293 @@
...
+        """Perform read operation on the specified GATT descriptor.
+
+        Args:
+            handle (int): The handle of the descriptor to read from.
+
+        Returns:
+            (bytearray) The read data.
+
+        """
+        descriptor = self.services.get_descriptor(handle)
```
JPHutchins, line 184: Type error; services can be None

bleak/backends/bumble/client.py
```diff
@@ -0,0 +1,293 @@
...
+        if not descriptor:
+            raise BleakError("Descriptor {} was not found!".format(handle))
```
JPHutchins, line 186: Prefer f strings

bleak/backends/bumble/client.py
```diff
@@ -0,0 +1,293 @@
...
+        val = await descriptor.obj.read_value()
+        logger.debug("Read Descriptor {0} : {1}".format(handle, val))
```
JPHutchins, line 188: Prefer f-strings

JPHutchins (commented):
Looks amazing!

Starting with a first round of type safety. While Bleak does not have 100% type coverage, I won't support any new python - in any context - that does not have 100% type coverage. That's just my opinion, and I'm happy to PR for it if it's too much of a lift.

I'll stay tuned to this and test it out on HW ASAP!

JPHutchins:
Re: the overall API.  If this is merged, I think that it should be importable as a separate class.  BleakClient can remain as an OS-selected class, but BumbleClient should be separately importable.  BumbleClient would be fully compatible with BleakClient - that is, they implement the same interface, Protocol, ABC, whatever you want to call it.

bleak/backends/bumble/__init__.py
```diff
@@ -0,0 +1,32 @@
+# SPDX-License-Identifier: MIT
+# Copyright (c) 2024 Victor Chavez
+"""Bumble backend."""
+
+from bumble.controller import Controller
+from bumble.link import LocalLink
+from bumble.transport import open_transport
+
+transports = {}
```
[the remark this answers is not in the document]
JPHutchins: Like why are these serialized to strings?  OMG, it parses them:

```python
scheme, *tail = name.split(':', 1)
    spec = tail[0] if tail else None
    metadata = None
    if spec:
        # Metadata may precede the spec
        if spec.startswith('['):
            metadata_str, *tail = spec[1:].split(']')
            spec = tail[0] if tail else None
            metadata = dict([entry.split('=') for entry in metadata_str.split(',')])
```
🤦‍♀️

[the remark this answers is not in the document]
JPHutchins: When Guido left Google he should have taken Python with him...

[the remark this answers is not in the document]
vChavezB: I think bumble uses directly strings. See [here](https://github.com/google/bumble/blob/b78f8951430e8b4c0f9ef14818666dbede97b009/bumble/transport/__init__.py#L106).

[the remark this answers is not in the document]
JPHutchins: Actually there are hidden imports here which is going to be made for compiling binaries...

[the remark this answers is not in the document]
JPHutchins: > I think bumble uses directly strings. See [here](https://github.com/google/bumble/blob/b78f8951430e8b4c0f9ef14818666dbede97b009/bumble/transport/__init__.py#L106).

Agreed - that API is bumming me out.  It should deserialize to some well defined types and then pass those to open the transport.  I see it was done this way to allow user input from the command line into this function.

vChavezB:
I agree, the Bumble client is largely OS independent (except for the transports that use the Linux HCI device). So it could be imported separately.

Thanks for the feedback on type safety! It really helps to remove ambiguity in Python and help other developers understand the intent of the code. I will apply the suggestions you mentioned.

JPHutchins:
> I agree, the Bumble client is largely OS independent (except for the transports that use the Linux HCI device). So it could be imported separately.
> 
> Thanks for the feedback on type safety! It really helps to remove ambiguity in Python and help other developers understand the intent of the code. I will apply the suggestions you mentioned.

Glad you like it! I was annoyed when I first saw it being required in projects but now I'm a "kool-aid drinker" 🤣!  It does in fact eliminate an entire class of runtime bugs that Python applications tend to suffer from.

bleak/backends/bumble/client.py
```diff
@@ -0,0 +1,293 @@
...
+
+        Returns:
+            Boolean representing connection status.
+
+        """
+        await start_transport(self._adapter)
+        await self._dev.power_on()
+        await self._dev.connect(self.address)
+
+        self.services = await self.get_services()
```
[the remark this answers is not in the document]
vChavezB: Good catch.  I checked the other backends and only p4android returns a boolean. Seems that an issue should be raised separately to verify the complete type safety of the project.

vChavezB:
I have applied some of the suggestions from @JPHutchins. I will also check the typesafety of the changes in this PR.

bleak/backends/bumble/__init__.py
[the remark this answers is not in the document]
vChavezB: I checked the Transports from bumble and I am not sure its trivial. The implementation specific transports are defined locally inside the function that opens each transport. 

Example for the TCP Server 

https://github.com/google/bumble/blob/b78f8951430e8b4c0f9ef14818666dbede97b009/bumble/transport/tcp_server.py#L65

This requires a PR to the bumble project to allow users to access the Transport implementations.

What I could do in the meantime is add enums to the strings of the transport with argument `adapter` and a separate argument string for the specific parameters that the transport needs.

bleak/backends/bumble/scanner.py
```diff
@@ -0,0 +1,124 @@
...
+        await start_transport(self._adapter)
+        await self.device.power_on()
+        await self.device.start_scanning(active=self._scan_active)
+
+    async def stop(self) -> None:
+        await self.device.stop_scanning()
+        await self.device.power_off()
+
+    def set_scanning_filter(self, **kwargs) -> None:
+        # Implement scanning filter setup
```
dlech, line 123: Is this a TODO?

bleak/backends/bumble/utils.py
```diff
@@ -0,0 +1,9 @@
+# SPDX-License-Identifier: MIT
+# Copyright (c) 2024 Victor Chavez
+def uuid_bytes_to_str(byte_data: bytes) -> str:
```
dlech, line 3: Why not use the `uuid` module from the standard library for this?

docs/backends/bumble.rst
```diff
@@ -0,0 +1,116 @@
...
+
+Use cases for this backend are:
+
+1. Bluetooth Functional tests without Hardware. Example of Bluetooth stacks that
+   support virtualization are |android_emulator| and |zephyr|.
+2. Connection of HCI Controllers that are not in the same radio network (virtual or physical).
+
+.. note::
+
+    To select this backend set the environmental variable ``BLEAK_BUMBLE``.
```
dlech, line 16: The environment variable should contain the transport info.

The idea is that users should just have to set the environment variable and use any Bleak script with the Bumble backend without having to modify any code.

docs/backends/bumble.rst
```diff
@@ -0,0 +1,116 @@
+Bumble backend
+===============
+
+This backend adds support for the |bumble| Bluetooth Controller Stack from Google.
+The backend enables support of multiple |bumble_transport| to communicate with
+a physical or virtual HCI controller.
+
+Use cases for this backend are:
```
dlech, line 8: I think the serial/USB use case is worth mentioning here too. (Probably as number 1)

pyproject.toml
```diff
@@ -38,6 +38,7 @@ bleak-winrt = { version = "^1.2.0", markers = "platform_system=='Windows'", pyth
 "winrt-Windows.Foundation.Collections" = { version = "^2", markers = "platform_system=='Windows'", python = ">=3.12" }
 "winrt-Windows.Storage.Streams" = { version = "^2", markers = "platform_system=='Windows'", python = ">=3.12" }
 dbus-fast = { version = ">=1.83.0, < 3", markers = "platform_system == 'Linux'" }
+bumble = "^0.0.201"
```
dlech, line 41: I think this should be an optional dependency (i.e. only installed if `pip install bleak[bumble]`). Otherwise it is going to bring in a lot of other dependencies that most people don't need.

tests/bleak/backends/bumble/test_scanner.py
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env python
+
+"""Tests for `bleak.backends.bumble` package, specifically scanning and advertising functionality."""
+
+import asyncio
+from queue import Queue
```
dlech, line 6: Why not asyncio Queue?

dlech (commented):
This looks really nice!

 I think it will be useful for people who have devices that don't work so well with the various OS Bluetooth stacks. And it could be useful for providing a test backend for anyone that needs some serious testing.

>  I think that it should be importable as a separate class.

I disagree with this suggestion. I would like to be able to run any script with or without the Bumble backend, so no code changes should be required to use the Bumble backend. I also don't want a bunch of features being added that aren't supported on other backends, so having a separate class should not be necessary.

docs/backends/android.rst
```diff
@@ -1,7 +1,7 @@
 Android backend
 ===============
 
-For an example of building an android bluetooth app, see 
+For an example of building an android bluetooth app, see
```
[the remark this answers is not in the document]
vChavezB: Yes, it was not my intention to modify code that is not related to this PR. Must have occured while I was formatting my changes.

bleak/backends/bumble/client.py
```diff
@@ -0,0 +1,309 @@
...
+        Returns:
+            Boolean representing connection status.
+
+        """
+        await start_transport(self._adapter)
+        await self._dev.power_on()
+        await self._dev.connect(self.address)
+
+        self.services: BleakGATTServiceCollection = await self.get_services()
+        return self.is_connected
```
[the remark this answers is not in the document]
vChavezB: According to the docs of the BleakClient class it should be

`Boolean representing connection status.`

Thanks for the clarification.

bleak/backends/bumble/client.py
```diff
@@ -0,0 +1,309 @@
...
+
+    async def connect(self, **kwargs) -> bool:
+        """Connect to the specified GATT server.
+
+        Returns:
+            Boolean representing connection status.
+
+        """
+        await start_transport(self._adapter)
+        await self._dev.power_on()
```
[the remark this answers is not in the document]
vChavezB: If there a BleakAdapter is proposed I am glad to use it instead.

bleak/backends/bumble/client.py
```diff
@@ -0,0 +1,309 @@
...
+    ) -> None:
+        """
+        Perform a write operation on the specified GATT characteristic.
+
+        Args:
+            characteristic: The characteristic to write to.
+            data: The data to send.
+            response: If write-with-response operation should be done.
+        """
+        await characteristic.obj.write_value(data)
```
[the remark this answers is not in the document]
vChavezB: I will check how this is done in Bumble.

According to the bluetooth spec. you are referring to this ?

grafik

bleak/backends/bumble/client.py
```diff
@@ -0,0 +1,309 @@
...
+
+        """
+        descriptor = self.services.get_descriptor(handle)
+        if not descriptor:
+            raise BleakError(f"Descriptor {handle} was not found!")
+        await descriptor.obj.write_value(data)
+        logger.debug(f"Write Descriptor {handle} : {data}")
+
+    def __notify_handler(self, characteristic: BleakGATTCharacteristic, value):
+        for sub in self._subs[characteristic.obj.uuid]:
```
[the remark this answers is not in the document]
vChavezB: True, thanks for the info!

bleak/backends/bumble/scanner.py
```diff
@@ -0,0 +1,124 @@
...
+SERVICE_UUID_TYPES = [
+    AdvertisingData.COMPLETE_LIST_OF_128_BIT_SERVICE_CLASS_UUIDS,
+    AdvertisingData.COMPLETE_LIST_OF_16_BIT_SERVICE_CLASS_UUIDS,
+    AdvertisingData.COMPLETE_LIST_OF_32_BIT_SERVICE_CLASS_UUIDS,
+    AdvertisingData.INCOMPLETE_LIST_OF_16_BIT_SERVICE_CLASS_UUIDS,
+    AdvertisingData.INCOMPLETE_LIST_OF_32_BIT_SERVICE_CLASS_UUIDS,
+    AdvertisingData.INCOMPLETE_LIST_OF_128_BIT_SERVICE_CLASS_UUIDS,
+]
+
+SCANNER_MAC_ADDR = "F0:F1:F2:F3:F4:F5"
```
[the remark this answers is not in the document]
vChavezB: Just an arbitrary value.

bleak/backends/bumble/scanner.py
```diff
@@ -0,0 +1,124 @@
...
+        self.device.host.controller = Controller("Scanner", link=get_link())
+        self.device.listener = self
+        self._scan_active = scanning_mode == "active"
+        self._adapter: BumbleTransport = kwargs.get("adapter", get_default_transport())
+
+    def on_advertisement(self, advertisement: Advertisement):
+        service_uuids: List[str] = []
+        service_data: Dict[str, AdvertisingDataObject] = {}
+        local_name = advertisement.data.get(AdvertisingData.COMPLETE_LOCAL_NAME)
+        if not local_name:
```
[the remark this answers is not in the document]
vChavezB: I will double check again what do I get if the device does not advertise the complete local name

bleak/backends/bumble/scanner.py
vChavezB: Yes, I have not checked in detail what type of filters can you set with Bumble

bleak/backends/bumble/client.py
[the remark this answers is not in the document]
vChavezB: Nevermind, it was really easy. Bumble provides an argument for this.

bleak/backends/bumble/scanner.py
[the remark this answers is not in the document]
vChavezB: I checked again, and the local name might be empty. I was reading the Bluetooth Core Spec v6.0 and from what I understood is that it is desirable to include the local name but not a must.

bleak/backends/bumble/scanner.py
```diff
@@ -0,0 +1,124 @@
...
+        manuf_data = advertisement.data.get(AdvertisingData.MANUFACTURER_SPECIFIC_DATA)
+        for uuid_type in SERVICE_UUID_TYPES:
+            adv_uuids = advertisement.data.get(uuid_type)
+            if adv_uuids is None:
+                continue
+            if not isinstance(adv_uuids, list):
+                continue
+            for uuid in adv_uuids:
+                if uuid not in service_uuids:
+                    service_uuids.append(str(uuid))
```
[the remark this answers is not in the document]
vChavezB: The type of `adv_uuids` is a List of `Bumble.Core.UUID`.  Which is not a normalized uuid but a custom class from bumble.

I have to convert it to string to get the string representation, otherwise the implementation specific class for a UUID from Bumble is used.

bleak/backends/bumble/utils.py
vChavezB: I should have made this more clear. The function should convert data bytes from the custom UUID class from bumble (bumble.core.UUID). This is not a standard uuid and requires a conversion.

Example

- **Shortened 16 bit UUID**: 0x1800
- **bumble UUID string representation**: 'UUID-16:1800 (Generic Access)'
- **bumble UUID bytes representation:** b'\x00\x18'

bleak/backends/bumble/client.py
```diff
@@ -0,0 +1,309 @@
...
+
+        Returns:
+            Boolean representing connection status.
+
+        """
+        await start_transport(self._adapter)
+        await self._dev.power_on()
+        await self._dev.connect(self.address)
+
+        self.services: BleakGATTServiceCollection = await self.get_services()
```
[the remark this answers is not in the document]
vChavezB: What do you mean by disconnect? 

should I do wrap it around a `try` statement ? 

or do you mean if the BLE device does not have any services we should disconnect?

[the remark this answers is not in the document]
vChavezB: I did not find any similar logic in the backends that refer to check if the service fails. The only one is the WinRT where this is raised: https://github.com/hbldh/bleak/blob/e01e2640994b99066552b6161f84799712c396fa/bleak/backends/winrt/client.py#L796

vChavezB:
Resolved some of the reviews pending is:

- Wait for Answer to https://github.com/hbldh/bleak/pull/1681#discussion_r1827566918
- Wait for PR for Bleak adapter ? https://github.com/hbldh/bleak/issues/1060
- Wait for PR (?) to get Rid of BleakGattChar, BleakGatt Svc , see https://github.com/hbldh/bleak/pull/1681#discussion_r1827081869 
- Wait for comment on detailed description of USB/Serial use case. See https://github.com/hbldh/bleak/pull/1681#discussion_r1827470772
- Define exactly expected use cases for bumble https://github.com/hbldh/bleak/pull/1681#discussion_r1827656973

JPHutchins:
> I disagree with this suggestion. I would like to be able to run any script with or without the Bumble backend, so no code changes should be required to use the Bumble backend. I also don't want a bunch of features being added that aren't supported on other backends, so having a separate class should not be necessary.

I think that this works for me.  My application can simply specify the Bumble backend, right?

https://github.com/hbldh/bleak/blob/e01e2640994b99066552b6161f84799712c396fa/bleak/__init__.py#L517

Obviously I cannot ask users to be setting environment variables 🤣.

Note to self: investigate import times due to use of runtime vs compile time OS checks: https://github.com/hbldh/bleak/blob/e01e2640994b99066552b6161f84799712c396fa/bleak/backends/client.py#L250-L274

dlech:
> My application can simply specify the Bumble backend, right?

Yes, if you have an application that requires the Bumble backend and doesn't work with other backend, that is fine. 



> Obviously I cannot ask users to be setting environment variables 🤣.

Of course! But your application could set the environment variable itself after it starts too. But the `backend` arg seems like the more straightforward way to do it.

vChavezB:
Well, in this case, if for bleak we want an easy way for users to replace the backend with bumble, then a variable like `BLEAK_BUMBLE_TRANSPORT` can be defined (or whatever name is more convenient). 

And then if we don't want the env variable, then the user can pass the backend option 👍 

I can then mention these two options in the docs.

docs/backends/bumble.rst
vChavezB: I will add an env. variable to define use of bumble + transport options. In addition, I will just mention in the docs that if a specific backend must be selected then it can be changed over the `BleakClient` argument `backend` 👍

bleak/backends/bumble/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def get_default_transport() -> BumbleTransport:
+    return BumbleTransport(TransportScheme.TCP_SERVER, "127.0.0.1:1234")
+
+
+async def start_transport(transport: BumbleTransport) -> None:
+    transport_cmd = str(transport)
+    if transport_cmd not in transports.keys():
+        transports[transport_cmd] = await open_transport(transport_cmd)
```
JPHutchins, line 55: I appreciate all the work done to make this less error-prone; but I'd like to understand why the user cannot simply import their "open transport function" - like `open_serial_transport`, `open_udp_transport`, etc.

It's taking the burden off of Bleak to maintain. Because the function is awaitable and has side effects, it would likely be passed as a partial or lambda (a closure):

```python

async def start_transport(
    open_transport_func: partial[Coroutine[None, None, Transport]]
) -> Transport:
    await open_transport_func()
    ...

my_transport = partial(open_serial_transport, "my serial transport args")

...

await start_transport(my_transport)
```

The specific transport functions are still poorly typed with serialized strings as arguments, but at least now they could change upstream without Bleak having any responsibility to maintain them.

vChavezB: I am using the `open_transport` instead of specific `open_transport_xyz` because all the bumble examples and applications use this function. So I assumed this was the official Bumble way 🐝 .

vChavezB: Couldnt we map the possible transports functions to an Enum, then the user does not need to pass a partial or callback but just an enum plus string arguments.

JPHutchins: Yes, that would be cool, but still requires maintenance here when upstream changes.  The closure is perhaps confusing, and an alternative would be that start_transport takes the open_transport_func and spec as args.  If one of the open transport functions upstream changes its number of args, this could be a problem, but it feels like they are using serialized strings with the goal of the function signature being general.

vChavezB: My intention with the current implementation is to reuse `open_transport` with the string syntax used in bumble's examples and applications. This provides familiarity for bumble users. The specific string syntax for each transport is documented [here](https://google.github.io/bumble/transports). If I tell the user to import the `open_transport_xyz`, additional documentation needs to be written telling them to look in the source of [`bumble/transport`](https://github.com/google/bumble/tree/main/bumble/transport) and list which functions are available.

I understand your point about upstream changes. Bumble is an experimental project, so there is no guarantee of stability. Since it is still a work in progress, I could also keep this backend as an external module to avoid maintenance costs due to the maturity of the bumble project.

JPHutchins: > If I tell the user to import the `open_transport_xyz`, additional documentation needs to be written telling them to look in the source of [`bumble/transport`](https://github.com/google/bumble/tree/main/bumble/transport) and list which functions are available.

That's a good point and I agree it's a good justification to use the serialized strings!

pyproject.toml
```diff
@@ -38,6 +38,9 @@ bleak-winrt = { version = "^1.2.0", markers = "platform_system=='Windows'", pyth
 "winrt-Windows.Foundation.Collections" = { version = "^2", markers = "platform_system=='Windows'", python = ">=3.12" }
 "winrt-Windows.Storage.Streams" = { version = "^2", markers = "platform_system=='Windows'", python = ">=3.12" }
 dbus-fast = { version = ">=1.83.0, < 3", markers = "platform_system == 'Linux'" }
+[project.optional-dependencies]
+bumble = ["bumble=^0.0.201"]
+
```
JPHutchins, line 43: Bumble does not seem to be conforming to semver.  Unless no release from 0.0.0 -> 0.0.202 broke an API or added a feature 🤣.

What this means is that users of Bumble should lock it instead of trusting its version.  Please change this to

```toml
[project.optional-dependencies]
bumble = ["bumble==0.0.201"]
```

Or 0.0.202 if I feel it is not a breaking change.

The `^=` syntax is shorthand for "update until a breaking version" - but it only works if the package uses semver.

vChavezB: Good catch, 201 minor patches. Yes that makes more sense

bleak/backends/bumble/characteristic.py
```diff
@@ -0,0 +1,82 @@
...
+        props = [flag for flag in Characteristic.Properties if flag in obj.properties]
+        self.__props: Final = [str(prop) for prop in props]
+        self.__svc: Final = svc
+        uuid = bumble_uuid_to_str(obj.uuid)
+        self.__uuid: Final = normalize_uuid_str(uuid)
+
+    @property
+    def service_uuid(self) -> str:
+        """The uuid of the Service containing this characteristic"""
+        return self.__svc.uuid
```
JPHutchins, line 36: Type error, got UUID, expected str

bleak/backends/bumble/client.py
```diff
@@ -0,0 +1,322 @@
...
+            characteristic = self.services.get_characteristic(char_specifier)
+            if not characteristic:
+                raise BleakCharacteristicNotFoundError(char_specifier)
+        else:
+            characteristic = char_specifier
+
+        value = await self._peer.read_characteristics_by_uuid(characteristic.obj.uuid)
+        value = bytearray(value[0])
+        logger.debug(f"Read Characteristic {characteristic.uuid} : {value}")
+        return value
```
JPHutchins, line 221: So, the peer returns a list of bytes, then we extract the first one and return it? What is in the rest of the list?

There are type errors here because value is mutated. Best to use a different variable name for the data that's extracted from the list of bytes.

bleak/backends/bumble/client.py
```diff
@@ -0,0 +1,322 @@
...
+
+        Args:
+            handle (int): The handle of the descriptor to read from.
+
+        Returns:
+            (bytearray) The read data.
+
+        """
+        descriptor = self.services.get_descriptor(handle)
+        if not descriptor:
```
JPHutchins, line 234: Avoid truthiness/falseyness: `if descriptor is None:`

bleak/backends/bumble/descriptor.py
```diff
@@ -0,0 +1,36 @@
...
+from bleak.backends.descriptor import BleakGATTDescriptor
+
+
+class BleakGATTDescriptorBumble(BleakGATTDescriptor):
+    """GATT Descriptor implementation for Bumble backend."""
+
+    def __init__(self, obj: DescriptorProxy, characteristic: CharacteristicProxy):
+        super(BleakGATTDescriptorBumble, self).__init__(obj)
+        self.obj: Final = obj
+        self.__characteristic: Final = characteristic
```
JPHutchins, line 16: Replace all `__` with `_`.

bleak/backends/bumble/descriptor.py
```diff
@@ -0,0 +1,36 @@
...
+
+    @property
+    def characteristic_handle(self) -> int:
+        """handle for the characteristic that this descriptor belongs to"""
+        return self.__characteristic.handle
+
+    @property
+    def uuid(self) -> str:
+        """UUID for this descriptor"""
+        return str(self.obj.uuid)
```
JPHutchins, line 31: "DescriptorProxy" has no attribute "uuid"

bleak/backends/bumble/scanner.py
```diff
@@ -0,0 +1,125 @@
...
+            for uuid in adv_uuids:  # type: UUID
+                if uuid not in service_uuids:
+                    service_uuids.append(str(uuid))
+
+        for service_data in advertisement.data.get_all(AdvertisingData.SERVICE_DATA):
+            service_uuid, data = service_data
+            service_data[str(service_uuid)] = data
+
+        advertisement_data = AdvertisementData(
+            local_name=local_name,
```
JPHutchins, line 93:
```
Argument "local_name" to "AdvertisementData" has incompatible type "list[UUID] | tuple[UUID, bytes] | bytes | str | int | tuple[int, int] | tuple[int, bytes] | Appearance | None"; expected "str | None"
```

bleak/backends/bumble/scanner.py
```diff
@@ -0,0 +1,125 @@
...
+            manufacturer_data=manuf_data,
```
JPHutchins, line 94:
```
Argument "manufacturer_data" to "AdvertisementData" has incompatible type "list[UUID] | tuple[UUID, bytes] | bytes | str | int | tuple[int, int] | tuple[int, bytes] | Appearance | None"; expected "dict[int, bytes]"
```

bleak/backends/bumble/scanner.py
```diff
@@ -0,0 +1,125 @@
...
+            service_data=service_data,
```
JPHutchins, line 95:
```
Argument "service_data" to "AdvertisementData" has incompatible type "dict[str, list[UUID] | tuple[UUID, bytes] | bytes | str | int | tuple[int, int] | tuple[int, bytes] | Appearance]"; expected "dict[str, bytes]"
```

bleak/backends/bumble/scanner.py
```diff
@@ -0,0 +1,125 @@
...
+            service_uuids=service_uuids,
+            tx_power=advertisement.tx_power,
+            rssi=advertisement.rssi,
+            platform_data=(None, None),
+        )
+
+        device = self.create_or_update_device(
+            str(advertisement.address),
+            local_name,
```
JPHutchins, line 104:
```
Argument 2 to "create_or_update_device" of "BaseBleakScanner" has incompatible type "list[UUID] | tuple[UUID, bytes] | bytes | str | int | tuple[int, int] | tuple[int, bytes] | Appearance | None"; expected "str"
```

bleak/backends/bumble/scanner.py
```diff
@@ -0,0 +1,125 @@
...
+                continue
+            if not isinstance(adv_uuids, list):
+                continue
+            for uuid in adv_uuids:  # type: UUID
+                if uuid not in service_uuids:
+                    service_uuids.append(str(uuid))
+
+        for service_data in advertisement.data.get_all(AdvertisingData.SERVICE_DATA):
+            service_uuid, data = service_data
+            service_data[str(service_uuid)] = data
```
JPHutchins, line 90: Incompatible types in assignment (expression has type "list[UUID] | tuple[UUID, bytes] | bytes | str | int | tuple[int, int] | tuple[int, bytes] | Appearance", variable has type "dict[str, list[UUID] | tuple[UUID, bytes] | bytes | str | int | tuple[int, int] | tuple[int, bytes] | Appearance]")

JPHutchins: Service data expects `dict[str, bytes`, but the for loop mutates the dict and fills some of the values with strings. Is this intentional?

bleak/backends/bumble/scanner.py
JPHutchins, line 1: Most of the type errors can be resolved by avoiding the reuse of variable names for objects that are different types.

bleak/backends/bumble/service.py
```diff
@@ -0,0 +1,40 @@
...
+from bleak.backends.bumble.utils import bumble_uuid_to_str
+from bleak.backends.service import BleakGATTService
+
+
+class BleakGATTServiceBumble(BleakGATTService):
+    """GATT Characteristic implementation for the Bumble backend."""
+
+    def __init__(self, obj: ServiceProxy):
+        super().__init__(obj)
+        self.__characteristics = []
```
JPHutchins, line 18: Replace all `__` with `_`.  Add type annotation.

bleak/backends/bumble/service.py
```diff
@@ -0,0 +1,40 @@
...
+    def characteristics(self) -> List[CharacteristicProxy]:
+        """List of characteristics for this service"""
+        return self.__characteristics
+
+    def add_characteristic(self, characteristic: CharacteristicProxy):
+        """Add a :py:class:`~BleakGATTCharacteristicWinRT` to the service.
+
+        Should not be used by end user, but rather by `bleak` itself.
+        """
+        self.__characteristics.append(characteristic)
```
JPHutchins, line 40: Argument 1 of "add_characteristic" is incompatible with supertype "BleakGATTService"; supertype defines the argument type as "BleakGATTCharacteristic"

JPHutchins (commented):
Looking good! Will be testing with an nrf52840dk momentarily!

Overall, a lot of time will be saved by getting mypy working in your editor.

bleak/backends/bumble/utils.py
```diff
@@ -0,0 +1,20 @@
...
+    Convert a native Bumble UUID to a string representation.
+    Reason for this conversion is that the string representation (__str__) done in bumble
+    is not a standard UUID string representation.
+    In addition, the byte representation from data member uuid_bytes of bumble.core.UUID
+    is represented in reverse order.
+    Example:
+        Shortened 16 bit UUID: 0x1800Bumble UUID class
+        bumble UUID string representation: 'UUID-16:1800 (Generic Access)'
+        bumble UUID bytes representation: b'\x00\x18'
+    """
```
JPHutchins, line 17: This docstring isn't getting parsed correctly, needs to be updated to conform to google spec.

bleak/backends/bumble/scanner.py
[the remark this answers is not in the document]
JPHutchins: Ha!  In that case, `if local_name is None or local_name == "":` if we want to be pedantic. Truthiness does not effectively convey intent.

bleak/backends/bumble/descriptor.py
```diff
@@ -0,0 +1,36 @@
...
+
+from bleak.backends.descriptor import BleakGATTDescriptor
+
+
+class BleakGATTDescriptorBumble(BleakGATTDescriptor):
+    """GATT Descriptor implementation for Bumble backend."""
+
+    def __init__(self, obj: DescriptorProxy, characteristic: CharacteristicProxy):
+        super(BleakGATTDescriptorBumble, self).__init__(obj)
+        self.obj: Final = obj
```
JPHutchins, line 15: Update to:

```python
    def __init__(self, obj: DescriptorProxy, characteristic: CharacteristicProxy):
        super().__init__(obj)
```

which is the same, AFAICT.

JPHutchins:
@vChavezB @dlech What do you think about committing FW images for some common platforms? If you'd rather not, we can create a repo that is solely for that.

dlech:
> What do you think about committing FW images for some common platforms?

Sounds like a job for a separate repo.

JPHutchins:
Build HCI USB sample for NRF52840DK:
```
west build -b nrf52840dk/nrf52840 zephyr/samples/bluetooth/hci_usb
```

Flash
```
west flash
```

NRF52840DK device shows up!

image

Confirmed `poetry run python -m examples.discover` running normally on the regular BT adapter. Now will set env variables to test the Bumble backend!

```
$env:BLEAK_BUMBLE="usb:2FE3:0008"
```

```
poetry run python -m examples.discover
scanning for 5 seconds, please wait...
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:\Users\jp\repos\bleak\examples\discover.py", line 51, in <module>
    asyncio.run(main(args))
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.12_3.12.2032.0_x64__qbz5n2kfra8p0\Lib\asyncio\runners.py", line 194, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.12_3.12.2032.0_x64__qbz5n2kfra8p0\Lib\asyncio\runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.12_3.12.2032.0_x64__qbz5n2kfra8p0\Lib\asyncio\base_events.py", line 687, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "C:\Users\jp\repos\bleak\examples\discover.py", line 20, in main
    devices = await BleakScanner.discover(
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\jp\repos\bleak\bleak\__init__.py", line 320, in discover
    async with cls(**kwargs) as scanner:
               ^^^^^^^^^^^^^
  File "C:\Users\jp\repos\bleak\bleak\__init__.py", line 158, in __aenter__
    await self._backend.start()
  File "C:\Users\jp\repos\bleak\bleak\backends\bumble\scanner.py", line 115, in start
    await start_transport(self._adapter)
  File "C:\Users\jp\repos\bleak\bleak\backends\bumble\__init__.py", line 112, in start_transport
    transports[transport_cmd] = await open_transport(transport_cmd)
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\jp\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\Local\pypoetry\Cache\virtualenvs\bleak-dSQ9Yv1H-py3.12\Lib\site-packages\bumble\transport\__init__.py", line 93, in open_transport
    transport = await _open_transport(scheme, spec)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\jp\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\Local\pypoetry\Cache\virtualenvs\bleak-dSQ9Yv1H-py3.12\Lib\site-packages\bumble\transport\__init__.py", line 165, in _open_transport
    return await open_usb_transport(spec)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\jp\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\Local\pypoetry\Cache\virtualenvs\bleak-dSQ9Yv1H-py3.12\Lib\site-packages\bumble\transport\usb.py", line 456, in open_usb_transport
    raise TransportInitError('device not found')
bumble.transport.common.TransportInitError: device not found
```

Tried lower case hex for VID/PID as well.

Reference is here: https://google.github.io/bumble/transports/usb.html

OK, updated to:
```
$env:BLEAK_BUMBLE="usb:0"
```

And now I get:
```
scanning for 5 seconds, please wait...
!!! failed to open USB device: LIBUSB_ERROR_NOT_SUPPORTED [-12]
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:\Users\jp\repos\bleak\examples\discover.py", line 51, in <module>
    asyncio.run(main(args))
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.12_3.12.2032.0_x64__qbz5n2kfra8p0\Lib\asyncio\runners.py", line 194, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.12_3.12.2032.0_x64__qbz5n2kfra8p0\Lib\asyncio\runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.12_3.12.2032.0_x64__qbz5n2kfra8p0\Lib\asyncio\base_events.py", line 687, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "C:\Users\jp\repos\bleak\examples\discover.py", line 20, in main
    devices = await BleakScanner.discover(
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\jp\repos\bleak\bleak\__init__.py", line 320, in discover
    async with cls(**kwargs) as scanner:
               ^^^^^^^^^^^^^
  File "C:\Users\jp\repos\bleak\bleak\__init__.py", line 158, in __aenter__
    await self._backend.start()
  File "C:\Users\jp\repos\bleak\bleak\backends\bumble\scanner.py", line 115, in start
    await start_transport(self._adapter)
  File "C:\Users\jp\repos\bleak\bleak\backends\bumble\__init__.py", line 112, in start_transport
    transports[transport_cmd] = await open_transport(transport_cmd)
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\jp\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\Local\pypoetry\Cache\virtualenvs\bleak-dSQ9Yv1H-py3.12\Lib\site-packages\bumble\transport\__init__.py", line 93, in open_transport
    transport = await _open_transport(scheme, spec)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\jp\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\Local\pypoetry\Cache\virtualenvs\bleak-dSQ9Yv1H-py3.12\Lib\site-packages\bumble\transport\__init__.py", line 165, in _open_transport
    return await open_usb_transport(spec)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\jp\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\Local\pypoetry\Cache\virtualenvs\bleak-dSQ9Yv1H-py3.12\Lib\site-packages\bumble\transport\usb.py", line 536, in open_usb_transport
    device = found.open()
             ^^^^^^^^^^^^
  File "C:\Users\jp\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\Local\pypoetry\Cache\virtualenvs\bleak-dSQ9Yv1H-py3.12\Lib\site-packages\usb1\__init__.py", line 2055, in open
    mayRaiseUSBError(libusb1.libusb_open(self.device_p, byref(handle)))
  File "C:\Users\jp\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\Local\pypoetry\Cache\virtualenvs\bleak-dSQ9Yv1H-py3.12\Lib\site-packages\usb1\__init__.py", line 127, in mayRaiseUSBError
    __raiseUSBError(value)
  File "C:\Users\jp\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\Local\pypoetry\Cache\virtualenvs\bleak-dSQ9Yv1H-py3.12\Lib\site-packages\usb1\__init__.py", line 119, in raiseUSBError
    raise __STATUS_TO_EXCEPTION_DICT.get(value, __USBError)(value)
usb1.USBErrorNotSupported: LIBUSB_ERROR_NOT_SUPPORTED [-12]
```

Installing Zadig 🤣 ...

Tried libusb-win32 driver - discover script hangs until I unplug the board.

Tried libusbK driver - same

Tried WinUSB - same

Could test from linux first, for example, then try Windows.

vChavezB:
> Looking good! Will be testing with an nrf52840dk momentarily!
> 
> Overall, a lot of time will be saved by getting mypy working in your editor.

Thanks for the review. I will definitely check again with the typesafety extensions in pycharm. It would be nice to get some field tests to check if everything works. So far I have only tested with virtual devices with Zephyr posix.

JPHutchins:
> Tried lower case hex for VID/PID as well.

Would you believe that the PID is 000B but I'd set 0008 every time? 😭. Will run through again ASAP.

JPHutchins:
> Would you believe that the PID is 000B but I'd set 0008 every time? 😭. Will run through again ASAP.

OK, picking up where I left off. Starting by trying to revert the drivers to Windows defaults.  It's on libwdi now, after unisntalling and restart.

Set logging and correct USB VID:

```
$env:BLEAK_LOGGING=1
$env:BLEAK_BUMBLE="usb:2fe3:000B"
```

```
poetry run python -m examples.discover
scanning for 5 seconds, please wait...
```

Still hang. No Ctrl-C.  Can exit by removing the Bumble device.

Call stack reveals it's maybe doing some USB stuff (I didn't include whole stack)

```
!!! IN transfer not completed: status=4
Exception ignored on calling ctypes callback function: <bound method USBTransfer.__callbackWrapper of <class 'usb1.USBTransfer'>>
  File "C:\Users\jp\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\Local\pypoetry\Cache\virtualenvs\bleak-dSQ9Yv1H-py3.12\Lib\site-packages\usb1\__init__.py", line 327, in __callbackWrapper
    callback(self)
  File "C:\Users\jp\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\Local\pypoetry\Cache\virtualenvs\bleak-dSQ9Yv1H-py3.12\Lib\site-packages\bumble\transport\usb.py", line 280, in transfer_callback
    self.loop.call_soon_threadsafe(self.on_transport_lost)
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.12_3.12.2032.0_x64__qbz5n2kfra8p0\Lib\asyncio\base_events.py", line 840, in call_soon_threadsafe
    self._check_closed()
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.12_3.12.2032.0_x64__qbz5n2kfra8p0\Lib\asyncio\base_events.py", line 541, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed
!!! IN transfer not completed: status=5
Exception ignored on calling ctypes callback function: <bound method USBTransfer.__callbackWrapper of <class 'usb1.USBTransfer'>>
  File "C:\Users\jp\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\Local\pypoetry\Cache\virtualenvs\bleak-dSQ9Yv1H-py3.12\Lib\site-packages\usb1\__init__.py", line 327, in __callbackWrapper
    callback(self)
  File "C:\Users\jp\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\Local\pypoetry\Cache\virtualenvs\bleak-dSQ9Yv1H-py3.12\Lib\site-packages\bumble\transport\usb.py", line 280, in transfer_callback
    self.loop.call_soon_threadsafe(self.on_transport_lost)
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.12_3.12.2032.0_x64__qbz5n2kfra8p0\Lib\asyncio\base_events.py", line 840, in call_soon_threadsafe
    self._check_closed()
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.12_3.12.2032.0_x64__qbz5n2kfra8p0\Lib\asyncio\base_events.py", line 541, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed
Exception in thread Thread-1 (run):
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.12_3.12.2032.0_x64__qbz5n2kfra8p0\Lib\threading.py", line 1075, in _bootstrap_inner
    self.run()
```

JPHutchins:
So, it looks like Bumble uses a flexible event callback system. Unfortunately, they're using strings again, but here's the magic string: "advertisement".  Found in source as

```python
self.emit('advertisement', advertisement)
```

So it seems like intended usage may be with a closure, like this:

```python
    async def start(self) -> None:

        def on_advertisement(advertisement: Advertisement):
            logger.debug(f"Received advertisement: {advertisement}")

            service_uuids: List[str] = []
            service_data: Dict[str, AdvertisingDataObject] = {}
            local_name = advertisement.data.get(AdvertisingData.COMPLETE_LOCAL_NAME)
            if not local_name:
                local_name = advertisement.data.get(
                    AdvertisingData.SHORTENED_LOCAL_NAME
                )
            manuf_data = advertisement.data.get(
                AdvertisingData.MANUFACTURER_SPECIFIC_DATA
            )
            for uuid_type in SERVICE_UUID_TYPES:
                adv_uuids = advertisement.data.get(uuid_type)
                if adv_uuids is None:
                    continue
                if not isinstance(adv_uuids, list):
                    continue
                for uuid in adv_uuids:  # type: UUID
                    if uuid not in service_uuids:
                        service_uuids.append(str(uuid))

            for service_data in advertisement.data.get_all(
                AdvertisingData.SERVICE_DATA
            ):
                service_uuid, data = service_data
                service_data[str(service_uuid)] = data

            advertisement_data = AdvertisementData(
                local_name=local_name,
                manufacturer_data=manuf_data,
                service_data=service_data,
                service_uuids=service_uuids,
                tx_power=advertisement.tx_power,
                rssi=advertisement.rssi,
                platform_data=(None, None),
            )

            device = self.create_or_update_device(
                str(advertisement.address),
                local_name,
                {},
                advertisement_data,
            )
            self.call_detection_callbacks(device, advertisement_data)

        self.device.on("advertisement", on_advertisement)

        await start_transport(self._adapter)
        await self.device.power_on()
        await self.device.start_scanning(active=self._scan_active)
```

That said, I still have no success. After adding some logging to bumble and pyee I can confirm that no events are being emitted, so I'm not really sure where to look next.

JPHutchins:
Aw, I see this was supposed to work with multiple inheritance of `Device.Listener`.  I would favor dependency injection instead, for clarity: `self._listener = OurListenerImpl()`.

JPHutchins:
Can confirm the scanner (Zephyr FW) is working well with Bumble's example!

```
python -m examples.run_scanner usb:2fe3:000B
```

JPHutchins:
I think this can be refactored to align with the example more closely: https://github.com/google/bumble/blob/5e959d638e6a9c99e62536d0a3472cf4e6616ccf/examples/run_scanner.py#L36-L78

Specifically, class consrtuctors in Python async should not do any IO. So we can wait until some action is taken to open connections and do async IO, for example.

JPHutchins:
The following scanner implementation is working well for me.  Obviously would need to be adapted to fit with the rest of the API, but it's nice to see it running.

```python
# SPDX-License-Identifier: MIT
# Copyright (c) 2024 Victor Chavez

import logging
import os
from typing import Dict, Final, List, Literal, Optional

from bumble.core import AdvertisingData, AdvertisingDataObject
from bumble.device import Advertisement, Device
from bumble.hci import Address
from bumble.transport import open_transport

from bleak.backends.scanner import (
    AdvertisementData,
    AdvertisementDataCallback,
    BaseBleakScanner,
)

logger = logging.getLogger(__name__)

SERVICE_UUID_TYPES = [
    AdvertisingData.COMPLETE_LIST_OF_128_BIT_SERVICE_CLASS_UUIDS,
    AdvertisingData.COMPLETE_LIST_OF_16_BIT_SERVICE_CLASS_UUIDS,
    AdvertisingData.COMPLETE_LIST_OF_32_BIT_SERVICE_CLASS_UUIDS,
    AdvertisingData.INCOMPLETE_LIST_OF_16_BIT_SERVICE_CLASS_UUIDS,
    AdvertisingData.INCOMPLETE_LIST_OF_32_BIT_SERVICE_CLASS_UUIDS,
    AdvertisingData.INCOMPLETE_LIST_OF_128_BIT_SERVICE_CLASS_UUIDS,
]

# Arbitrary BD_ADDR for the scanner device
SCANNER_BD_ADDR = "F0:F1:F2:F3:F4:F5"


class BleakScannerBumble(BaseBleakScanner):
    """
    Interface for Bleak Bluetooth LE Scanners

    Args:
        detection_callback:
            Optional function that will be called each time a device is
            discovered or advertising data has changed.
        service_uuids:
            Optional list of service UUIDs to filter on. Only advertisements
            containing this advertising data will be received.
    Keyword Args:
        adapter (BumbleTransport): Bumble transport adapter to use.
    """

    def __init__(
        self,
        detection_callback: Optional[AdvertisementDataCallback],
        service_uuids: Optional[List[str]],
        scanning_mode: Literal["active", "passive"],
        **kwargs,
    ):
        super().__init__(detection_callback, service_uuids)

        self._scanning_mode: Final = scanning_mode

        self._device: Optional[Device] = None

    async def on_connection(self, connection):
        pass

    async def start(self) -> None:

        def on_advertisement(advertisement: Advertisement):
            logger.debug(f"Received advertisement: {advertisement}")

            service_uuids: List[str] = []
            service_data: Dict[str, AdvertisingDataObject] = {}
            local_name = advertisement.data.get(AdvertisingData.COMPLETE_LOCAL_NAME)
            if not local_name:
                local_name = advertisement.data.get(
                    AdvertisingData.SHORTENED_LOCAL_NAME
                )
            manuf_data = advertisement.data.get(
                AdvertisingData.MANUFACTURER_SPECIFIC_DATA
            )
            for uuid_type in SERVICE_UUID_TYPES:
                adv_uuids = advertisement.data.get(uuid_type)
                if adv_uuids is None:
                    continue
                if not isinstance(adv_uuids, list):
                    continue
                for uuid in adv_uuids:
                    if uuid not in service_uuids:
                        service_uuids.append(str(uuid))

            for service_data in advertisement.data.get_all(
                AdvertisingData.SERVICE_DATA
            ):
                service_uuid, data = service_data
                service_data[str(service_uuid)] = data

            advertisement_data = AdvertisementData(
                local_name=local_name,
                manufacturer_data=manuf_data,
                service_data=service_data,
                service_uuids=service_uuids,
                tx_power=advertisement.tx_power,
                rssi=advertisement.rssi,
                platform_data=(None, None),
            )

            device = self.create_or_update_device(
                str(advertisement.address),
                local_name,
                {},
                advertisement_data,
            )
            self.call_detection_callbacks(device, advertisement_data)

        hci_transport: Final = await open_transport(os.environ["BLEAK_BUMBLE"])

        self._device = Device.with_hci(
            "scanner_dev",
            Address(SCANNER_BD_ADDR),
            hci_transport.source,
            hci_transport.sink,
        )

        self._device.on("advertisement", on_advertisement)

        await self._device.power_on()
        await self._device.start_scanning(active=self._scanning_mode == "active")

    async def stop(self) -> None:
        if self._device is None:
            raise RuntimeError("Scanner not started")

        await self._device.stop_scanning()
        await self._device.power_off()

        self._device = None

    def set_scanning_filter(self, **kwargs) -> None:
        # Implement scanning filter setup
        pass
```

JPHutchins:
There's another wrinkle with Bumble. The library defines `open_transport` as an async function, but this is not necessary in the case of opening a USB transport - that is, `open_usb_transport` has no `await`s.

Defining the USBTransport class in a factory function is a problem.

vChavezB:
Thanks for the tests and suggestions to make it work with usb 👍

vChavezB:
> The following scanner implementation is working well for me. Obviously would need to be adapted to fit with the rest of the API, but it's nice to see it running.

Can you confirm if it works without creating a `Controller`?

I tried the changes (mainly `Device.with_hci`) but if I do not instantiate the `Controller` then my virtual HCI Host (Zephyr Posix, Zephyr NRF52840 in Renode) hang as there is no HCI Controller stack.

vChavezB:
Just a short update. You can now set the HCI Mode (Host or controller).

Set as HCI Host either env. variable `BLEAK_BUMBLE_HOST` or argument.

jpwright:
Hey @vChavezB - thanks for your work on this! bumble is a tremendously valuable tool for Bluetooth development of all sorts, and it would be phenomenal to have this merged into bleak.

jpwright:
Overall, it worked fine for me. The only feedback I would give is to consider the use case where you have both a BleakScanner and BleakClient alive at the same time. Right now you have both `BleakScanner.start()` and `BleakClient.connect()` open the bumble transport, but it can't be shared. I made a slight tweak just to externally manage the transport, and would suggest passing that as an optional argument.

vChavezB:
I have decided to make the bumble backend as an out of tree project. @jpwright  if you want to contribute your changes feel free to do so at  https://github.com/vChavezB/bleak-bumble/

PR #1699, opened 2024-12-16
Fix DBus connection leak

### LINKED ISSUE #1698: DBUS connection not closed properly in all cases
* bleak version: 0.22.2
* Python version: 3.11
* Operating System: Linux (BalenaOS 5.3.7)
* BlueZ version (`bluetoothctl -v`) in case of Linux: 5.66
* Docker base image: balenalib/aarch64-python:3.11

### Description

In our company we have a product which scans and reads bluetooth devices continuously. These devices run 24/7 and reboot every night. In some cases we would notice a peak in errors from a device; which would spam EOFError's each few seconds. Further inspection revealed a notice in the journalctl `dbus-daemon[1688]: [system] The maximum number of active connections for UID 0 has been reached (max_connections_per_user=256)`. From what I could see, some dbus connections could not be closed. Some more inspection revealed that almost all of the connections originated in our python script.

### What I Did

I tried to find the root cause by correlating with errors. What I found was that the about 240 errors originated from `BlueZManager::_check_device()`. Which gets called when adding a device watcher (`BlueZManager::add_device_watcher()`). As it turns out this method is called in [`BleakClientBlueZDBus::connect()` line 184](https://github.com/hbldh/bleak/blob/e01e2640994b99066552b6161f84799712c396fa/bleak/backends/bluezdbus/client.py#L184). Which is outside the try..except which does handle properly cleaning up the dbus connection in case of a failure.

Below you can find a Minimal Working Example. The increase of dbus connections can be seen with the help of the following command: `dbus-send --system --print-reply --dest=org.freedesktop.DBus /org/freedesktop/DBus org.freedesktop.DBus.ListNames | wc -l`

```python
import os
import asyncio
from asyncio import sleep

os.environ['BLEAK_LOGGING'] = ''

from bleak import AdvertisementData, BleakClient, BleakScanner, BLEDevice


async def run() -> None:
    print("Start scanning")
    devices: dict[str, tuple[BLEDevice, AdvertisementData]] = await BleakScanner().discover(timeout=20, return_adv=True)

    print(f"Found {len(devices)} devices in last scan")

    for i in range(0, 120, 10):
        print(f'sleeping {i+10}/120 seconds')
        # sleep in order to make sure bluez 'forgets' the scanned devices
        await sleep(10)

    while True:
        # avoid spamming
        await sleep(1)
        for address in devices:
            await sleep(0.1)

            print(address)
            try:
                device = devices[address][0]
                    # read battery
                    battery = int.from_bytes(await client.read_gatt_char('00002a19-0000-1000-8000-00805f9b34fb'), byteorder='little')
                    print(f"{client.address}: Battery level {battery}%")
            except Exception as e:
                print(e)
                continue


asyncio.run(run())
```

### Logs
When running my test example script this, was the output of the commands. The first command ran just before starting the script, the second command ran after running the script for a while.
```
user:~$ dbus-send --system --print-reply --dest=org.freedesktop.DBus /org/freedesktop/DBus org.freedesktop.DBus.ListNames | wc -l
120
user:~$ dbus-send --system --print-reply --dest=org.freedesktop.DBus /org/freedesktop/DBus org.freedesktop.DBus.ListNames | wc -l
353
```

koenvsteenb:
Moved the try (in bluezdbus/client.py) in order to include the call to the add_device_watcher function, since it is able to throw an error when the device does not exists in bluez anymore. This causes the just opened dbus connection to not be properly closed.

fixes #1698 

   guidelines in that document.
    - OS X / macOS >= 10.11

### DIFF
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -7,6 +7,7 @@ import logging
 import os
 import sys
 import warnings
+from contextlib import AsyncExitStack
 from typing import Callable, Dict, Optional, Set, Union, cast
 from uuid import UUID
 
@@ -148,7 +149,7 @@ class BleakClientBlueZDBus(BaseBleakClient):
 
         manager = await get_global_bluez_manager()
 
-        async with async_timeout(timeout):
+        async with async_timeout(timeout), AsyncExitStack() as stack:
             while True:
                 # Each BLE connection session needs a new D-Bus connection to avoid a
                 # BlueZ quirk where notifications are automatically enabled on reconnect.
@@ -157,6 +158,8 @@ class BleakClientBlueZDBus(BaseBleakClient):
                     negotiate_unix_fd=True,
                     auth=get_dbus_authenticator(),
                 ).connect()
+
+                stack.callback(self._cleanup_all)
 
                 def on_connected_changed(connected: bool) -> None:
                     if not connected:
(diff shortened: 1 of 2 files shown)


dlech:
Thanks for the fix. I wonder if we could improve this even more with a `contextlib.AsyncExitStack` .

koenvsteenb:
You mean something like this? My python skills are not that great, so it could be nothing. And I'm not sure if I understand the `contextlib.AsyncExitStack` correctly.
```python
    @asynccontextmanager
    async def _connect_bus(self):
        self._bus = await MessageBus(
            bus_type=BusType.SYSTEM,
            negotiate_unix_fd=True,
            auth=get_dbus_authenticator(),
        ).connect()

        try:
            yield
        finally:
            self._cleanup_all()

    def connect():
        # ...rest of the code...
        async with async_timeout(timeout), AsyncExitStack() as stack:
            while True:
                # Each BLE connection session needs a new D-Bus connection to avoid a
                # BlueZ quirk where notifications are automatically enabled on reconnect.
                await stack.enter_async_context(self._connect_bus())
```
Also I'm unsure if this means that in some places the `_cleanup_all()` call can be removed.

dlech:
Sort of. But `AsyncExitStack` also has a `push_async_callback()` method, so we don't need to make a context manager for the cleanup fucntion. And at the very end of the function, we need to call `stack.pop_all()` so that the cleanup doesn't run on successful return.

koenvsteenb:
Ah so it should be more something like this (I left out most of the irrelevant code):
```python
    def connect():
        async with async_timeout(timeout), AsyncExitStack() as stack:
            while True:
                self._bus = await MessageBus(
                    bus_type=BusType.SYSTEM,
                    negotiate_unix_fd=True,
                    auth=get_dbus_authenticator(),
                ).connect()

                stack.callback(self._cleanup_all)
                try:
                    try:
                        await self.get_services(
                            dangerous_use_bleak_cache=dangerous_use_bleak_cache
                        )

                        stack.pop_all() # added line
                        return True
                    except: BaseException:
                except BaseException:
                    local_disconnect_monitor_event.set()
                    # self._cleanup_all() line can be removed
                    raise
```

dlech:
We should be able to get rid of both try/except statements. And at the very end, we need to call `stack.pop_all()`.

koenvsteenb:
I committed some changes in which I used the `AsyncExitStack` and removed the try..except statements

koenvsteenb:
Hi, did you find some time to review my latest changes?

dlech:
Sorry it took me so long to get back to this one. Nice work, thanks!

bleak/backends/bluezdbus/client.py
```diff
@@ -148,7 +149,7 @@ async def connect(self, dangerous_use_bleak_cache: bool = False, **kwargs) -> bo
 
         manager = await get_global_bluez_manager()
 
-        async with async_timeout(timeout):
+        async with async_timeout(timeout), AsyncExitStack() as stack:
```
dlech, line 152: Just noticed an issue too late. We need a new stack for each while loop, otherwise we would end up leaking on retries. I'll fix it.

PR #1705, opened 2024-12-28
backends: winrt: scanner: enable extended advertising

JordanYates:
Enable extended advertising scanning by default by setting the `AllowExtendedAdvertisements` property, which defaults to false.

https://learn.microsoft.com/en-us/uwp/api/windows.devices.bluetooth.advertisement.bluetoothleadvertisementwatcher.allowextendedadvertisements


dlech:
What are the consequences of having this enabled all of the time? Could it break existing users?

How do macOS and BlueZ handle extended advertisements?

PR #1709, opened 2025-01-14
Log to stderr instead of stdout by default

frereit:
This PR changes the default logging stream to `sys.stderr` when `BLEAK_LOGGING` is set.

There's nothing really _wrong_ with logging to `sys.stdout`, it is a bit unusual in my opinion. It makes redirecting logging output to a file more difficult when there's already other output on stdout for example. Additionally, stdout is usually buffered, while stderr is usually unbuffered, which makes it a bit more convenient to work with for diagnostic information (e.g. grep & tail might work better).

I have opened this PR because I figured opening an issue first to ask for permission to open the PR would just take up more time. You may close this PR if this is not something you want to change :smile:


frereit:
> Sounds reasonable. Can you please also add a changelog entry?

Sorry, done.

frereit:
Hmm, I am unsure why the flake8 check fails: I did not touch any of that code, and it Works On My Machine :tm:

```
$ git rev-parse HEAD
a8996c5830815f12b220a5b09058a17fcc648ec2
$ uv tool run flake8 . --count --show-source --statistics
0
```

dlech:
> flake8 check fails

No worries. If you want to, you can add an extra commit to add E531 [here](https://github.com/hbldh/bleak/blob/e01e2640994b99066552b6161f84799712c396fa/setup.cfg#L3). Or I will take care of it later.

PR #1713, opened 2025-01-17
fix(client): Ensure that the bus is connected before pairing

pmessan:
If pairing is attempted before the bus is connected, the pairing process fails due to an exception being raised.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,10 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 `Unreleased`_
 =============
 
+Fixed
+-------
+* Fixed uninitialized bus connection when pairing before connection on BlueZ backend.
+
 `0.22.3`_ (2024-10-05)
 ======================
 
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -24,6 +24,7 @@ Contributors
 * David Johansen <davejohansen@gmail.com>
 * JP Hutchins <jphutchins@gmail.com>
 * Bram Duvigneau <bram@bramd.nl>
+* Peter-Newman Messan <p.newmwn.messan@gmail.com>
 
 Sponsors
 --------
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -436,6 +436,11 @@ class BleakClientBlueZDBus(BaseBleakClient):
             Boolean regarding success of pairing.
 
         """
+        if self._bus is None:
+            self._bus = await MessageBus(
+                bus_type=BusType.SYSTEM, negotiate_unix_fd=True
+            ).connect()
+
         # See if it is already paired.
         reply = await self._bus.call(
             Message(


CHANGELOG.rst
```diff
@@ -18,6 +18,7 @@ Changed
 * Don't change ctypes' global state ``bleak.backends.winrt.util``.
 * Improved performance of BlueZ backend when there are many adapters.
 * Added support for Python 3.13.
+* Fixed uninitialized bus connection when pairing before connection on BlueZ backend.
```
arthur-proglove, line 21: I guess that line goes under "Unreleased"

dlech:
Can you please describe how to reproduce the problem that this is fixing?

pmessan:
1. Enable discovery (have a device ready to pair / connect)
2. When device is discovered, in line with the documentation of BleakClientBlueZDBus.pair(), trust device, then pair, then connect. 
3. Calling `BleakClientBlueZDBus.pair()` on a Linux-based central will cause an error similar to `Object None has no attribute 'call'`. 
4. This is because, in `BleakClientBlueZDBus.pair()` we use the bus connection without checking if it exists whereas for example in `BleakClientBlueZDBus.connect()`, ([here](https://github.com/pmessan/bleak/blob/8aea8fb2ae7bd5f6e8b03badfc4dcb5394402efd/bleak/backends/bluezdbus/client.py#L155)), we initialize the bus connection before proceeding with the BLE connection. This means if you call `pair()` before you call `connect()`, an exception will be raised.

dlech:
OK, so I think what we really want here is to solve https://github.com/hbldh/bleak/issues/309 where we can just call `pair()` instead of `connect()`.

Or perhaps even better, add a `pair=True` paramter to the `connect()` method and also the `BleakClient()` constructor.

pmessan:
> OK, so I think what we really want here is to solve #309 where we can just call `pair()` instead of `connect()`.
> 
> Or perhaps even better, add a `pair=True` paramter to the `connect()` method and also the `BleakClient()` constructor.

Has anyone attempted to solve this already?
Also, since the pair method is currently broken and does not work without a workaround, would it make sense to fix that then tackle the enhancement as described? Currently with this workaround on my local system I can call `pair()`(to enable secure connections) then `connect()` with success.

dlech:
> Also, since the pair method is currently broken and does not work without a workaround, would it make sense to fix that then tackle the enhancement as described

I don't think it is broken. Everyone so far connects first, then pairs.

pmessan:
> > Also, since the pair method is currently broken and does not work without a workaround, would it make sense to fix that then tackle the enhancement as described
> 
> I don't think it is broken. Everyone so far connects first, then pairs.

But that's in direct contrast to the documentation that method 😄

dlech:
> But that's in direct contrast to the documentation of that method 😄

Can you share a link? I'm not finding that.

pmessan:
Additionally, per my testing, calling pair() which calls the Pair method on the BlueZ device, triggers a connection attempt by itself anyways.

pmessan:
> > But that's in direct contrast to the documentation of that method 😄
> 
> 
> 
> Can you share a link? I'm not finding that.

https://github.com/hbldh/bleak/blob/e01e2640994b99066552b6161f84799712c396fa/bleak/backends/bluezdbus/client.py#L433

dlech:
Closing in favor of #1741.

PR #1718, opened 2025-02-04
Update poetry to v2

cdce8p:
Poetry `v2` was released last month. It added support for the `[project]` table in `pyproject.toml`.
https://python-poetry.org/blog/announcing-poetry-2.0.0/

One of the main advantages of `v2` is that it's now possible to specify `project.requires-python` independently of `tool.poetry.dependencies.python` (which is used for locking). The latter only needs to be a subset of the former. This removes the need push a new release just to support the next Python version if `bleak` itself would be otherwise compatible.
https://python-poetry.org/docs/pyproject/#requires-python

Project metadata diff:
```diff
 ...
-Requires-Python: >=3.8,<3.14
+Requires-Python: >=3.8
 ...
```


pyproject.toml
```diff
@@ -16,13 +25,8 @@ classifiers = [
...
 ]
 
-[tool.poetry.urls]
-"Changelog" =  "https://github.com/hbldh/bleak/blob/develop/CHANGELOG.rst"
-"Support" =  "https://github.com/hbldh/bleak/discussions"
-"Issues" =  "https://github.com/hbldh/bleak/issues"
-
 [tool.poetry.dependencies]
-python = ">=3.8,<3.14"
+python = ">=3.8.1,<3.14"
```
dlech, line 29:
```suggestion
```

CHANGELOG.rst
```diff
@@ -11,9 +11,10 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 =============
 
 Changed
--------
+=======
```
cdce8p, line 14: Was this an intended changed during the merge?

dlech: oops, nope

dlech:
Thanks for getting things up-to-date!

cdce8p:
@dlech Sorry to bother you with this. Would you be able to create a new release with this change? Python 3.14 is just around the corner and removing the upper cap on the Python version here will make it much easier for the ecosystem to test and get things ready in time. A lot of downstream projects are probably still using poetry v1 / haven't updated their config just yet. So a pin in an upstream project like bleak propagates far. _With a new release all these projects could start removing there upper cap as well without having to update their config just yet._

This is completely independent from bleak itself supporting 3.14 or not. If it's not ready yet, that's completely fine. With a patch release now unpinning the version cap, all that would be needed would be a final release with support. By then downstream projects would already had time to remove the Python cap as well and with the typical constraints the new bleak version would get pulled in automatically during install.

PR #1731, opened 2025-03-30
Fix BleakClient response type

### LINKED ISSUE #1730: Implicit optional in BleakClient
Hi!

I'm really a fan of the library and using it a lot, so I ran into this issue:
Would it be possible (I'm volunteering to do a PR) to change the function signature of 

https://github.com/hbldh/bleak/blob/e01e2640994b99066552b6161f84799712c396fa/bleak/__init__.py#L723-L728

to

```python
    async def write_gatt_char(
        self,
        char_specifier: Union[BleakGATTCharacteristic, int, str, uuid.UUID],
        data: Buffer,
        response: bool | None = None,
    ) -> None:
```

to avoid mypy error
```
PEP 484 prohibits implicit Optional. Accordingly, mypy has changed its default to no_implicit_optional=True
Use https://github.com/hauntsaninja/no_implicit_optional to automatically upgrade your codebase
```
I have a number of classes derived from `BleakClient` and I get those complaints all over the place. I'm not sure there is any reason it was done like this or just missed out. Thanks!

patman15:
fixes #1730

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,7 +12,8 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 
 Changed
 -------
-* made response argument of class BleakClient explicitly optional
+* Made `response` argument of class `BleakClient.write_gatt_char()` explicitly optional. #1730
+* Deprecated omitting the `response` argument.
 
 `0.22.3`_ (2024-10-05)
 ======================
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,10 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 `Unreleased`_
 =============
 
+Changed
+-------
+* made response argument of class BleakClient explicitly optional
+
 `0.22.3`_ (2024-10-05)
 ======================
 
--- a/bleak/__init__.py
+++ b/bleak/__init__.py
@@ -738,10 +738,6 @@ class BleakClient:
         Consult the device's documentation or inspect the properties of the
         characteristic to find out which kind of writes are supported.
 
-        .. tip:: Explicit is better than implicit. Best practice is to always
-            include an explicit ``response=True`` or ``response=False``
-            when calling this method.
-
         Args:
             char_specifier:
                 The characteristic to write to, specified by either integer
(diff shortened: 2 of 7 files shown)


patman15:
I changed the formulation in the documentation from `the "best" operation will be used` because that originally made me omit the argument in my code, because I wanted the "best" operation to be used. :wink:

Linting errors are due to a [bug in flake8](https://stackoverflow.com/questions/77401175/how-to-make-flake8-ignore-syntax-within-strings) and is the same as in PR #1727.

bleak/__init__.py
```diff
@@ -758,8 +758,8 @@ async def write_gatt_char(
             response:
                 If ``True``, a write-with-response operation will be used. If
                 ``False``, a write-without-response operation will be used.
-                If omitted or ``None``, the "best" operation will be used
-                based on the reported properties of the characteristic.
+                For legacy reasons, if omitted or ``None`` the "write with response" operation
```
dlech, line 761: This is not entirely correct as some peripherals may incorrectly indicate that they don't support write with response even if they really do. This is why it was worded the way it was (and why we want to discourage not providing the arg - because it can be inconsistent between peripherals or change if a peripheral firmware "fixes" a missing property).

bleak/backends/bluezdbus/client.py
```diff
@@ -819,7 +819,7 @@ async def write_gatt_char(
         self,
         characteristic: BleakGATTCharacteristic,
         data: Buffer,
-        response: bool,
+        response: Optional[bool],
```
dlech, line 822: `response` is always a bool when `await self._backend.write_gatt_char(characteristic, data, response)` is called, so I don't think we should be chaning the type hints on any of the backend classes

patman15: Sorry, you are right that was overshooting by me.

bleak/__init__.py
patman15: Sorry, I do not get your point, how the original is more accurate than my fix. A few lines later it says `response = "write" in characteristic.properties` so it goes for response, if "write" is declared. In case this is wrongly declared I assume it falls back to "write NR" in the backend, which I meant by `if allowed by the characteristic`. Should I change to `if supported by the device` or do you just want to keep your formulation?
Your code, your call, just wanted to say that it lead me in the wrong direction (because the tip was not visible in the VScode hints.)

dlech: The new text doesn't mention the "write" property. The new text is only true if the property is correctly implemented by the peripheral. But there are peripherals out in the world that don't correctly implement the property. So it is possible that the peripheral actually does support write with response even if the "write" property is not present. The old text made it clear that the actual behavior it depends on if the "write" property is present or not.

dlech:
It think so. I'll probably merge it on the weekend.

CHANGELOG.rst
```diff
@@ -10,6 +10,10 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 `Unreleased`_
 =============
 
+Changed
+-------
+* made response argument of class BleakClient explicitly optional
```
dlech, line 15:
```suggestion
* Made `response` argument of class `BleakClient.write_gatt_char()` explicitly optional. #1730
* Deprecated omitting the `response` argument.
```

PR #1741, opened 2025-04-20
add pair param to BleakClient()

### LINKED ISSUE #309: Pair with devices that are not connected

### LINKED ISSUE #398: Unable to pair prior connect
* bleak version: 0.10.0 and master branch e968a7b666b4ba76477e25c24887fa57afda0e19
* Python version: 3.7.3
* Operating System: Linux-4.19.0-13-amd64-x86_64-with-debian-10.7
* BLE device address scheme: Private Random Resolvable address

### Description
Unable to pair (or trust) prior connect and can not connect before pair. 

### What I Did

As the D-bus resources `ClientBlueZDBus:{_reactor, _bus}` is only initialized in the `connect` method I added  check and initialization to both `connect` and `pair`:
```python
        if self._reactor is None:
            self._reactor = get_reactor(loop)

        if self._bus is None:
            self._bus = await txdbus_connect(self._reactor, busAddress="system").asFuture(
                loop
            )
```
However this did not resolve my issue. The only way I can successful pair is with following sequence:
 - remote call "StartDiscovery"
- set property "Trusted" = True
 - remote call "Pair"
- wait on property "Connected" to be True. This can take several seconds even though property "Paired" is True. Also, the state change only seem to occur if Discovery enabled/started.
- remote call "StopDiscovery"

Trying to call `BleakClientBlueZDBus:connect` before or after the above gives the following error:
 `txdbus.error.RemoteError: org.freedesktop.DBus.Error.UnknownObject: Method "Connect" with signature "" on interface "org.bluez.Device1" doesn't exist`. 

Questions:
- Either something behaves different with my setup or pairing have not been working for some time? 
  If paring works with another setup it would be interesting to know?

- What is the expected/intended behavior? Should it be possible to call `connect` after a successful call to `pair`?

- I am tempted to suggest that pairing should be handled when  connecting. (i.e. as in MacOS). Not sure
  it is good idea or how the implementation details would look like. Thoughts?

Somewhat related but closed issues:
- https://github.com/hbldh/bleak/issues/309
- https://github.com/hbldh/bleak/issues/363

dlech:
This is an alternative to #1713 and fixes #309.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,10 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 Fixed
 -----
 
+Added
+-----
+* Added ``pair`` parameter to ``BleakClient()`` constructor to allow pairing before connecting. Fixes #309.
+
 Changed
 -------
 * Made `response` argument of class ``BleakClient.write_gatt_char()`` explicitly optional. #1730
--- a/bleak/__init__.py
+++ b/bleak/__init__.py
@@ -477,6 +477,13 @@ class BleakClient:
         timeout:
             Timeout in seconds passed to the implicit ``discover`` call when
             ``address_or_ble_device`` is not a :class:`BLEDevice`. Defaults to 10.0.
+        pair:
+            Attempt to pair with the the device before connecting, if it is not
+            already paired. This has no effect on macOS since pairing is initiated
+            automatically when accessing a characteristic that requires authentication.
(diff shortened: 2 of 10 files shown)


dlech:
Hi @pmessan, I finally got around to implementing my suggestion. Let me know if this works for you.

PR #1742, opened 2025-04-26
Corebluetooth notify read fix

alanfischer:
I ran into an issue on CoreBluetooth where I have a peripheral that notifies on a certain characteristic multiple times a second, and requires a read of the characteristic after each notification.

Occasionally I would receive this error:

```Exception in callback PeripheralDelegate.did_update_value_for_characteristic()
handle: <Handle PeripheralDelegate.did_update_value_for_characteristic()>
  File "/opt/homebrew/Cellar/python@3.13/3.13.2/Frameworks/Python.framework/Versions/3.13/lib/python3.13/asyncio/events.py", line 89, in _run
    self._context.run(self._callback, *self._args)
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "~/bleak/backends/corebluetooth/PeripheralDelegate.py", line 389, in did_update_value_for_characteristic
    future.set_result(value)
    ~~~~~~~~~~~~~~~~~^^^^^^^
asyncio.exceptions.InvalidStateError: invalid state
```

I was able to reproduce it with this test script:

```python
import asyncio

TARGET_NAME_FRAGMENT = "DEVICE_NAME"

async def main():
    print("Scanning for devices...")
    devices = await BleakScanner.discover()
    target_device = None
    for device in devices:
        if TARGET_NAME_FRAGMENT in (device.name or ""):
            target_device = device
            break

    if not target_device:
        print(f"No device found with name containing '{TARGET_NAME_FRAGMENT}'")
        return

    print(f"Found device: {target_device.name} [{target_device.address}]")

    async with BleakClient(target_device) as client:
        print("Connected, discovering services...")
        services = await client.get_services()

        async def reread(char):
            data = await client.read_gatt_char(char)
            print(f"Read on {char}: {data.hex()}")

        async def notification_handler(char, data):
            asyncio.create_task(reread(char))
            print(f"Notification from {char}: {data.hex()}")

        # Subscribe to notifications for all characteristics that support it
        for service in services:
                if "notify" in char.properties:
                    print(f"Subscribing to {char.uuid}")
                    await client.start_notify(char.uuid, notification_handler)

        # Keep running to receive notifications
        while True:
            await asyncio.sleep(1)

if __name__ == "__main__":
    asyncio.run(main())
 ```
 
 And having my device rapidly emit notifications on a characteristic that is also readable.
 
 This change protects against the futures in `PeripheralDelegate.py` from being completed multiple times.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,7 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 =============
 Fixed
 -----
+* Fixed possible ``InvalidStateError`` when doing a read after a notify (OSX).
 
 Changed
 -------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,8 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 
 `Unreleased`_
 =============
+Fixed
+-----
 
 Changed
 -------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,7 +19,7 @@ Changed
 * Updated Poetry build system version to ``>=2.0``.
 * Log to stderr instead of stdout when ``BLEAK_LOGGING`` is enabled.
 * Updated ``winrt`` backend to use PyWinRT >= 3.1.
-* Added `notification_discriminator` parameter to start_notify on CoreBluetooth backend
+* Added ``notification_discriminator`` parameter to start_notify on CoreBluetooth backend
 
 Fixed
 -----
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,9 +9,6 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 
 `Unreleased`_
 =============
-Fixed
------
-* Fixed possible ``InvalidStateError`` when doing a read after a notify (OSX).
 
 Changed
 -------
@@ -20,6 +17,7 @@ Changed
 * Updated Poetry build system version to ``>=2.0``.
 * Log to stderr instead of stdout when ``BLEAK_LOGGING`` is enabled.
 * Updated ``winrt`` backend to use PyWinRT >= 3.1.
+* Added `notification_discriminator` parameter to start_notify on CoreBluetooth backend
 
 Fixed
 -----
--- a/bleak/backends/corebluetooth/PeripheralDelegate.py
+++ b/bleak/backends/corebluetooth/PeripheralDelegate.py
@@ -401,7 +401,9 @@ class PeripheralDelegate(NSObject):
                 return
 
         if not future:
-            logger.warning("Unexpected event didUpdateValueForCharacteristic")
+            logger.warning(
+                f"Unexpected event didUpdateValueForCharacteristic for {characteristic.handle()} with value: {bytes(value)} and error: {error}"
+            )
             return
 
         if error is not None:
--- a/bleak/backends/corebluetooth/PeripheralDelegate.py
+++ b/bleak/backends/corebluetooth/PeripheralDelegate.py
@@ -221,7 +221,7 @@ class PeripheralDelegate(NSObject):
         self,
         characteristic: CBCharacteristic,
         callback: NotifyCallback,
-        notification_discriminator: Optional[Callable[[bytearray], bool]] = None,
+        notification_discriminator: Optional[Callable[[bytes], bool]] = None,
     ) -> None:
         c_handle = characteristic.handle()
         if c_handle in self._characteristic_notify_callbacks:
@@ -398,11 +398,12 @@ class PeripheralDelegate(NSObject):
 
                 if notify_callback:
                     notify_callback(bytearray(value))
-                return
+                    return
 
         if not future:
             logger.warning(
-                f"Unexpected event didUpdateValueForCharacteristic for {characteristic.handle()} with value: {bytes(value)} and error: {error}"
+                "Unexpected event didUpdateValueForCharacteristic for 0x%04x with value: %r and error: %r",
+                c_handle, value, error
             )
             return
 
--- a/bleak/backends/corebluetooth/PeripheralDelegate.py
+++ b/bleak/backends/corebluetooth/PeripheralDelegate.py
@@ -383,17 +383,17 @@ class PeripheralDelegate(NSObject):
         future = self._characteristic_read_futures.get(c_handle)
 
         # If error is set, then we know this was a read response.
-        # Otherwise, we can't tell if this is a read response or notification.
+        # Otherwise, if there is a pending read request, we can't tell if this is a read response or notification.
         # If the user provided a notification discriminator, we can use that to
         # identify if this callback is due to a notification by analyzing the value.
-        # If not, and there is no future, we assume it is a notification but can't know for sure.
+        # If not, and there is a future (pending read request), we assume it is a read response but can't know for sure.
         if not error:
             notification_discriminator = (
                 self._characteristic_notification_discriminators.get(c_handle)
             )
-            if (
+            if not future or (
                 notification_discriminator and notification_discriminator(bytes(value))
-            ) or not future:
+            ):
                 notify_callback = self._characteristic_notify_callbacks.get(c_handle)
 
                 if notify_callback:
--- a/bleak/backends/corebluetooth/PeripheralDelegate.py
+++ b/bleak/backends/corebluetooth/PeripheralDelegate.py
@@ -12,7 +12,7 @@ import asyncio
 import itertools
 import logging
 import sys
-from typing import Any, Dict, List, Iterable, NewType, Optional
+from typing import Any, Callable, Dict, Iterable, NewType, Optional
 
 if sys.version_info < (3, 11):
     from async_timeout import timeout as async_timeout
@@ -39,6 +39,7 @@ CBPeripheralDelegate = objc.protocolNamed("CBPeripheralDelegate")
 
 CBCharacteristicWriteType = NewType("CBCharacteristicWriteType", int)
 
+
 class PeripheralDelegate(NSObject):
     """macOS conforming python class for managing the PeripheralDelegate for BLE"""
 
@@ -62,7 +63,7 @@ class PeripheralDelegate(NSObject):
         self._service_characteristic_discovered_futures: Dict[int, asyncio.Future] = {}
         self._characteristic_descriptor_discover_futures: Dict[int, asyncio.Future] = {}
 
-        self._characteristic_read_futures: Dict[int, List[asyncio.Future]] = {}
+        self._characteristic_read_futures: Dict[int, asyncio.Future] = {}
         self._characteristic_write_futures: Dict[int, asyncio.Future] = {}
 
         self._descriptor_read_futures: Dict[int, asyncio.Future] = {}
@@ -70,7 +71,9 @@ class PeripheralDelegate(NSObject):
 
         self._characteristic_notify_change_futures: Dict[int, asyncio.Future] = {}
         self._characteristic_notify_callbacks: Dict[int, NotifyCallback] = {}
-        self._characteristic_notify_check_callbacks: Dict[int, NotifyCheckCallback] = {}
+        self._characteristic_notification_discriminators: Dict[
+            int, Callable[[bytearray], bool]
+        ] = {}
 
         self._read_rssi_futures: Dict[NSUUID, asyncio.Future] = {}
 
@@ -152,19 +155,14 @@ class PeripheralDelegate(NSObject):
 
         future = self._event_loop.create_future()
 
-        if characteristic.handle() not in self._characteristic_read_futures:
-            self._characteristic_read_futures[characteristic.handle()] = [future]
-        else:
-            self._characteristic_read_futures[characteristic.handle()].append(future)
+        self._characteristic_read_futures[characteristic.handle()] = future
 
         try:
             self.peripheral.readValueForCharacteristic_(characteristic)
             async with async_timeout(timeout):
                 return await future
         finally:
-            self._characteristic_read_futures[characteristic.handle()].remove(future)
-            if not self._characteristic_read_futures[characteristic.handle()]:
-                del self._characteristic_read_futures[characteristic.handle()]
+            del self._characteristic_read_futures[characteristic.handle()]
 
     @objc.python_method
     async def read_descriptor(
@@ -220,14 +218,19 @@ class PeripheralDelegate(NSObject):
 
     @objc.python_method
     async def start_notifications(
-        self, characteristic: CBCharacteristic, callback: NotifyCallback, check_callback: NotifyCheckCallback = None
+        self,
+        characteristic: CBCharacteristic,
+        callback: NotifyCallback,
+        notification_discriminator: Callable[[bytearray], bool] = None,
     ) -> None:
         c_handle = characteristic.handle()
         if c_handle in self._characteristic_notify_callbacks:
             raise ValueError("Characteristic notifications already started")
 
         self._characteristic_notify_callbacks[c_handle] = callback
-        self._characteristic_notify_check_callbacks[c_handle] = check_callback
+        self._characteristic_notification_discriminators[c_handle] = (
+            notification_discriminator
+        )
 
         future = self._event_loop.create_future()
 
@@ -254,7 +257,7 @@ class PeripheralDelegate(NSObject):
             del self._characteristic_notify_change_futures[c_handle]
 
         self._characteristic_notify_callbacks.pop(c_handle)
-        self._characteristic_notify_check_callbacks.pop(c_handle)
+        self._characteristic_notification_discriminators.pop(c_handle)
 
     @objc.python_method
     async def read_rssi(self) -> NSNumber:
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by dlech** [comment]
Thanks for the clear explanation of the problem you are trying to solved. I always thought it was strange that Apple didn't make separate callback delegates for notifications and reads because it causes issues like this.

First, since `did_update_value_for_characteristic` is the only callback that has multiple triggers, I thing we should only be applying a fix to that one. If we hit this condition on any other callback, it would be because the user tried to do two things in parallel, which isn't supported. So let's drop all of the other changes to the other callbacks.

Then, for the actual problem at hand, I assume that this is a case where a notification isn't big enough to contain the full characteristic value and so you have to read after a notification to get the full value?

As for the fix, I'm afraid the `future.done()` check is still possibly racy. Consider:

* Receive a BLE notification (schedules `did_update_value_for_characteristic`).
* This triggers the notification handler callback because there is no read future (`self._characteristic_read_futures.get(c_handle)` returns `None`)
* This triggers a read request.
* Now one of two things can happen:
  1. We first receive another notification.
    *  This schedules `did_update_value_for_characteristic`.
    * `self._characteristic_read_futures.get(c_handle)` returns the `future` for the read request.
    * The `future` is **incorrectly** completed with the value from the notification.
    * We receive the BLE read response.
    * This schedules `did_update_value_for_characteristic`.
    * There are two possible states here:
      1. `self._characteristic_read_futures.get(c_handle)` returns `None` because the read function completed already.
        * The characteristic value from the read response is **incorrectly** passed to the notification handler.
        * This incorrectly triggers another read even though there hasn't been another notification since the last read.
      2. `self._characteristic_read_futures.get(c_handle)` returns a completed future because the read function has not resumed yet.
        * Without the fix, we get a crash. With the fix, we get the incorrect behavior from above.
  2. We receive the read response.
    *  This schedules `did_update_value_for_characteristic`.
    * `self._characteristic_read_futures.get(c_handle)` returns the `future` for the read request.
    * The `future` is correctly completed with the value from the notification.
    * We receive a BLE notification.
    * This schedules `did_update_value_for_characteristic`.
    * There are two possible states here:
      1. `self._characteristic_read_futures.get(c_handle)` returns `None` because the read function completed already.
        * The characteristic value from the notification is correctly passed to the notification handler.
        * This correctly triggers another read.
        * All is well and good in the world (this is the only bug-free path).
      2. `self._characteristic_read_futures.get(c_handle)` returns a completed future because the read function has not resumed yet.
        * Without the fix, we get a crash.
        * With the fix, we get the behavior above, but....
        * Since a 2nd read is triggered and the first has not resumed yet, we could get a situation where the second read function overwrites `self._characteristic_read_futures[c_handle]` then the first read function deletes  `self._characteristic_read_futures[c_handle]` from the 2nd read call. If this happened, we would get a timeout of the 2nd read function because the future was never completed, then it would crash when it tries to delete  `self._characteristic_read_futures[c_handle]` that was already deleted.

If you can log Bluetooth packets for a long time and show that we never get a notification between a read request and read response, then I will feel better that this fix might possibly be sufficient because that means we would never have the first branch of possibilities above. But if we see that we are hitting that branch, then I'm not really sure what we could do.

**Comment by alanfischer** [comment]
Thanks for the detailed feedback @dlech!

I did more investigation this morning, and unfortunately the first case looks to be whats happening.  Meaning there is no way I could get notifications consistently into the right handler without additional information.

You are correct in the behavior of notifications for this device.  It is to indicate a read for new data is required.

I ended up with a solution that works 100% of the time, however it required the passing in of another callback, which for this case I called notifiy_check_callback.  The purpose of this is only to verify if the data is intended as a notify, or read.  In my case all the notification responses are a single byte, whereas the read responses are longer.  For the completely reliable solution I also had to add in a queue for the read requests.

The only change to my example code above is something along these lines:
```
def notification_check_handler(data):
    return len(data) == 1

await client.start_notify(uuid.uuid, notification_handler, notification_check_handler)
```

I put up a quick and dirty change that adds this behavior to the corebluetooth backend, to help show what was necessary.

Is a cleaned up version of this something you would be interested in?  I understand the exceptionally unfortunate addition of an optional parameter only needed by one backend, but this does give us completely reliable data in our situation, and I'm guessing this isn't the only time this situation has arisen with corebluetooth.

**Comment by dlech** [comment]
> without additional information.

That is the same conclusion I came to.

> Is a cleaned up version of this something you would be interested in?

Yes, I'm sure this isn't the only device that could benefit. :smiley: 

Some quick feedback...

We've been trying to standardize backend-specific parameters as a separate dict with the backend name. It is a bit more verbose, but I would prefer this:

```python
await client.start_notify(uuid.uuid, notification_handler, cb=dict(notification_discriminator=notification_check_handler))
```

(`cb` is short for CoreBluetooth and I would like to have `discriminator` in the name since that is what the function is actually doing).

And we'll want an extra section in the docs at https://bleak.readthedocs.io/en/latest/backends/macos.html#specific-features-for-the-macos-backend with a longer explanation of why this exists and how to use it. I think your example is great for that.

Finally, I'm not super-keen on changing `self._characteristic_read_futures` to allow having more than one read active at a time for a characteristic. Ideally, the user of the library would be doing some sort of synchronization to make sure one read completes before we attempt another. Bluetooth chips tend to get confused if we are making multiple requests at the same time.

alanfischer:
Great suggestions!  I've added the notification_discriminator parameter, along with the removal of the read queue.  I've added the reread queue in my client code and that solved that part of the issue.  Also updated the docs.

One item I wasn't sure about:  The current notification_discriminator callback is not async, and also doesn't pass in the characteristic like the other notify callbacks.  Considering the async wrapping & characteristic passing was happening in __init__.py and not in the client backend, I wasn't sure if I should add some logic around that in the corebluetooth backend specifically, or just leave it as is (non async, no characteristic parameter)?

Thanks

alanfischer:
Hi @dlech  just checking to see if you had any additional thoughts on the above.  Thanks!

dlech:
> The current notification_discriminator callback is not async,

As it should be. It should just be a very simple function.



> and also doesn't pass in the characteristic like the other notify callbacks

That's fine. People can use `functools.partial()` if they need to bind a characteristic UUID or handle or other info to a common function used by multiple characteristics.

bleak/backends/corebluetooth/PeripheralDelegate.py
```diff
@@ -369,6 +380,19 @@ def did_update_value_for_characteristic(
     ) -> None:
         c_handle = characteristic.handle()
 
+        if not error:
```
dlech, line 383: We could probably also bring the `if not future` check up here too to avoid repeating some code.

bleak/backends/corebluetooth/PeripheralDelegate.py
```diff
@@ -369,6 +380,19 @@ def did_update_value_for_characteristic(
...
+            notification_discriminator = (
+                self._characteristic_notification_discriminators.get(c_handle)
+            )
+            if notification_discriminator and notification_discriminator(
+                bytearray(value)
```
dlech, line 388: I would use `bytes` here. The use of `bytearray` elsewhere is something of an odd choice, but can't change it now without breaking users.

bleak/backends/corebluetooth/client.py
```diff
@@ -361,7 +361,10 @@ async def start_notify(
         """
         assert self._delegate is not None
 
-        await self._delegate.start_notifications(characteristic.obj, callback)
+        cb = kwargs.get("cb") or dict()
```
dlech, line 364: Would be better to make this an explicit keyword-only argument, then we can have type hints for it and we can document it in the "frontend" `start_notify()` function in `__init__.py`.

docs/backends/macos.rst
```diff
@@ -28,7 +28,31 @@ As stated above, this will however only work the macOS machine that performed
 the scan and thus cached the device as ``243E23AE-4A99-406C-B317-18F1BD7B4CBE``.
 
 There is also no pairing functionality implemented in macOS right now, since it does not seem
-to be any explicit pairing methods in the COre Bluetooth.
+to be any explicit pairing methods in the Core Bluetooth.
+
```
dlech, line 32: Would be nice to add a section heading here. (probably have add another one above for the existing section too for it to make sense)

docs/backends/macos.rst
```diff
@@ -28,7 +28,31 @@ As stated above, this will however only work the macOS machine that performed
...
+CoreBluetooth does not differentiate between data from a notification, and data from a read.
+This can cause confusion in cases where a device may send a notification message on a characteristic
+that is an indication that the characteristic needs to be read again.
```
dlech, line 35: "indication" probably isn't the best choice of words here since it is the technical name of a BLE notification that requires a response. :wink:

docs/backends/macos.rst
```diff
@@ -28,7 +28,31 @@ As stated above, this will however only work the macOS machine that performed
...
+Bleak has can accept a ``notification_discriminator`` callback in the ``cb`` dict parameter that is
+passed to the start_notify function that can differentiate between these types of data.
+
+.. code-block:: python
+
+    async def reread(char):
+        data = await client.read_gatt_char(char)
+
+    async def notification_handler(char, data):
+        asyncio.create_task(reread(char))
```
dlech, line 46: This is buggy code. The returned task could be garbage-collected before completing. Also, it lacks any synchronization to prevent two reads at the same time. So best to not put that in an example that people might copy.

Instead, I would use an `asyncio.Event` and `set()` it here. Then in the main code after the `start_notify()`, have the following:

```python
while True:
    await event.wait()
    # we received a notification - prepare to receive another
    event.clear()
    # then read the characteristic to get the full value
    data = await client.read_gatt_char(char)
    # do stuff with data
```

docs/backends/macos.rst
```diff
@@ -28,7 +28,31 @@ As stated above, this will however only work the macOS machine that performed
...
+
+    def notification_check_handler(data):
+        return len(data) == 1
```
dlech, line 49:
```suggestion
        # We can identify notifications on this characteristic because they
        # only contain 1 byte of data. Read responses will have more than
        # 1 byte.
        return len(data) == 1
```

dlech (changes requested):

bleak/backends/corebluetooth/PeripheralDelegate.py
alanfischer: Rearranged!  I also added in a check for the case of 'no future' below this, since pulling that up still left it open for the case of "no future, but an error = call future and fail"

bleak/backends/corebluetooth/client.py
alanfischer: I made it an explicit argument in `corebluetooth/client.py`.  I didn't yet expose it explicitly in` __init__.py`, but let me know if you want me to go that far.

docs/backends/macos.rst
alanfischer: Sections added!

bleak/backends/corebluetooth/PeripheralDelegate.py
alanfischer: Now using bytes.

docs/backends/macos.rst
alanfischer: Great point!  That would be a bit confusing.  Reworded.

docs/backends/macos.rst
alanfischer: Comments added, documentation examples cleaned up to not be buggy.

alanfischer:
I appreciate the detailed feedback @dlech!  I've made another pass at the changes.

bleak/backends/corebluetooth/client.py
dlech: Yes, that is what I was suggesting.

bleak/backends/corebluetooth/PeripheralDelegate.py
```diff
@@ -371,14 +382,26 @@ def did_update_value_for_characteristic(
...
+                notification_discriminator and notification_discriminator(bytes(value))
+            ) or not future:
                 notify_callback = self._characteristic_notify_callbacks.get(c_handle)
 
                 if notify_callback:
                     notify_callback(bytearray(value))
+                return
+
+        if not future:
+            logger.warning("Unexpected event didUpdateValueForCharacteristic")
```
dlech, line 404: Could be helpful to include value and error here.

bleak/backends/corebluetooth/client.py
```diff
@@ -354,14 +354,19 @@ async def start_notify(
         self,
         characteristic: BleakGATTCharacteristic,
         callback: NotifyCallback,
+        cb: Optional[Dict[str, Any]] = None,
```
dlech, line 357:
```suggestion
        *,
        cb: Dict[str, Any],
```

This will make it a keyword only arg. (we're doing the same elsewhere already)

dlech, line 357: Would also be nice to make a `TypedDict` for this. (I don't mind doing that later though)

docs/backends/macos.rst
```diff
@@ -27,8 +27,45 @@ In the example files, this is handled in this fashion:
 As stated above, this will however only work the macOS machine that performed
 the scan and thus cached the device as ``243E23AE-4A99-406C-B317-18F1BD7B4CBE``.
 
-There is also no pairing functionality implemented in macOS right now, since it does not seem
-to be any explicit pairing methods in the COre Bluetooth.
+Pairing
+^^^^^^^
+There is no pairing functionality implemented in macOS right now, since it does not seem
+to be any explicit pairing methods in CoreBluetooth.
```
dlech, line 33: Note to self, this section could use updating.

docs/backends/macos.rst
```diff
@@ -27,8 +27,45 @@ In the example files, this is handled in this fashion:
...
+
+Notifications
+^^^^^^^^^^^^^
+CoreBluetooth does not differentiate between data from a notification and data from a read.
+This can cause confusion in cases where a device may send a notification message on a characteristic
+as a signal that the characteristic needs to be read again.
+
+Bleak has can accept a ``notification_discriminator`` callback in the ``cb`` dict parameter that is
+passed to the start_notify function that can differentiate between these types of data.
```
dlech, line 42:
```suggestion
passed to the :method:`BleakClient.start_notify` method that can differentiate between these types of data.
```

dlech (commented):
Looking good. I think we got the code all sorted out. Just a few more suggestions to polish things up a bit.

bleak/backends/corebluetooth/client.py
alanfischer: Made a keyword only arg with default of {}

bleak/backends/corebluetooth/PeripheralDelegate.py
alanfischer: Added

bleak/backends/corebluetooth/client.py
alanfischer: I started going down this path, and ran right into the `TypedDict` comment below, and was unsure what to name this. `CBStartNotifyArgs` maybe?

If you'd like this exposed at the `__init__.py` level, just let me know what you'd like the Args dict to be named and where it should live.

alanfischer:
Hi @dlech, I’ve made another pass when you have time for a review. Thanks!

CHANGELOG.rst
```diff
@@ -19,6 +19,7 @@ Changed
 * Updated Poetry build system version to ``>=2.0``.
 * Log to stderr instead of stdout when ``BLEAK_LOGGING`` is enabled.
 * Updated ``winrt`` backend to use PyWinRT >= 3.1.
+* Added `notification_discriminator` parameter to start_notify on CoreBluetooth backend
```
dlech, line 22:
```suggestion
* Added ``notification_discriminator`` parameter to start_notify on CoreBluetooth backend
```

bleak/backends/corebluetooth/PeripheralDelegate.py
```diff
@@ -214,13 +218,19 @@ async def write_descriptor(self, descriptor: CBDescriptor, value: NSData) -> Non
 
     @objc.python_method
     async def start_notifications(
-        self, characteristic: CBCharacteristic, callback: NotifyCallback
+        self,
+        characteristic: CBCharacteristic,
+        callback: NotifyCallback,
+        notification_discriminator: Optional[Callable[[bytearray], bool]] = None,
```
dlech, line 224:
```suggestion
        notification_discriminator: Optional[Callable[[bytes], bool]] = None,
```

bleak/backends/corebluetooth/PeripheralDelegate.py
```diff
@@ -371,14 +382,28 @@ def did_update_value_for_characteristic(
...
+            ):
                 notify_callback = self._characteristic_notify_callbacks.get(c_handle)
 
                 if notify_callback:
                     notify_callback(bytearray(value))
+                return
+
+        if not future:
+            logger.warning(
+                f"Unexpected event didUpdateValueForCharacteristic for {characteristic.handle()} with value: {bytes(value)} and error: {error}"
```
dlech, line 405:
```suggestion
                "Unexpected event didUpdateValueForCharacteristic for 0x%04x with value: %r and error: %r",
                c_handle, value, error
```

bleak/backends/corebluetooth/PeripheralDelegate.py
```diff
@@ -371,14 +382,28 @@ def did_update_value_for_characteristic(
...
+                self._characteristic_notification_discriminators.get(c_handle)
+            )
+            if not future or (
+                notification_discriminator and notification_discriminator(bytes(value))
+            ):
                 notify_callback = self._characteristic_notify_callbacks.get(c_handle)
 
                 if notify_callback:
                     notify_callback(bytearray(value))
+                return
```
dlech, line 401:
```suggestion
                    return
```

bleak/backends/corebluetooth/PeripheralDelegate.py
```diff
@@ -371,14 +382,29 @@ def did_update_value_for_characteristic(
...
                 notify_callback = self._characteristic_notify_callbacks.get(c_handle)
 
                 if notify_callback:
                     notify_callback(bytearray(value))
+                    return
+
+        if not future:
+            logger.warning(
+                "Unexpected event didUpdateValueForCharacteristic for 0x%04x with value: %r and error: %r",
+                c_handle, value, error
```
dlech, line 406:
```suggestion
                c_handle,
                value,
                error,
```

dlech:
OK, calling this good enough for now. Thanks for all of the work and testing on this.

PR #1744, opened 2025-05-05
Allow getting notifications from BLE devices connected to HCI adapters with index > 9

mmatthe:
This seems like a bit constructed problem, but it indeed occured to me in real world (see below for background info).

When the BT adapter that a device is connected to has an index >9 (i.e. `hci10`, `hci11` etc.), no notifications can be received. This is due to the hardcoded character count in `device_path_from_characteristic_path`. This PR makes the extraction of the device path more robust by splitting the string by `/` and rebuilding it accordingly.

Background: I was fighting with a BLE adapter which constantly would lose connection to the devices and needed unplug+replug to get back the connection. Every unplugging would increase the hci index if replugged too fast. When I finally got a connection to my device, I could not get the notifications in bleak, despite gatttool could receive them. Also, in the debug output of bleak the notifications came in, so the issue was then relatively quickly found.


bleak/backends/bluezdbus/utils.py
```diff
@@ -52,7 +52,7 @@ def device_path_from_characteristic_path(characteristic_path: str) -> str:
         A D-Bus object path of the device.
     """
     # /org/bluez/hci1/dev_FA_23_9D_AA_45_46/service000c/char000d
-    return characteristic_path[:37]
+    return "/".join(characteristic_path.split("/")[:5])
```
dlech, line 55: This is in a hot path that gets called a lot, so would want to see that we don't have a performance regression.

Better would probably be:

```suggestion
    return characteristic_path[:-21]
```

CHANGELOG.rst
```diff
@@ -25,6 +25,7 @@ Fixed
 * Fixed possible ``KeyError`` when getting services in BlueZ backend. Fixes #1435.
 * Fix D-Bus connection leak when connecting to a device fails in BlueZ backend. Fixes #1698.
 * Fixed possible deadlock when connecting on WinRT backend when device is already connected.
+* Allow getting notifications from devices connected to BLE adapters with index >9 (hci10, hci11, ...)
```
dlech, line 28:
```suggestion
* Fixed getting notifications from devices connected to BLE adapters with index >9 (hci10, hci11, ...)
```

bleak/backends/bluezdbus/utils.py
mmatthe: Thanks for the quick feedback! Performance is a good point, which I didn't think of. Indeed, the `join` version is 5 times slower on my PC:

```
In [1]: N="/org/bluez/hci1/dev_FA_23_9D_AA_45_46/service000c/char000d"

In [2]: %timeit N[:37]
119 ns ± 0.996 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [3]: %timeit "/".join(N.split("/")[:5])
636 ns ± 1.74 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [4]: %timeit N[:-21]
120 ns ± 1.02 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

For my use case, using `[:-21]` would also work. Is it certain that the service and char always have 4 number digits?

mmatthe:
I included the change using `[-21]` and hence this is ready for review.

dlech:
Thanks!

PR #1746, opened 2025-05-12
Att error 15 name fix

Niflheim-uk:
Small change to correct the description of error 15. Found while testing a BLE device in development.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,7 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 =============
 Fixed
 -----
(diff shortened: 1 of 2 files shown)


CHANGELOG.rst
```diff
@@ -11,6 +11,7 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 =============
 Fixed
 -----
+Corrected ATT error code 15 description to Insufficient Encryption
```
dlech, line 14: It looks like somehow we ended up with duplicate Fixes headings. There is another one below that already has entries, so we should move this with the rest and delete this header.

dlech:
Thanks!

Issue #1759, opened 2025-05-17
Windows does not close GATT session when another app is using the same peripheral


dlech:
> Currently, the disconnect call always hits the 2mins timeout, which is expected because my bleak script didn't own that BLE connection. If possible, it would be helpful if bleak could detect when it's reusing a BLE connection opened by another app. In this scenario, it might not be necessary to wait 2mins to for a timeout. maybe an immediate exception like "Disconnect failed: connection owned by another app" would be more helpful
> 
> ```
> 239468 client.py(711):             for service in services:
> 239469 client.py(768):             return new_services
> 239470 client.py(476):                     await event.wait()
> 239474 client.py(424):                 async with async_timeout(timeout):
> 239499 client.py(478):                 is_connect_complete = True
> 239500 client.py(480):                 self._services_changed_events.remove(services_changed_event)
> 239501 client.py(486):             self._session_active_events.remove(event)
> 239563 client.py(1061):             for c in self._callbacks:
> 239773 client.py(544):             if self._session is None
> 239774 client.py(545):             else self._session.session_status == GattSessionStatus.ACTIVE
> 239782 client.py(542):         return (
> 239788 client.py(490):         logger.debug("Disconnecting from BLE device...")
> 240004 client.py(492):         assert self.services
> 240005 client.py(495):         for handle, event_handler_token in list(self._notification_callbacks.items()):
> 240006 client.py(500):         self._notification_callbacks.clear()
> 240007 client.py(503):         if self.services:
> 240008 client.py(507):             await asyncio.sleep(0.1)
> 240218 client.py(509):             for service in self.services:
> 240223 client.py(510):                 service.obj.close()
> 240224 client.py(509):             for service in self.services:
> 240225 client.py(510):                 service.obj.close()
> 240226 client.py(509):             for service in self.services:
> 240227 client.py(510):                 service.obj.close()
> 240228 client.py(509):             for service in self.services:
> 240229 client.py(510):                 service.obj.close()
> 240230 client.py(509):             for service in self.services:
> 240231 client.py(510):                 service.obj.close()
> 240232 client.py(509):             for service in self.services:
> 240233 client.py(510):                 service.obj.close()
> 240234 client.py(509):             for service in self.services:
> 240235 client.py(510):                 service.obj.close()
> 240236 client.py(509):             for service in self.services:
> 240237 client.py(511):             self.services = None
> 240238 client.py(514):         if self._session:
> 240239 client.py(515):             self._session.maintain_connection = False
[...]

PR #1768, opened 2025-05-21
backends: winrt: don't wait for disconnect

dlech:
We have found that if another app is also connected to the same BLE device, we never get the GATT session close event. This causes the `disconnect()` method to hang forever.

Instead, just close the GATT session and return immediately (and hope for the best).

Fixes: https://github.com/hbldh/bleak/issues/1759

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -514,7 +514,7 @@ Changed
 * Replaced usage of deprecated ``@abc.abstractproperty``.
 * Use ``asyncio.get_running_loop()`` instead of ``asyncio.get_event_loop()``.
 * Changed "service is already present" exception to logged error in BlueZ backend. Merged #622.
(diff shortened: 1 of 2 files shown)


CHANGELOG.rst
```diff
@@ -514,6 +514,7 @@ Changed
 * Replaced usage of deprecated ``@abc.abstractproperty``.
 * Use ``asyncio.get_running_loop()`` instead of ``asyncio.get_event_loop()``.
 * Changed "service is already present" exception to logged error in BlueZ backend. Merged #622.
+* WinRT backend no longer wait for GATT session to close on disconnect. Fixes #1759.
```
dlech, line 517:
```suggestion
* WinRT backend no longer waits for GATT session to close on disconnect. Fixes #1759.
```

PR #1778, opened 2025-06-21
BlueZ: handle private resolvable address change

### LINKED ISSUE #1737: `BleakClient` fails to track address updates for devices using RPA on Linux with BlueZ.
#### Summary
When pairing with a device that uses Resolvable Private Addresses (RPA), `BleakClient` does not update its internal `address` to reflect the new address assigned by BlueZ. This leads to inconsistencies when interacting with the device post-pairing.

#### Steps to Reproduce
Run the code below:

```python
import asyncio
import subprocess

import bleak


class MyBleakClient(bleak.BleakClient):
    async def get_bluez_address(self) -> str:
        from dbus_fast.message import Message

        m = Message(
            path=await self._backend._get_device_path(),
            member="Get",
            signature="ss",
            body=[defs.DEVICE_INTERFACE, "Address"],
        )
        reply = await self._backend._bus.call(m)

        return reply.body[0].value


async def main():
    original_addr = "75:e5:ed:87:1c:28"
    d = await bleak.BleakScanner.find_device_by_address(original_addr)
    assert d is not None
    c = MyBleakClient(d)

    await c.connect()
    print("address before pairing", c.address)
    print("bluez address before pairing", await c.get_bluez_address())

    await c.pair()
    print("address after pairing", c.address)
    bluez_addr = await c.get_bluez_address()
    print("bluez address after pairing", bluez_addr)

    await c.disconnect()

    print(
        "scan for original",
        await bleak.BleakScanner.find_device_by_address(original_addr),
    )
    print("scan for bluez", await bleak.BleakScanner.find_device_by_address(bluez_addr))


if __name__ == "__main__":
    # Run bluetoothctl in the background as the pairing agent.
    with subprocess.Popen(
        ["bluetoothctl"],
        stdout=subprocess.PIPE,
        stderr=subprocess.PIPE,
        stdin=subprocess.PIPE,
    ):
        asyncio.run(main())

```

#### Expected Behavior
`BleakClient.address` updates to match the new address assigned by BlueZ after pairing.

#### Actual Behavior
The output shows `BleakClient.address` remains unchanged, while BlueZ updates the address:

```
address before pairing 75:E5:ED:87:1C:28
bluez address before pairing 75:E5:ED:87:1C:28
address after pairing 75:E5:ED:87:1C:28
bluez address after pairing EA:28:4F:29:B1:40
scan for original None
scan for bluez EA:28:4F:29:B1:40: xxxxx
```

#### Root Cause
`BleakClient` does not listen for D-Bus property changes (`org.freedesktop.DBus.Properties.PropertiesChanged`) to update the `address` property when BlueZ updates it.

I have experience with BLE on Android and iOS, and neither of these platforms change the address under your feet. I'm not sure if this root cause is because of a incorrect behavior of bluez or if its working as designed. Either way, I think this should be (a) fixed in bleak, or (b) mentioned in the documentation and a workaround added to the bluez backend.

#### Impact
- Device scanning fails with the old address post-pairing.
- Applications relying on `address` for identification experience inconsistencies.

#### Proposed Fix
Listen for `PropertiesChanged` signals on the `org.bluez.Device1.Address` property and update `BleakClient.address` accordingly.

#### Environment
- **OS**: Pop-os 22.04
- **BlueZ**: 5.64
- **Bleak**: 0.22.2
- **Python**: 3.10

dlech:
Devices with resolvable private addresses will change their address after pairing in the BlueZ backend. So for the address property to return the "correct" value, we need to update the device's address after pairing.

Fixes #1737.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -29,6 +29,7 @@ Changed
 
 Fixed
 -----
+* Fixed resolvable private address not updated after connecting in BlueZ backend. Fixes #1737.
 * Fixed possible ``KeyError`` when getting services in BlueZ backend. Fixes #1435.
 * Fix D-Bus connection leak when connecting to a device fails in BlueZ backend. Fixes #1698.
 * Fixed possible deadlock when connecting on WinRT backend when device is already connected. Fixes #1757.
--- a/bleak/__init__.py
+++ b/bleak/__init__.py
@@ -450,6 +450,10 @@ class BleakClient:
         **kwargs:
             Additional keyword arguments for backwards compatibility.
 
+    .. tip:: If you enable pairing with the ``pair`` argument, you will also
+        want to extend the timeout to allow enough time for the user to find
+        and enter the PIN code on the device, if required.
+
     .. warning:: Although example code frequently initializes :class:`BleakClient`
         with a Bluetooth address for simplicity, it is not recommended to do so
         for more complex use cases. There are several known issues with providing
--- a/bleak/__init__.py
+++ b/bleak/__init__.py
@@ -292,7 +292,7 @@ class BleakScanner:
 
         .. versionadded:: 0.19
         """
-        return self._backend.seen_devices
+        return {d[0].address: d for d in self._backend.seen_devices.values()}
 
     @classmethod
     async def find_device_by_address(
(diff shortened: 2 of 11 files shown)


dlech:
FYI @vhakulinen if you are able to test this.

PR #1782, opened 2025-06-29
"tomli" was included in Python 3.11 as "tomllib"

a-detiste:
Hi,

We are slowly trimming old backports out of Debian

https://wiki.debian.org/Python/Backports

> 2. Modify the `CHANGELOG.rst`, describing your changes as is specified by the

Changelog is lagging behind


pyproject.toml
```diff
@@ -47,7 +47,7 @@ optional = true
 [tool.poetry.group.docs.dependencies]
 Sphinx = { version = ">=8.2.3", python = ">=3.11" }
 sphinx-rtd-theme = { version = ">=3.0.2", python = ">=3.11" }
-tomli = ">=2.2.1"
+tomli = { version = ">=2.2.1", python = "<3.11" }
```
dlech, line 50: The Sphinx version already requires Python >= 3.11, so we can just drop this dependency altogether.

Also, as the CI caught, we need to run `poetry lock` after doing this.

docs/conf.py
```diff
@@ -17,7 +17,10 @@
 import pathlib
 import sys
 
-import tomli
+try:
+    import tomllib
+except ImportError:
+    import tomli as tomllib
```
dlech, line 23: As mentioned in the other comment. we don't need the fallback anymore, so we can just do `import tomillib` here.

dlech (changes requested):

pyproject.toml
a-detiste: indeed

pyproject.toml
```diff
@@ -26,7 +26,6 @@ classifiers = [
 ]
 
 [tool.poetry.dependencies]
-async-timeout = { version = ">=3.0.0", python = "<3.11" }
```
dlech, line 29: This is a runtime dependency, so we can't remove this.

It is only building the docs that requires Python >=3.11. Bleak itself still supports down to 3.9.

dlech (changes requested):

a-detiste:
Or juste obey `isort` wish and add this newline

PR #1784, opened 2025-06-30
backends/device: restore **kwargs in BLEDevice constructor

dlech:
The `BLEDevice()` constructor technically isn't considered part of the public API, so a breaking change was made removing `rssi` and `**kwargs`. However, there are users that implement their own backends, so effectively this should be considered public and the breaking change reverted.

It looks like all users are passing `rssi` as a keyword argument, so just restoring `**kwargs` should be sufficient.

Cc: @cdce8p @balloob @bdraco


bdraco:
Thank you

Issue #1783, opened 2025-06-30
Issues migrating from `0.22.3` to `1.0.0`


cdce8p:
First of all thanks for the new release!

I'm currently looking into updating bleak in Home Assistant and noticed quite a few breaking changes. In itself this would be fine. Unfortunately though it's quite difficult to support both versions `0.22.3` and `1.0.0` at the same time. Especially problematic here is the removal of `rssi` argument to `BLEDevice` in https://github.com/hbldh/bleak/commit/9af83f7aa149bb6926d921dbbe267404560652f3. It was a required argument in `0.22.3` and is now completely removed together with `**kwargs` essentially breaking every package. A better approach IMO would be to offer at least one version where `rssi` is changed to an `optional` argument (with default) before outright removing it.

Ref: https://github.com/home-assistant/core/pull/147742#issuecomment-3018372924

There are other breaking changes like the move to new sub packages in https://github.com/hbldh/bleak/commit/3c777c2f9fe59faea20889964db7bf3ab6d4c203 which also creates some headaches. But these are easy enough to work around with import guards.

Ref: https://github.com/Bluetooth-Devices/habluetooth/issues/248#issuecomment-3016578920

balloob:
I agree. If 1.0.1 would allow an unused `rssi` keyword, all libraries could be migrated to support both `1.0.0` and `0.22.3`. Then Bleak `1.0.2` can drop the `rssi` keyword again. 

Without `rssi` in 1.0.0, Home Assistant needs to update every Bluetooth package at the same time, to avoid breaking integrations. There are 51 one of them, and it's unclear if all have active maintainers.

Expl0dingBanana:
It looks like `rssi` has been deprecated since March 17, 2023 through https://github.com/hbldh/bleak/releases/tag/v0.20.0. I am not sure the correct fix is to reintroduce the deprecated code. It is unfortunate that it may break things, but then its up to HA if they want to restrict `bleak` to a lower version. If some of the integrations are no longer maintained folks will have the option to not upgrade, find a maintainer for the libraries, remove the devices, or fork HA and manage it themselves. I think a two year adjustment period is sufficient. Just my 2 cents.

dlech:
[...]

PR #1785, opened 2025-06-30
Restore deprecated imports

dlech:
There were some imports that were moved to `bleak.args.bluez` in v1.0 without a proper deprecation period. Restore these imports and add the proper deprecation warnings.

https://github.com/hbldh/bleak/issues/1783#issuecomment-3019828790

TODO: need to do the same for windows and corebluetooth

### DIFF
--- a/bleak/backends/bluezdbus/advertisement_monitor.py
+++ b/bleak/backends/bluezdbus/advertisement_monitor.py
@@ -15,15 +15,33 @@ if TYPE_CHECKING:
 
 import logging
 from collections.abc import Iterable
-from typing import no_type_check
+from typing import Any, no_type_check
+from warnings import warn
 
 from dbus_fast import PropertyAccess
 from dbus_fast.service import ServiceInterface, dbus_property, method
 
-from bleak.args.bluez import OrPatternLike
+from bleak.args.bluez import OrPattern as _OrPattern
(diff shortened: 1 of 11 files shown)


bleak/backends/bluezdbus/advertisement_monitor.py
```diff
@@ -20,11 +21,27 @@
...
+    "OrPatternLike": _OrPatternLike,
+}
+
+
+def __getattr__(name: str):
+    if value := _DEPRECATED.get(name):
+        warn(
+            f"importing {name} from bleak.backends.bluezdbus.advertisement_monitor is deprecated",
+            DeprecationWarning,
+        )
```
cdce8p, line 41: Please check the stacklevel for these. I believe it will need to be `2` or even `3`. Otherwise the warning won't be that helpful if you don't know where it's coming from.

cdce8p (commented):
Not sure this is really necessary. At least for `habluetooth` it's easy enough to add the import guard. Even though it's a breaking change without deprecation, this one isn't a big deal. _Once you know about it._
```py
    try:
         from bleak.args.bluez import OrPattern
     except ImportError:
         from bleak.backends.bluezdbus.advertisement_monitor import OrPattern
```

PR #1795, opened 2025-07-20
Fix service_explorer.py for non-unique characteristics UUID

patman15:
Simple fix to avoid
```
__main__ INFO: [Service] 0000fff0-0000-1000-8000-00805f9b34fb (Handle: 17): Vendor specific
__main__ INFO:   [Characteristic] 0000fff1-0000-1000-8000-00805f9b34fb (Handle: 18): Vendor specific (read,notify), Error: Multiple Characteristics with this UUID, refer to your desired characteristic by the `handle` attribute instead.
```
for devices where the `UUID` is not unique.

### DIFF
--- a/examples/service_explorer.py
+++ b/examples/service_explorer.py
@@ -65,7 +65,7 @@ async def main(args: Args):
             for char in service.characteristics:
                 if "read" in char.properties:
                     try:
-                        value = await client.read_gatt_char(char.uuid)
+                        value = await client.read_gatt_char(char.handle)
                         extra = f", Value: {value}"
                     except Exception as e:
                         extra = f", Error: {e}"


dlech:
I'm surprised I don't remember getting any complaints about this before!

Thanks for the fix!

examples/service_explorer.py
```diff
@@ -65,7 +65,7 @@ async def main(args: Args):
             for char in service.characteristics:
                 if "read" in char.properties:
                     try:
-                        value = await client.read_gatt_char(char.uuid)
+                        value = await client.read_gatt_char(char.handle)
```
dlech, line 68:
```suggestion
                        value = await client.read_gatt_char(char)
```

Drat, after merging this, I realized, we could just do this instead. It makes more sense to pass the object directly since we already have it. Otherwise, the method call just has to look up the object again by the handle.

dlech: Fixed in https://github.com/hbldh/bleak/commit/d65d1921ef478722ece2e610e4b6499885c6d5eb

patman15: :facepalm: sorry, also missed that. So easy and I just followed the error message. :laughing:

PR #1796, opened 2025-07-20
Add BleakUUIDNonUnique exception

patman15:
In addition to #1795 this fix makes it possible to detect non-unique `UUIDs` and handle them if necessary. I think this could be useful for some cases. Alternative would be to strongly recommend to use the `BleakGATTService.get_characteristic` function instead.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,10 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 `Unreleased`_
 =============
 
+Changed
+-------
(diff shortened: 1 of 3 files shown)


dlech:
Adding a specialized exception type for handling things at runtime is in general a good idea. But this doesn't seems like the kind of thing that should actually be handled at runtime. As the error message suggests, the code should be changed to avoid it in the first place.

patman15:
Ok, makes sense. Shouldn't it then better be an `assert`? If yes, I can modify, if not, I would withdraw this PR.

dlech:
Hmm... not sure if `assert` would be better or not since it can be optimized out.

PR #1798, opened 2025-07-29
Added bitchat bluetooth mesh service and characteristic UUIDs

o-murphy:
Added Bitchat Bluetooth Mesh Service and Characteristic UUIDs, [Referred to official Bitchat repo](https://github.com/permissionlesstech/bitchat).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,8 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 `Unreleased`_
 =============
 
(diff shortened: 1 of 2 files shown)


o-murphy:
> I guess this is OK, but would really rather prefer moving towards some package that automatically generates these sorts of lookup tables from https://github.com/NordicSemiconductor/bluetooth-numbers-database rather than maintaining our own tables in Bleak.

But wouldn't using such databases overburden the library (it's quite thin at the moment)? Or do you want to implement this as an optional dependency for those who really need these descriptors?

Issue #1813, opened 2025-08-20
Too much repeating conditional imports


o-murphy:
Bleak has too much repeating conditional imports that do the same things, some of them are deprecated and can be simplified with `typing_extensions`
It should be refactored to some single module like `_compat.py` and then imported
It can improve and make simlplier to create and maintain platform dependend backends (as one described in #1812)
I can refactor it and create a PR with just this changes

There are the few examples:

```python
# __init__.py
if sys.version_info < (3, 12):
    from typing_extensions import Buffer
else:
    from collections.abc import Buffer

if sys.version_info < (3, 11):
    from async_timeout import timeout as async_timeout
    from typing_extensions import Never, Self, Unpack, assert_never
else:
    from asyncio import timeout as async_timeout
    from typing import Never, Self, Unpack, assert_never
```

```python
# client.py

if sys.version_info < (3, 12):
    from typing_extensions import Buffer
else:
    from collections.abc import Buffer

```

```python
# winrt/client.py
if sys.version_info < (3, 12):
    from typing_extensions import Buffer, override
else:
    from collections.abc import Buffer
    from typing import override

if sys.version_info < (3, 11):
    from async_timeout import timeout as async_timeout
    from typing_extensions import Self, assert_never
else:
    from asyncio import timeout as async_timeout
    from typing import Self, assert_never
```

o-murphy:
Done in #1815

Issue #1812, opened 2025-08-19
Adafruit СircuitPython backend


o-murphy:
I have a sample solution to get bleak working on `CircuitPython`. For now I able to run BleakScanner on my `esp32-s3` with no issues. 
The general problem is some built-in dependencies is absent in CircuitPython (functools, inspect, uuid, asyncio.timeout). I used to replace it with native approaches for CircuitPython and it works, I used just adafruit `_bleio` and `asyncio` modules to make it.
The second problem is adafruit `_bleio` module is global and synchronous and can be used just on main thread, but there also few sollutions there.

I'm thinking about using a monkeypatching for the absent modules to get a seamless way to use bleak on CircuitPython

nomadbyte:
CircuitPython should be able to use `bleak` already. See https://github.com/adafruit/Adafruit_Blinka_bleio , that's on low-level. Adafruit expects BLE users to actually load the higher-level helper https://github.com/adafruit/Adafruit_CircuitPython_BLE, which should be part of Adafruit CircuitPython Bundle.

o-murphy:
> The goal of https://github.com/adafruit/Adafruit_Blinka_bleio is to be able to use `_bleio` (and higher level libraries) on CPython. It uses `bleak` under the covers (though we haven't updated to newer bleak in a long time). See https://learn.adafruit.com/circuitpython-ble-libraries-on-any-computer. It is not meant to be run under CircuitPython.
> 
> We did not consider `bleak` as an API to be implemented by `_bleio`, because it did not support peripheral mode, only central.

Схоже ви помиляєтесь, це трохи інша річ, `blinka bleio` це для linux і тільки, вони самі пишуть використовувати [Adafruit_CircuitPython_BLE](https://github.com/adafruit/Adafruit_CircuitPython_BLE) на мікроконтролерах, яка в свою чергу базується суто на низькорівневому `_bleio` і ніяким чином не використовує `bleak`. And also it's not provide bleak compatible interface, which I trying to do. Yeap I made a mistake saying I use `_bleio` directly, I use `Adafruit_CircuitPython_BLE`

o-murphy:
> The goal of https://github.com/adafruit/Adafruit_Blinka_bleio is to be able to use `_bleio` (and higher level libraries) on CPython. It uses `bleak` under the covers (though we haven't updated to newer bleak in a long time). See https://learn.adafruit.com/circuitpython-ble-libraries-on-any-computer. It is not meant to be run under CircuitPython.
[...]

PR #1814, opened 2025-08-20
CircuitPython backend: WIP

o-murphy:
This is the work in progress PR that tries to implement `bleak`-compatible backend for `CircuitPython`
This PR is relative to #1812 issue

### Keynotes
* For now the only `BleakScanner` works, `BleakClient` is not complete.
* It uses monkeypatching but better approach can be the conditonal imports 
* `bleak` have a lot of conditional imports that is not DRY (#1813), as you can see that creates the need to make many changes to the original code, i can create a separate PR to refactor it


o-murphy:
This can be done if #1815 will be merged

PR #1817, opened 2025-08-21
bluezdbus: improve D-Bus connection error handling in manager

skbeh:
In the `BlueZManager.async_init` method, the call to `bus.connect()` was performed before the `try...except` block that handles initialization errors.

If `bus.connect()` itself were to fail, the exception would not be caught by the block, and the `bus.disconnect()` cleanup logic would not run.

This commit moves the `await bus.connect()` call inside the `try` block to ensure that any connection-related exceptions are properly handled and the bus is disconnected.

### DIFF
--- a/bleak/backends/bluezdbus/manager.py
+++ b/bleak/backends/bluezdbus/manager.py
@@ -244,9 +244,10 @@ class BlueZManager:
             # dbus-next will destroy the underlying file descriptors
             # when the previous one is closed in its finalizer.
             bus = MessageBus(bus_type=BusType.SYSTEM, auth=get_dbus_authenticator())
-            await bus.connect()
 
             try:
+                await bus.connect()
+
                 # Add signal listeners
 
                 bus.add_message_handler(self._parse_msg)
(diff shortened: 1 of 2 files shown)


dlech:
Thanks for the fix. This API is a bit weird that `bus.disconnect()` is really `bus.disconnect()` and what should be - but does not exist - `bus.close()`. Can you also add a comment on `bus.disconnect()` in the exception handler to explain this?

bleak/backends/bluezdbus/manager.py
```diff
@@ -251,9 +251,10 @@ async def async_init(self) -> None:
             # dbus-next will destroy the underlying file descriptors
             # when the previous one is closed in its finalizer.
             bus = MessageBus(bus_type=BusType.SYSTEM, auth=get_dbus_authenticator())
-            await bus.connect()
 
             try:
+                await bus.connect()
```
dlech, line 256:
```suggestion
                # We need to call bus.disconnect() even when bus.connect() fails in
                # order to release the file handles created in the constructor.
                await bus.connect()
```

dlech:
I went ahead and added the comment.

PR #1830, opened 2025-09-19
backends/winrt/client: fix potential race condition for session status

dlech:
We need to check the current session status *after* registering the event handler, otherwise we might miss the event if the session status changes between checking the status and registering the handler.


dlech:
This didn't fix the issue we were trying to fix, but it still seems like a reasonable thing to do.

Issue #1833, opened 2025-09-21
Add support for iOS (add rubicon-objc)


timrid:
**Description**
Now that iOS is an official supported Python platform since 3.13 (see [PEP 730](https://peps.python.org/pep-0730/)), bleak should also support iOS. iOS shares almost the same CoreBluetooth API as macOS. Unfortunately `pyobjc` is not supported on iOS. Only [`rubicon-objc`](https://github.com/beeware/rubicon-objc) is available. So the main work to support iOS is to implement `rubicon-objc` instead of/in addition to `pyobjc`.

**What I have done**
I tried to replace pyobjc with rubicon-obj and tested it on a macOS. After a few hours the Scanner was basically working. So I would like to continue the work on it and get it working on iOS.

**Questions**
1. If I have a working version in the future, is there a chance that such a PR will be merged?
2. Can `pyobjc` generally be replaced by `rubicon-objc` for macOS? (I'm afraid that's not an option, but I wanted to ask the question anyway.)
3. If not, should `rubicon-objc` be implemented as a second CoreBluetooth backend? That has the disadvantage that much code is duplicated, but the advantage that the current `pyobjc` based CoreBluetooth backend is not touched. Or should the current CoreBluetooth backend be extended to support both objc-frameworks `rubicon-objc` and `pyobjc`? In this way more code is shared, but the risk of regressions is greater.
4. In either way on macOS there would be two possible backends/frameworks  available. How to distinguish between the frameworks? My first idea is to add `rubicon-objc` as `optional-dependencies` on macOS. So to use this on macOS you have to `pip install bleak[rubicon-objc]`. But this would also install `pyobjc` as this is installed by default on macOS. So there has to be an additional runtime parameter, to tell bleak which framework to use. Where is the best place for this parameter?

timrid:
> > But this would also install `pyobjc` as this is installed by default on macOS.
> 
> We should be able to add `; extra != 'rubicon-objc'` to the pyobjc dependencies to take care of that. And we will want a way to install both at the same time for development/testing.

That would be ideal, but unfortunately, extras can only be used to add additional packages, not to exclude some. The alternative would be to define both `pyobjc` and `rubicon-objc` as extras and then call `pip install bleak[pyobjc]` or `pip install bleak[rubicon-obj]`. If you only run `pip install bleak` on macOS, nothing would be installed, and if you run `pip install bleak[pyobjc,rubicon-objc]`, both would be installed. But that would of course be a breaking change.

PR #1838, opened 2025-10-09
Integration tests

timrid:
This adds integration tests for bleak. It works by emulating a BLE device with [`bumble`](https://github.com/google/bumble) in conjunction with an [nRF52840 Dongle](https://www.nordicsemi.com/Products/Development-hardware/nRF52840-Dongle).

I currently only tested this with a MacBook and the CoreBluetooth backend. But in principle this should also work similar on Linux and Windows. I dont think that it will be possible to get 100% coverage (e.g. permission handling is not automatically testable), but it will catch many cases. I think, that it should also be possible to add `tox` to automatically test all supported python versions.

I am sharing this as a draft to get high level feedback on whether integration tests are interesting in this way. Then I would continue working on this to test the other platforms as well. I developed this, because I need some way to test my work on an `rubicon-objc` backend (see https://github.com/hbldh/bleak/issues/1833).


timrid:
I have limited the tests in this PR to scanner tests for now to keep this PR small. I will submit client tests in a future PR. I hope this will speed up the merging of this PR.

I have now also run the tests on Windows and Linux. Windows did not cause any problems. For Linux, the integration tests are still disrupted by the unit tests (see https://github.com/hbldh/bleak/pull/1869).

I have completely outsourced all binaries and instructions for setting up the nRF dongle to https://github.com/timrid/ble-dongle-firmware, as this is not directly related to bleak. The binary there is compiled and released entirely in GitHub actions (including hash).

PR #1840, opened 2025-10-13
Centralize backend detection

timrid:
Currently the backend detection logic is duplicated in `bleak.backends.client` and `bleak.backends.scanner`. This PR centralizes the backend detection and added a new `BleakBackend` enum.

This additional `BleakBackend` has also the benefit for #1838 to simplify the conditional test execution depending on the detected backend.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,8 +12,8 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 
 Added
 -----
-- Added ``bleak.backends.get_default_backend()`` and ``BleakBackend`` enum for a centralized backend detection.
-- Added ``BleakClient().backend_id`` and ``BleakScanner().backend_id`` properties to identify the backend in use.
+* Added ``bleak.backends.get_default_backend()`` and ``BleakBackend`` enum for a centralized backend detection.
+* Added ``BleakClient().backend_id`` and ``BleakScanner().backend_id`` properties to identify the backend in use.
 
 Changed
 -------


dlech (commented):
As a suggestion, we could take this one step farther. Instead of creating an enum class, we could create a `BackendProvider` class that has two attributes, the `scanner_type` and `client_type`. Then we could remove the two `get_platform_{client,scanner}_backend_type()` functions completely and just use the new `get_backend()` (or `get_backend_provider()` function to replace that.

That is, replace:

```python
PlatformBleakClient = (
            get_platform_client_backend_type() if backend is None else backend
        )
```

with

```python
PlatformBleakClient = (
            get_backend_provider().client_type if backend is None else backend
        )
```

dlech:
> As a suggestion, we could take this one step farther. 

Not sure how clear I was on this and it was pretty simple anyway, so I went ahead and tried it myself in #1845.

timrid:
> > As a suggestion, we could take this one step farther.
> 
> Not sure how clear I was on this and it was pretty simple anyway, so I went ahead and tried it myself in #1845.

So, your PR #1845 is providing only a solution for the code deduplication.

But my second intention of this PR was to provide an  `BleakBackend` enum that can be used to selectively run/skip tests for some backends in #1838 . See [here](https://github.com/timrid/bleak/blob/6883f24d42dde7b35c34bc08e76fd87d6617409d/tests/integration/test_client.py#L357-L363
) for example.

dlech:
> But my second intention of this PR was to provide an `BleakBackend` enum that can be used to selectively run/skip tests f

Ah-ha. That could be useful as well at runtime to help handle platform-specific quirks.

bleak/backends/__init__.py
```diff
@@ -1,5 +1,37 @@
...
+import sys
+
+from bleak.exc import BleakError
+
+
+class BleakBackend(enum.Enum):
+    P4Android = enum.auto()
+    BlueZDBus = enum.auto()
+    PythonistaCB = enum.auto()
+    CoreBluetooth = enum.auto()
```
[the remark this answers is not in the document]
timrid: I would differentiate *bleak backends* and the *binding frameworks*. The bleak backend is `CoreBluetooth`, the binding frameworks for this backend are `pyobjc` or `rubicon-objc`. 

In my currently local testing for `rubicon-objc` the most code for `CoreBluetooth` is shared. So I don’t think that it should be listed as a complete new backend. The detection logic which binding framework is used can be inside the CoreBluetooth backend itself.

bleak/backends/scanner.py
```diff
@@ -286,20 +284,16 @@ def get_platform_scanner_backend_type() -> type[BaseBleakScanner]:
...
     """
-    if os.environ.get("P4A_BOOTSTRAP") is not None:
+    backend = get_backend()
+    if backend == BleakBackend.P4Android:
         from bleak.backends.p4android.scanner import BleakScannerP4Android
 
         return BleakScannerP4Android
-
-    if platform.system() == "Linux":
+    elif backend == BleakBackend.BlueZDBus:
```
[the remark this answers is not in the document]
timrid: Fixed it.

bleak/backends/__init__.py
```diff
@@ -1,5 +1,37 @@
...
+    WinRT = enum.auto()
+
+
+def get_backend() -> BleakBackend:
+    """Detect the backend to use for the current platform."""
```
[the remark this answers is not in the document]
timrid: Fixed

timrid:
To use `client._backend.__class__.__name__` looks more like a workaround than an official way to check the backend. But I see your point, that you can have a `BleakClient` or `BleakScanner` that uses a custom backend.

I implemented the following changes:
- I renamed `get_backend` to `get_default_backend` to make it more clear that this is only the default backend and not the one that is enforced for a `BleakClient` or `BleakScanner`.
- I added a `backend_id` parameter for `BleakClient` and `BleakScanner` that returns the enum to the current backend (or a string for a custom backend).

bleak/backends/__init__.py
```diff
@@ -1,5 +1,58 @@
...
-"""
+import enum
+import os
+import platform
+import sys
+
+from bleak.exc import BleakError
+
+
+class BleakBackend(str, enum.Enum):
```
dlech:
```suggestion
class BleakBackend(enum.StrEnum):
```

Unless I'm missing something?

dlech:
```suggestion
class BleakBackend(str, enum.Enum):
    """
    Identifiers for available built-in Bleak backends.
    
    .. version-added:: unreleased
```

bleak/backends/__init__.py
```diff
@@ -1,5 +1,58 @@
...
+    WinRT = "WinRT"
+    """
+    Windows Runtime backend for Windows.
+    """
+
+
+def get_default_backend() -> BleakBackend:
+    """
+    Returns the preferred backend for the current platform/environment.
+    """
```
dlech:
```suggestion
    Returns the preferred backend for the current platform/environment.
    
    .. version-added:: unreleased
    """
```

bleak/__init__.py
```diff
@@ -519,6 +535,17 @@ def __init__(
...
+        self._backend_id = backend_id
+
+    @property
+    def backend_id(self) -> BleakBackend | str:
+        """
+        Gets the identifier of the backend in use.
+
+        The value is one of the :class:`BleakBackend` enum values in case of
+        built-in backends, or a string identifying a custom backend.
+        """
```
```suggestion
        built-in backends, or a string identifying a custom backend.
        
        .. version-added:: unreleased
        """
```

dlech (commented):
Thanks for the update. I'm happy with the changes. Just a few more minor suggestions before we merge.

bleak/backends/__init__.py
dlech: > Added in version 3.11.

Oops, I am missing something. :facepalm:

bleak/backends/__init__.py
```diff
@@ -1,5 +1,58 @@
...
+import enum
+import os
+import platform
+import sys
+
+from bleak.exc import BleakError
+
+
+class BleakBackend(str, enum.Enum):
+    P4Android = "P4Android"
```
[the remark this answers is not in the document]
dlech: I guess we can't do `auto()` yet because it requires Python 3.11, but we could make the values lowercase if we want to leave open the possibility of changing it some day.

bleak/backends/__init__.py
timrid: Fixed

dlech:
https://github.com/hbldh/bleak/pull/1855 should fix the failing tests here.

dlech:
I added a bit to make the new stuff actually show up in the docs.

Ready to merge now, thanks!

PR #1842, opened 2025-10-16
Separate CoreBluetooth delegates from python Code in preparation for rubicon-objc

timrid:
When trying to add iOS via rubicon-objc (#1833) one big issue is, that rubicon-objc does not support python methods inside objc delegate classes.

I could not find a direct comment in the [rubicon-objc docs](https://rubicon-objc.readthedocs.io/en/stable/tutorial/tutorial-2.html) that this is forbidden, but it generates error when I tried it. I also saw in [toga](https://github.com/beeware/toga) (a GUI library from the same authors as rubicon-objc) that they strictly separate the objc-delegate from the python code. See e.g. [here](https://github.com/beeware/toga/blob/458ab046a250e6939c7b32431edcafc87a5d7f29/iOS/src/toga_iOS/hardware/location.py#L34-L64).

So to keep a future rubicon-objc PR smaller and simplify the future review, I separated the objc code from python code in the pyobjc backend in advance in this PR.

This may look like a lot of changes, but most of the code is just shifted around.

### DIFF
--- a/bleak/backends/corebluetooth/CentralManagerDelegate.py
+++ b/bleak/backends/corebluetooth/CentralManagerDelegate.py
@@ -181,9 +181,9 @@ class CentralManagerDelegate:
 
     def __init__(self) -> None:
         """macOS init function for NSObject"""
-        self.objc_delegate = ObjcCentralManagerDelegate.alloc().initWithPyDelegate_(
-            self
-        )
+        delegate = ObjcCentralManagerDelegate.alloc().initWithPyDelegate_(self)
+        assert delegate is not None
+        self.objc_delegate = delegate
 
         self.event_loop = asyncio.get_running_loop()
         self._connect_futures: dict[NSUUID, asyncio.Future[bool]] = {}


bleak/backends/corebluetooth/CentralManagerDelegate.py
```diff
@@ -61,22 +60,128 @@
...
 
-class CentralManagerDelegate(NSObject):
+class ObjcCentralManagerDelegate(NSObject):
+    """macOS conforming python class for managing the CentralManger for BLE"""
+
+    ___pyobjc_protocols__ = [CBCentralManagerDelegate]
+
+    def initWithPyDelegate_(
+        self, py_delegate: CentralManagerDelegate
+    ) -> Optional[ObjcCentralManagerDelegate]:
```
dlech:
```suggestion
    ) -> Optional[Self]:
```

I know we are never going to subclass this, but `Self` is still technically correct.

bleak/backends/corebluetooth/CentralManagerDelegate.py
```diff
@@ -61,22 +60,128 @@
 DisconnectCallback = Callable[[], None]
 
 
-class CentralManagerDelegate(NSObject):
+class ObjcCentralManagerDelegate(NSObject):
+    """macOS conforming python class for managing the CentralManger for BLE"""
+
+    ___pyobjc_protocols__ = [CBCentralManagerDelegate]
```
dlech:
```suggestion
class ObjcCentralManagerDelegate(NSObject, protocols=[CBCentralManagerDelegate]):
    """
    CoreBluetooth central manager delegate for bridging callbacks to asyncio.
    """
```

Since we are touching this anyway, I'd like to change to the "new" way of specifying the protocol.

bleak/backends/corebluetooth/PeripheralDelegate.py
```diff
@@ -49,25 +49,174 @@
 CBPeripheralDelegate = objc.protocolNamed("CBPeripheralDelegate")
 
 
-class PeripheralDelegate(NSObject):
+class ObjcPeripheralDelegate(NSObject):
     """macOS conforming python class for managing the PeripheralDelegate for BLE"""
 
     ___pyobjc_protocols__ = [CBPeripheralDelegate]
```
dlech:
```suggestion
class ObjcPeripheralDelegate(NSObject, protocols=[CBPeripheralDelegate]):
    """
    CoreBluetooth peripheral manager delegate for bridging callbacks to asyncio.
    """
```

bleak/backends/corebluetooth/CentralManagerDelegate.py
timrid: Even better. The [`protocols=` is the same syntax `rubicon-objc` is using](https://rubicon-objc.readthedocs.io/en/stable/reference/rubicon-objc-api.html#creating-custom-objective-c-classes-and-protocols).

bleak/backends/corebluetooth/PeripheralDelegate.py
timrid: Fixed

typings/Foundation/__init__.pyi
```diff
@@ -26,6 +26,8 @@ class NSObject:
         context: int,
     ) -> None: ...
     def removeObserver_forKeyPath_(self, observer: NSObject, keyPath: str) -> None: ...
+    @classmethod
+    def __init_subclass__(cls, /, protocols: Sequence[Any]) -> None: ...
```
dlech:
```suggestion
    def __init_subclass__(cls, /, protocols: Sequence[Any] = []) -> None: ...
```

bleak/backends/corebluetooth/CentralManagerDelegate.py
```diff
@@ -61,22 +62,128 @@
...
-    def init(self: Self) -> Optional[Self]:
+    def __init__(self) -> None:
         """macOS init function for NSObject"""
-        self = objc.super(CentralManagerDelegate, self).init()
-
-        if self is None:
-            return None
+        self.objc_delegate = ObjcCentralManagerDelegate.alloc().initWithPyDelegate_(
+            self
+        )
```
dlech:
```suggestion
        delegate = ObjcCentralManagerDelegate.alloc().initWithPyDelegate_(self)
        assert delegate is not None
        self.objc_delegate = delegate
```

bleak/backends/corebluetooth/PeripheralDelegate.py
```diff
@@ -49,25 +51,172 @@
...
+
+class PeripheralDelegate:
+    """macOS conforming python class for managing the PeripheralDelegate for BLE"""
+
+    def __init__(self, peripheral: CBPeripheral) -> None:
+        self.objc_delegate = ObjcPeripheralDelegate.alloc().initWithPyDelegate_(self)
+
         self.peripheral = peripheral
-        self.peripheral.setDelegate_(self)
+        self.peripheral.setDelegate_(self.objc_delegate)
```
dlech:
```suggestion
        self.peripheral = peripheral
        delegate = ObjcPeripheralDelegate.alloc().initWithPyDelegate_(self)
        assert delegate is not None
        self.objc_delegate = delegate
        self.peripheral.setDelegate_(self.objc_delegate)
```

PR #1848, opened 2025-10-25
bleak: use Python 3.10 type unions

dlech:
Since we now require Python 3.10+, we can use the new union type syntax e.g. str | None of using typing.Optional or typing.Union.


dlech:
Probably should wait on merging this one in order to not conflict with other ongoing work.

dlech:
CI is failing because of https://github.com/sphinx-doc/sphinx/issues/11211.

PR #1849, opened 2025-10-26
Implement checks for Bluetooth support in WinRT backend

dlech:
Continuation of #1847 and #1832, this time for the Windows backend.


dlech:
I don't know an easy way to reproduce a permission denied error on Windows, so that isn't implemented yet. It would probably require making a packaged Windows app with Python embedded in it. Not sure if anyone is actually doing that. So we'll just save that for later.

PR #1853, opened 2025-10-29
Add support for `rubicon-objc`

### LINKED ISSUE #1833: Add support for iOS (add rubicon-objc)

timrid:
This adds support for [`rubicon-objc`](https://github.com/beeware/rubicon-objc) to potentially support iOS. Currently I only tested this on macOS and not on iOS.

On macOS it is possible to activate the `rubicon-objc` framework by setting the environment variable `BLEAK_COREBLUETOOTH_FRAMEWORK` to `rubicon-objc`.

I just wanted to share this as a first draft until I actual did tests with an iPhone.

Fixes #1833

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -34,7 +34,7 @@ jobs:
           uv tool install poetry
 
       - name: Test with pytest
-        run: tox -e py -- -v --cov-report=xml --junitxml=junit.xml -o junit_family=legacy
+        run: tox -m py${{ matrix.python-version }} -- -v --cov-report=xml --junitxml=junit.xml -o junit_family=legacy
 
       - name: Upload coverage reports to Codecov
         uses: codecov/codecov-action@v5
@@ -56,7 +56,7 @@ jobs:
         # on both the platform and the Python version in the code, so we
         # need to check each matrix combination.
         run: |
-          tox -e typecheck
+          tox -m typecheck
 
 
   integration_tests_bluez:
--- a/mypy-rubicon-objc.ini
+++ b/mypy-rubicon-objc.ini
@@ -1,0 +1,9 @@
+[mypy]
+python_version = 3.10
+disable_error_code = import-not-found
+exclude = (.venv|kivy|recipes)/
+mypy_path = typings
+packages = bleak, tests, examples
+
+always_true = BLEAK_OBJC_FRAMEWORK_IS_RUBICON
+always_false = BLEAK_OBJC_FRAMEWORK_IS_PYOBJC
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,9 @@
+[mypy]
+python_version = 3.10
+disable_error_code = import-not-found
+exclude = (.venv|kivy|recipes)/
+mypy_path = typings
+packages = bleak, tests, examples
+
+always_true = BLEAK_OBJC_FRAMEWORK_IS_PYOBJC
+always_false = BLEAK_OBJC_FRAMEWORK_IS_RUBICON
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -78,17 +78,6 @@ py_version=310
 src_paths = ["bleak", "examples", "tests"]
 extend_skip = [".buildozer", "docs"]
 
-[tool.mypy]
-python_version = "3.10"
-disable_error_code = ["import-not-found"]
-exclude = "(.venv|kivy|recipes|.tox)/"
-mypy_path = "typings"
-packages = ["bleak", "tests", "examples"]
-
-[tool.pyright]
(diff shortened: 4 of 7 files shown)


dlech:
I was hoping for something a bit less invasive to the existing code, e.g. creating a `_pyobjc_compat/` folder with modules having the same name and containing classes that have the same API as PyObjC that calls into the equivalent Rubicon API.

Then the existing code would just have to be modified like:

```python
try:
    from CoreBluetooth import CBUUID, ...
except ImportError:
    from ._pyobjc_compat.CoreBluetooth import CBUUID, ...
```

dlech:
> Maybe I could split this PR in smaller chunks to simplify the review? Would that be a compromise?

Smaller chunks is always helpful.

I think the only thing here that I will really strongly object to is changing the files in `typings` to be for anything other than PyObjC (fixes are OK, of course). 

Rubicon ObjC should have it's own separate typings. I was using it for USB stuff recently and so I have some idea of how I would want to do this, so I don't mind helping with that part.

timrid:
I have rebased the PR to the current version and built it based on https://github.com/hbldh/bleak/pull/1922. You can use
`tox -- --bleak-hci-transport=serial:/dev/tty.usbmodem1401` to test both `pyobjc` and `rubicon-objc` completely. This means unit tests and type checks for both versions. It takes a while, but then you can be sure that everything still works.

I would adapt the documentation afterwards, once the details have been clarified. I also removed everything specific to iOS. I will do that later via a separate PR.

bleak/backends/corebluetooth/_objc_compat.py
```diff
@@ -0,0 +1,322 @@
...
+    def to_int(objc_obj: NSNumber) -> int:
+        """Converts an NSNumber to a Python int."""
+        ret = py_from_ns(objc_obj)
+        if not isinstance(ret, int):
+            raise TypeError(f"Expected int, got {type(ret)}")
+        return ret
+
+    def to_str(objc_obj: NSString) -> str:
+        """Converts an NSString to a Python string."""
+        return py_from_ns(objc_obj)
```
johnzhou721, line 238: strs autoconverts in rubicon-objc as well.  So this shim is not needed.  Is this for consistency?

bleak/backends/corebluetooth/_objc_compat.py
```diff
@@ -0,0 +1,322 @@
...
+    from rubicon.objc.api import objc_method as _objc_method
+
+    F = TypeVar("F", bound=Callable[..., Any])
+
+    def objc_method(func: F) -> F:
+        """Decorator for Objective-C methods."""
+        # Normally rubicon-objc parses the type hints from `func` for e.g. ctypes types, but the type hints
+        # we use are not compatible. So we remove all type hints with `no_type_check` before passing the
+        # function to rubicon-objc.
+        return cast(F, _objc_method(no_type_check(func)))
```
johnzhou721, line 298: If we don't have type-check, return values, etc. are going to be very inconsistently handled.  All return values are None, and all parameters are of type `id`.

Does there exist any other functions other than this type of signatures in the codebase?  If so, I think it's not too hard to convert the type hints used in the codebase to actual ctypes type hints.

PR #1855, opened 2025-11-04
backends/corebluetooth: check run loop running before call_soon_threadsafe()

dlech:
Add checks to all delegate callbacks to make sure the asyncio run loop is still running before attempting to call call_soon_threadsafe().

In rare cases, one of these can be triggered while Python is shutting down causing the process to crash rather than exiting successfully.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,7 @@ Changed
 
 Fixed
 -----
+* Fixed crash in CoreBluetooth backend if an ObjC delegate callback is called after the asyncio run loop stops.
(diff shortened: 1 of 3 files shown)


timrid:
Is it really nessesary to always check this? The CI errors e.g. in [here](https://github.com/hbldh/bleak/actions/runs/19050012847/job/54407665831) are all failing just in `observeValueForKeyPath_ofObject_change_context_`. And I would guess that this is because of [this](https://github.com/hbldh/bleak/commit/67b5371819902b948c11ac3756763569bed26e54) change.

dlech:
Updated using the try/except pattern.

PR #1856, opened 2025-11-06
Bluez notify read fix

TristanOpbroek:
I ran into a problem using the BlueZ backend whereas when notifying on a characteristic and triggering a read, the data from the response comes back twice. Once as a read, and once as a notification. This occurs in instances where a peripheral sends a notification which requires a reread.

I wrote this test script to show the error:

```python
DEVICE_NAME = "DEVICE_NAME_HERE"
CHAR_UUID = "CHAR_UUID_HERE"

def is_notification(data):
    return len(data) == 1 and data[0] == 0xAA

async def run_ble_test():
    print(f"Scanning for {DEVICE_NAME}")
    devices = await BleakScanner.discover(timeout=5.0)
    target = None
    for d in devices:
        if d.name == DEVICE_NAME:
            target = d
            break

    if target is None:
        print(f"Device '{DEVICE_NAME}' not found.")
        return

    async with BleakClient(target.address) as client:
        print(f"Connected to {target.name} ({target.address})")

        async def notification_handler(sender: str, data: bytearray):
            print(f"Notification from {sender}: {data.hex()}")
            # If notification, trigger re-read
            if is_notification(data):
                new_data = await client.read_gatt_char(CHAR_UUID)
                print(f"Notification re-read triggered: {new_data.hex()}")

        await client.start_notify(CHAR_UUID, notification_handler)

        print("Listening for notifications (10 seconds)...")
        await asyncio.sleep(10.0)

        print("Stopping notifications...")
        await client.stop_notify(CHAR_UUID)


def test_bleak():
    asyncio.run(run_ble_test())


if __name__ == "__main__":
    test_bleak()

```

On MacOS with a peripheral I can trigger notifications on, I get a single notification, and then single re-read (which is expected):

```
Listening for notifications (10 seconds)...
Notification from <CHAR_UUID_HERE> (Handle: 56): Unknown: aa
Notification re-read triggered: <DATA>
Stopping notifications...
```

On my Linux machine however, I receive the data from the characteristics in both the re-read, and as an additional notification:

```
Listening for notifications (10 seconds)...
Notification from <CHAR_UUID_HERE> (Handle: 56): Unknown: aa
Notification from <CHAR_UUID_HERE> (Handle: 56): Unknown: <DATA>
Notification re-read triggered: <DATA>
Stopping notifications...
```

This change adds a 'notification_discriminator' argument to Bleak's start_notify method to suppress notifications which don't pass the discriminator.

### DIFF
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -1008,15 +1008,31 @@ class BleakClientBlueZDBus(BaseBleakClient):
         assert self._bus is not None
 
         if "NotifyAcquired" in characteristic.obj[1]:
+            logger.debug(
+                "Closing notification fd for characteristic %d", characteristic.handle
+            )
             fd = self._notification_fds.pop(characteristic.obj[0], None)
-            if fd is not None:
+
+            if fd is None:
+                logger.debug(
+                    "No notification fd found for characteristic %d",
+                    characteristic.handle,
+                )
+            else:
                 loop = asyncio.get_running_loop()
                 try:
                     loop.remove_reader(fd)
+                except RuntimeError:
+                    # Run loop is closed
+                    pass
+                try:
                     os.close(fd)
-                except Exception as e:
-                    logger.debug("Failed to remove file descriptor %d: %d", fd, e)
+                except OSError as e:
+                    logger.debug("Failed to remove file descriptor %d: %s", fd, e)
         else:
+            logger.debug(
+                "Calling StopNotify for characteristic %d", characteristic.handle
+            )
             reply = await self._bus.call(
                 Message(
                     destination=defs.BLUEZ_SERVICE,


dlech:
> ```python
> if is_notification(data):
> ```

Ah, I see you built the discriminator into the notification callback instead of using the one in `CBStartNotifyArgs`. So if this works on Mac but doesn't work on Linux, I'm not sure how adding a notification discriminator on Linux would work. It should be doing the exact same thing.

dlech:
> It should be doing the exact same thing.

Oh right, in the CoreBluetooth backend, we assume that if there is a pending read request then the changed value must be the response to the read request and don't send the notification in that case. However, that isn't 100% foolproof without a discriminator since a notification could be received after the read request but before the read response.

dlech (commented):
Nice, thanks for updating!

I made a few quick comments, then I would like to try it out when I have a chance.

bleak/backends/bluezdbus/client.py
```diff
@@ -904,6 +916,27 @@ async def write_gatt_descriptor(
...
+        os.set_blocking(fd, False)
+
+        def on_data():
+            try:
+                data = os.read(fd, 1024)
+                if not data:  # EOF, close file descriptor
+                    os.close(fd)
+                    return
+                callback(bytes(data))
+            except Exception as e:
```
dlech: We don't catch exceptions from notification callbacks anywhere else. We just let it crash the program since it is considered a bug in the program. So I would take that out of try/except here.

bleak/backends/bluezdbus/client.py
```diff
@@ -904,6 +916,27 @@ async def write_gatt_descriptor(
             "Write Descriptor %s | %s: %s", descriptor.handle, descriptor.obj[0], data
         )
 
+    def _read_notify_fd(self, fd, callback):
+        loop = asyncio.get_running_loop()
+        os.set_blocking(fd, False)
```
dlech: We shouldn't actually need to make it non-blocking since this callback is only called when there is actually data ready.

bleak/backends/bluezdbus/client.py
```diff
@@ -904,6 +916,27 @@ async def write_gatt_descriptor(
...
+
+        def on_data():
+            try:
+                data = os.read(fd, 1024)
+                if not data:  # EOF, close file descriptor
+                    os.close(fd)
+                    return
+                callback(bytes(data))
+            except Exception as e:
+                logger.error("AcquireNotify: Read error on fd %d: %s", fd, e)
```
```suggestion
                logger.error("AcquireNotify: Read error on fd %d: %s. Notifications have been stopped.", fd, e)
```

bleak/backends/bluezdbus/client.py
```diff
@@ -904,6 +916,27 @@ async def write_gatt_descriptor(
...
         )
 
+    def _read_notify_fd(self, fd, callback):
+        loop = asyncio.get_running_loop()
+        os.set_blocking(fd, False)
+
+        def on_data():
+            try:
+                data = os.read(fd, 1024)
+                if not data:  # EOF, close file descriptor
```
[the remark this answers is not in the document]
TristanOpbroek: I'll keep it like this and raise just to make sure its removed from the loop.

bleak/backends/bluezdbus/client.py
```diff
@@ -904,6 +916,29 @@ async def write_gatt_descriptor(
             "Write Descriptor %s | %s: %s", descriptor.handle, descriptor.obj[0], data
         )
 
+    def _read_notify_fd(self, fd, callback):
```
```suggestion
    def _register_notify_fd_reader(self, fd: int, callback: : NotifyCallback) -> None:
```

bleak/backends/bluezdbus/client.py
```diff
@@ -914,20 +949,44 @@ async def start_notify(
...
             )
-        )
-        assert reply
-        assert_reply(reply)
+            assert reply
+            assert_reply(reply)
+
+            unix_fd = reply.unix_fds[0]
+            self._notification_fds[characteristic.obj[0]] = unix_fd
+            self._read_notify_fd(unix_fd, callback)
```
```suggestion
            self._register_notify_fd_reader(unix_fd, callback)
```

bleak/backends/bluezdbus/client.py
```diff
@@ -904,6 +916,29 @@ async def write_gatt_descriptor(
...
+                logger.error(
+                    "AcquireNotify: Read error on fd %d: %s. Notifications have been stopped.",
+                    fd,
+                    e,
+                )
+                try:
+                    loop.remove_reader(fd)
+                    os.close(fd)
+                except OSError:
+                    pass
```
```suggestion
                try:
                    loop.remove_reader(fd)
                except RuntimeError:
                    # run loop is closed
                    pass
                try:
                    os.close(fd)
                except OSError:
                    # bad file descriptor
                    pass
```

dlech: Although doing this here without removing `fd` from `self._notification_fds` is problematic when it comes to later calls.

bleak/backends/bluezdbus/client.py
```diff
@@ -942,15 +1001,24 @@ async def stop_notify(self, characteristic: BleakGATTCharacteristic) -> None:
...
         assert self._bus is not None
 
-        reply = await self._bus.call(
-            Message(
-                destination=defs.BLUEZ_SERVICE,
-                path=characteristic.obj[0],
-                interface=defs.GATT_CHARACTERISTIC_INTERFACE,
-                member="StopNotify",
+        if "NotifyAcquired" in characteristic.obj[1]:
+            fd = self._notification_fds.pop(characteristic.obj[0])
```
```suggestion
            fd = self._notification_fds.pop(characteristic.obj[0], None)
```

I think I was wrong in my previous comment and we do need `None` here to avoid `KeyError` if there is no match.

bleak/backends/bluezdbus/client.py
TristanOpbroek: Good suggestion, the name wasn't very clear.

PR #1864, opened 2025-11-28
Add support for pairing callbacks

tgagneret-embedded:
Add support for pairing callbacks.

This is based on:

- https://github.com/hbldh/bleak/pull/1133
- https://github.com/hbldh/bleak/pull/1100

### DIFF
--- a/bleak/agent.py
+++ b/bleak/agent.py
@@ -1,0 +1,27 @@
+import abc
+
+from .backends.device import BLEDevice
+
+
+class BaseBleakAgentCallbacks(abc.ABC):
+    @abc.abstractmethod
+    async def request_passkey(self, device: BLEDevice) -> str:
+        """
+        Implementers should ask the user the passkey displayed on the device.
+
+        Returns:
+            Passkey provided by the user
+        """
+
+    @abc.abstractmethod
+    async def confirm_passkey(self, device: BLEDevice, passkey: str) -> bool:
+        """
+        Implementers should display the passkey code to the user and prompt the
+        user to validate the passkey code and confirm or reject the pairing request.
+
+        Args:
+            pin: The passkey to be confirmed.
+
+        Returns:
+            ``True`` to accept the pairing request or ``False`` to reject it.
+        """
--- a/bleak/backends/bluezdbus/agent.py
+++ b/bleak/backends/bluezdbus/agent.py
@@ -1,0 +1,210 @@
+"""
+Agent
+-----
+
+This module contains types associated with the BlueZ D-Bus `agent api
+<https://github.com/bluez/bluez/blob/master/doc/agent-api.txt>`.
+"""
+
+import asyncio
+import contextlib
+import logging
+import os
+from typing import Set, no_type_check
+
+from dbus_fast import DBusError, Message
+from dbus_fast.aio import MessageBus
+from dbus_fast.service import ServiceInterface, method
+from dbus_fast.signature import Variant
+
+from bleak.agent import BaseBleakAgentCallbacks
+from bleak.backends.bluezdbus import defs
+from bleak.backends.bluezdbus.manager import get_global_bluez_manager
+from bleak.backends.bluezdbus.utils import assert_reply
+from bleak.backends.device import BLEDevice
+
+logger = logging.getLogger(__name__)
+
+
+class Agent(ServiceInterface):
+    """
+    Implementation of the org.bluez.Agent1 D-Bus interface.
+    """
+
+    def __init__(self, callbacks: BaseBleakAgentCallbacks):
+        """
+        Args:
+        """
+        super().__init__(defs.AGENT_INTERFACE)
+        self._callbacks = callbacks
+        self._tasks: Set[asyncio.Task] = set()
+
+    @staticmethod
+    async def _create_ble_device(device_path: str) -> BLEDevice:
+        manager = await get_global_bluez_manager()
+        return BLEDevice(
+            manager.get_device_address(device_path),
+            manager.get_device_name(device_path),
+            {"path": device_path},
+        )
+
+    @method()
+    def Release(self):  # noqa: N802
+        logger.debug("Release")
+
+    # REVISIT: mypy is broke, so we have to add redundant @no_type_check
+    # https://github.com/python/mypy/issues/6583
+
+    @method()
+    @no_type_check
+    async def RequestPinCode(self, device: "o") -> "s":  # noqa: F821 N802
+        logger.debug("RequestPinCode %s", device)
+        raise NotImplementedError
+
+    @method()
+    @no_type_check
+    async def DisplayPinCode(self, device: "o", pincode: "s"):  # noqa: F821 N802
+        logger.debug("DisplayPinCode %s %s", device, pincode)
+        raise NotImplementedError
+
+    @method()
+    @no_type_check
+    async def RequestPasskey(self, device: "o") -> "u":  # noqa: F821 N802
+        logger.debug("RequestPasskey %s", device)
+
+        ble_device = await self._create_ble_device(device)
+
+        task = asyncio.create_task(self._callbacks.request_passkey(ble_device))
+        self._tasks.add(task)
+
+        try:
+            key = await task
+        except asyncio.CancelledError:
+            raise DBusError("org.bluez.Error.Canceled", "task canceled")
+        finally:
+            self._tasks.remove(task)
+
+        if not key:
+            raise DBusError("org.bluez.Error.Rejected", "user rejected")
+
+        try:
+            passkey = int(key)
+        except ValueError:
+            raise DBusError("org.bluez.Error.Rejected", "invalid passkey")
+
+        return passkey
+
+    @method()
+    @no_type_check
+    async def DisplayPasskey(  # noqa: N802
+        self, device: "o", passkey: "u", entered: "q"  # noqa: F821
+    ):
+        passkey = f"{passkey:06}"
+        logger.debug("DisplayPasskey %s %s %d", device, passkey, entered)
+        raise NotImplementedError
+
+    @method()
+    @no_type_check
+    async def RequestConfirmation(self, device: "o", passkey: "u"):  # noqa: F821 N802
+        passkey = f"{passkey:06}"
+        logger.debug("RequestConfirmation %s %s", device, passkey)
+
+        ble_device = await self._create_ble_device(device)
+
+        task = asyncio.create_task(self._callbacks.confirm_passkey(ble_device, passkey))
+        self._tasks.add(task)
+
+        try:
+            result = await task
+        except asyncio.CancelledError:
+            raise DBusError("org.bluez.Error.Canceled", "task canceled")
+        finally:
+            self._tasks.remove(task)
+
+        if not result:
+            raise DBusError("org.bluez.Error.Rejected", "user rejected")
+        else:
+            manager = await get_global_bluez_manager()
+            # Set device as trusted.
+            reply = await manager._bus.call(
+                Message(
+                    destination=defs.BLUEZ_SERVICE,
+                    path=device,
+                    interface=defs.PROPERTIES_INTERFACE,
+                    member="Set",
+                    signature="ssv",
+                    body=[defs.DEVICE_INTERFACE, "Trusted", Variant("b", True)],
+                )
+            )
+            assert reply
+            assert_reply(reply)
+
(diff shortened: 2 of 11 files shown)


bleak/backends/bluezdbus/client.py
```diff
@@ -257,36 +264,7 @@ async def disconnect_device() -> None:
...
-                            # Pairing means device is authenticated
-                            reply = await self._bus.call(
-                                Message(
-                                    destination=defs.BLUEZ_SERVICE,
-                                    interface=defs.DEVICE_INTERFACE,
-                                    path=self._device_path,
-                                    member="Pair",
-                                )
-                            )
+                            reply = await self._pair(self._pairing_callbacks)
```
dlech: IIRC, there was some slight difference between the pairing here and in `self.pair()`. I'm guessing that is why `self._pair()` is introduced. 

Can we pull this change out into a separate pull request to make it easier to look at on it's own?

bleak/backends/bluezdbus/manager.py
```diff
@@ -780,6 +780,20 @@ async def get_services(
 
         return services
 
+    def get_device_props(self, device_path: str) -> Device1:
```
dlech: I would prefer to stick with having methods to get individual properties using the `_get_device_property()` so that we get a more helpful error message when things go wrong.

examples/pairing_agent.py
```diff
@@ -0,0 +1,85 @@
...
+from bleak.exc import (
+    BleakDeviceNotFoundError,
+    BleakPairingCancelledError,
+    BleakPairingFailedError,
+)
+
+
+class AgentCallbacks(BaseBleakAgentCallbacks):
+    def __init__(self) -> None:
+        super().__init__()
```
dlech:
```suggestion
```

This should not be needed.

dlech (commented):
Thanks for getting this moving again. I've made some preliminary comments. It will take me some time to go back and look at the previous discussions to refresh my memory on all of the details that have been looked at before. And I'll need some time to put something together for testing.

In the meantime, it looks like there is perhaps some preliminary refactoring that could be split out into a separate pull request to keep things moving.

EuleMitKeule:
@tgagneret-embedded I opened a PR against your `feature/pairing` branch with the changes that @dlech requested here (except the one about moving the `_pair` method into a separate PR). Hope that helps.

https://github.com/tgagneret-embedded/bleak/pull/1

dlech:
FYI, we now have some new test infrastructure that it should be possible to write tests and get full coverage for anything we add here (since this is BlueZ-only). You don't even need real Bluetooth hardware for this, so it should make it a lot easier now.

But that does mean that I will be expecting tests to cover possibilities like Bleak initiating the pairing request, the peripheral initiating the pairing request, different paring types, etc.

bleak/backends/bluezdbus/client.py
```diff
@@ -454,32 +470,34 @@ async def _pair(self) -> Message:
...
-        # Trusted means device is authorized
-        reply = await self._bus.call(
-            Message(
-                destination=defs.BLUEZ_SERVICE,
-                path=self._device_path,
-                interface=defs.PROPERTIES_INTERFACE,
-                member="Set",
-                signature="ssv",
-                body=[defs.DEVICE_INTERFACE, "Trusted", Variant("b", True)],
+        if callbacks is None:
```
dlech: Why does this depend on callbacks?

Maybe we should move this after the call to "Pair"?

bleak/backends/bluezdbus/agent.py
```diff
@@ -0,0 +1,210 @@
...
+
+    @method()
+    @no_type_check
+    async def RequestConfirmation(self, device: "o", passkey: "u"):  # noqa: F821 N802
+        passkey = f"{passkey:06}"
+        logger.debug("RequestConfirmation %s %s", device, passkey)
+
+        ble_device = await self._create_ble_device(device)
+
+        task = asyncio.create_task(self._callbacks.confirm_passkey(ble_device, passkey))
```
dlech: Could use a comment explaining that we have to create a task so that it can be canceled by a different callback.

bleak/backends/bluezdbus/agent.py
```diff
@@ -0,0 +1,210 @@
...
+        try:
+            result = await task
+        except asyncio.CancelledError:
+            raise DBusError("org.bluez.Error.Canceled", "task canceled")
+        finally:
+            self._tasks.remove(task)
+
+        if not result:
+            raise DBusError("org.bluez.Error.Rejected", "user rejected")
+        else:
```
dlech: Don't need `else` since `if` always raises.

bleak/backends/bluezdbus/agent.py
```diff
@@ -0,0 +1,210 @@
...
+                Message(
+                    destination=defs.BLUEZ_SERVICE,
+                    path=device,
+                    interface=defs.PROPERTIES_INTERFACE,
+                    member="Set",
+                    signature="ssv",
+                    body=[defs.DEVICE_INTERFACE, "Trusted", Variant("b", True)],
+                )
+            )
+            assert reply
```
dlech:
```suggestion
```

No longer needed to make the type checker happy.

bleak/backends/bluezdbus/client.py
```diff
@@ -374,6 +386,10 @@ def _cleanup_all(self) -> None:
         """
         logger.debug("_cleanup_all(%s)", self._device_path)
 
+        # Reset all stored services.
```
dlech: This looks like something for a separate pull request.

bleak/backends/winrt/client.py
```diff
@@ -196,6 +196,13 @@ def __init__(
         self._session_status_changed_token: Optional[EventRegistrationToken] = None
         self._max_pdu_size_changed_token: Optional[EventRegistrationToken] = None
 
+        if kwargs.get("pairing_callbacks"):
+            logger.warning(
```
dlech: This should raise `NotImplementedError` since we intend to implement it.

bleak/backends/bluezdbus/client.py
```diff
@@ -112,6 +118,10 @@ def __init__(
         # used to override mtu_size property
         self._mtu_size: Optional[int] = None
 
+        self._pairing_callbacks: Optional[BaseBleakAgentCallbacks] = kwargs.get(
+            "pairing_callbacks"
+        )
+
```
```suggestion
```

This is redundant since `super().__init__()` already sets it.

bleak/agent.py
```diff
@@ -0,0 +1,27 @@
+import abc
+
+from .backends.device import BLEDevice
+
+
+class BaseBleakAgentCallbacks(abc.ABC):
+    @abc.abstractmethod
```
dlech: Making methods abstract here isn't going to be future-proof. 

We should consider what it would take to implement the various capabilities, e.g. if you claim to support capability X then you need to implement method y().

All methods can have a default implementation that raises `NotImplementedError`. And we can make an `__init__(self, capabilities: BleakAgentFlags)` method that verifies that the required methods have been overridden.

bleak/agent.py
```diff
@@ -0,0 +1,27 @@
+import abc
+
+from .backends.device import BLEDevice
+
+
+class BaseBleakAgentCallbacks(abc.ABC):
```
dlech: This needs a doc comment that includes an overview.

Some important points:
* Method implementations that `await` need to consider that an `asyncio.CancelledError` can be raised on any coroutine that is called.
* Include `.. versionadded:: unreleased` (will be updated with actual version on release).

bleak/backends/bluezdbus/agent.py
```diff
@@ -0,0 +1,210 @@
+"""
+Agent
+-----
+
+This module contains types associated with the BlueZ D-Bus `agent api
```
dlech:
```suggestion
This module contains types associated with the BlueZ D-Bus `org.bluez.Agent1 API
```

bleak/backends/bluezdbus/agent.py
```diff
@@ -0,0 +1,210 @@
...
+<https://github.com/bluez/bluez/blob/master/doc/agent-api.txt>`.
```
dlech:
```suggestion
<https://github.com/bluez/bluez/blob/master/doc/agent-api.txt>`_.
```

I think that is the correct rst format, but I always forget and have to look it up.

bleak/backends/bluezdbus/agent.py
```diff
@@ -0,0 +1,210 @@
...
+    @method()
+    @no_type_check
+    async def RequestAuthorization(self, device: "o"):  # noqa: F821 N802
+        logger.debug("RequestAuthorization %s", device)
+        raise NotImplementedError
+
+    @method()
+    @no_type_check
+    async def AuthorizeService(self, device: "o", uuid: "s"):  # noqa: F821 N802
+        logger.debug("AuthorizeService %s", device, uuid)
```
dlech:
```suggestion
        logger.debug("AuthorizeService %s %s", device, uuid)
```

bleak/backends/bluezdbus/agent.py
```diff
@@ -0,0 +1,210 @@
...
+    async def DisplayPinCode(self, device: "o", pincode: "s"):  # noqa: F821 N802
+        logger.debug("DisplayPinCode %s %s", device, pincode)
+        raise NotImplementedError
+
+    @method()
+    @no_type_check
+    async def RequestPasskey(self, device: "o") -> "u":  # noqa: F821 N802
+        logger.debug("RequestPasskey %s", device)
+
+        ble_device = await self._create_ble_device(device)
```
dlech: Hmm... this is slightly dangerous because "Cancel" could be called if this actually awaits. In that case we would miss the cancellation.

Although the only time this should actually do that is if we lost the D-Bus connection, which means we have bigger problems anyway.

I suppose just adding a comment explaining this would suffice.

Alternative could be to pass `manager` to the constructor so that we have a reference to use without awaiting.

bleak/backends/bluezdbus/agent.py
```diff
@@ -0,0 +1,210 @@
+"""
+Agent
+-----
+
+This module contains types associated with the BlueZ D-Bus `agent api
+<https://github.com/bluez/bluez/blob/master/doc/agent-api.txt>`.
+"""
+
```
dlech:
```suggestion

import sys
from typing import TYPE_CHECKING

from bleak.args.bluez import BlueZNotifyArgs

if TYPE_CHECKING:
    if sys.platform != "linux":
        assert False, "This backend is only available on Linux"

```

Needed to avoid type checking failures on other platforms.

examples/pairing_agent.py
```diff
@@ -0,0 +1,108 @@
...
+
+    parser.add_argument(
+        "--unpair", action="store_true", help="unpair first before pairing"
+    )
+    parser.add_argument(
+        "--auto", action="store_true", help="automatically pair during connect"
+    )
+    args = parser.parse_args()
+
+    asyncio.run(main(args.address, args.name, args.unpair, args.auto))
```
dlech:
```suggestion
    asyncio.run(main(args))
```

Then we don't have to change it every time we add an arg.

examples/pairing_agent.py
```diff
@@ -0,0 +1,108 @@
...
+                await client.pair()
+                print("pairing successful")
+            except BleakPairingCancelledError:
+                print("paring was canceled")
+            except BleakPairingFailedError:
+                print("pairing failed (bad pin?)")
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser("pairing_agent.py")
```
dlech:
```suggestion
    parser = argparse.ArgumentParser(namespace=Args())
```

Similar to other examples.

dlech:
> But that does mean that I will be expecting tests to cover possibilities like Bleak initiating the pairing request,

Another thing we could do first (in a separate PR) is add some thorough tests for the existing pairing so that we can be sure we don't break anything that was already working with these changes. There are a couple of basic pairing tests in #1888, but we could split those out to a separate file and expand on that.

mohamedkallel82:
@dlech any news about this feature? when will be merged?
we need such feature in Home Assistant since our ble device require '(yes/no) question in the connect process

dlech:
This PR needs more work before we can merge it.

Now that we have integration tests, there is no reason to not have full coverage on this.

There has been quite a bit of other activity in Bleak that has been keeping me busy, so someone else will need to do the work here or we can just keep waiting.

PR #1865, opened 2025-11-29
Remove Python 2 remnants

gudnimg:
1. There is no need to set file encoding unless its not utf-8 (See https://docs.python.org/3/reference/lexical_analysis.html#encoding-declarations)
2. Calls to super() may omit the class name in Python 3. On Python 2 these were required.

No change in functionality.


dlech:
Thanks!

PR #1869, opened 2025-12-09
Remove side effects in bluez unit tests

timrid:
While working on #1838, I noticed that the bluez version tests have side effects on other tests, as the version is written to global variables that life beyond the actual tests. Therefore, I now use `monkeypatch` to set the corresponding global variables to a predefined value. Any changes during the test will be automatically reset after it. This means that the variables will not be set to test values for subsequent integration tests.

In addition, the organization of the unit tests is somewhat confusing, because corebluetooth is also a backend but not in the backend folder. I have also corrected this here.

### DIFF
--- a/tests/backends/bluezdbus/test_version.py
+++ b/tests/backends/bluezdbus/test_version.py
@@ -2,7 +2,7 @@
 
 """Tests for `bleak.backends.bluezdbus.version` package."""
 
-from unittest.mock import AsyncMock, Mock, patch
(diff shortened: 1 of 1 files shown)


dlech:
Can you please fill out the commit messages with *why* we need these changes?

The explanation above is fine for the reorg commit. I would like to know more about why `with patch` didn't work though and which global variables were not reset after a test.

PR #1888, opened 2025-12-28
Basic integrationtests for `BleakClient`

timrid:
This PR adds basic integration tests for the `BleakClient`.

Unfortunately, there is a problem in the Alpine VM where terminating a `BleakClient` connection always takes 20 seconds. This greatly increases the testing time in CI... Locally on my Ubuntu, terminating the connection always takes a constant 4 seconds. I find that long too, but it's still reasonably tolerable. However, I can't explain why it takes 20 seconds in Alpine VM. @dlech, do you have any idea what could be causing this?

My last commit adds additional logging output, enables the log output in CI and only executes one test case because the log output is huge... Searching for `Disconnected from device after` helps :) I will revert that changes once the problem is found.

E.g [this run](https://github.com/hbldh/bleak/actions/runs/20546080473/job/59016633745?pr=1888) outputs:

```
00:02:28 [INFO] Disconnected from device after 20.12 seconds.
```

### DIFF
--- a/tests/integration/test_client_properties.py
+++ b/tests/integration/test_client_properties.py
@@ -1,3 +1,6 @@
+import sys
+from typing import TYPE_CHECKING
+
 import pytest
 from bumble.device import Device
 
@@ -20,6 +23,9 @@ async def test_get_rssi(bumble_peripheral: Device):
     device = await find_ble_device(bumble_peripheral)
 
     async with BleakClient(device) as client:
+        if TYPE_CHECKING:
+            if sys.platform != "darwin":
+                assert False, "This backend is only available on macOS"
         from bleak.backends.corebluetooth.client import BleakClientCoreBluetooth
 
         backend = client._backend  # pyright: ignore[reportPrivateUsage]


dlech:
> do you have any idea what could be causing this?

According to the BlueZ source code, it should only be 2 seconds.

```c
#define DISCONNECT_TIMER	2

...

void device_request_disconnect(struct btd_device *device, DBusMessage *msg)
{
	...
	device->disconn_timer = timeout_add_seconds(DISCONNECT_TIMER,
							disconnect_all,
							device, NULL);
	...
}
```

There are 3 possible implementations for `timeout_add_seconds()` though depending on which run loop is being used (ell, glib or mainloop). So maybe whichever implementation is being used on alpine isn't working well? Or maybe the timer source in qemu isn't working well?

tests/integration/test_client.py
```diff
@@ -0,0 +1,363 @@
...
+    from async_timeout import timeout as async_timeout
+else:
+    from asyncio import timeout as async_timeout
+
+
+async def test_client_connect(bumble_peripheral: Device):
+    """Connecting to a BLE device is possible."""
+    add_default_advertising_data(bumble_peripheral)
+    await bumble_peripheral.power_on()
+    await bumble_peripheral.start_advertising()
```
dlech: These 3 lines are getting repeated a bunch. Maybe time to make a utility function for it?

tests/integration/test_client.py
```diff
@@ -0,0 +1,363 @@
...
+
+    device = await BleakScanner.find_device_by_name(bumble_peripheral.name)
+    assert device
```
dlech:
```suggestion
    if device is None:
        raise RuntimeError("failed to discover device, is Bumble working?")
```

This isn't what we are testing, so assert could be misleading.

This could be included in the utility function as well.

tests/integration/test_client.py
```diff
@@ -0,0 +1,363 @@
...
+        READ_CHARACTERISITC_UUID,
+        Characteristic.Properties.READ,
+        Characteristic.READABLE,
+        b"DATA",
+    )
+    virtual_service = Service(
+        READ_SERVICE_UUID,
+        [virtual_characteristic],
+    )
+    bumble_peripheral.add_services([virtual_service])
```
dlech:
```suggestion
    bumble_peripheral.add_service(virtual_service)
```

tests/integration/test_client.py
```diff
@@ -0,0 +1,363 @@
...
+    await bumble_peripheral.start_advertising()
+
+    device = await BleakScanner.find_device_by_name(bumble_peripheral.name)
+    assert device
+
+    async with BleakClient(device) as client:
+        await client.write_gatt_char(
+            WRITE_CHARACTERISITC_UUID,
+            b"DATA",
+        )
```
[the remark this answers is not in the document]
dlech: Also should use explicit `response=True` as recommended in Bleak docs.

tests/integration/test_client.py
```diff
@@ -0,0 +1,363 @@
...
+async def test_pairing(bumble_peripheral: Device):
+    """Check if pairing is possible."""
+    # TODO
+
+
+@pytest.mark.skipif(
+    get_default_backend() != BleakBackend.CORE_BLUETOOTH,
+    reason="pairing is not possible on CoreBluetooth",
+)
+async def test_pairing_unavailabe(bumble_peripheral: Device):
```
dlech: spelling: unavailable

dlech (commented):
Looks like a good start.

I wouldn't mind splitting this up into a few files already (connecting, characteristics, descriptors, pairing, properties) as I expect we will eventually add many more tests to cover error paths. Doing that in a single file would make the file too long.

dlech:
Profiling shows that it is spending all of it's time on `epoll()` in the asyncio run loop. I haven't been able to figure out how to show which tasks/coroutines are waiting at this time though.

dlech:
The problem seems to be in the Linux kernel. Running `btmon` shows:

```
> ACL Data RX: Handle 0 flags 0x02 dlen 5                                                                           #178 [hci0] 8.012991
      ATT: Write Response (0x13) len 0
bluetoothd[3701]: @ MGMT Command: Disconnect (0x0014) plen 7                                                    {0x0001} [hci0] 9.918092
        LE Address: DA:96:79:7A:C8:B2 (Static)
< HCI Command: LE Create Connection Cancel (0x08|0x000e) plen 0                                                    #179 [hci0] 28.448065
```

We can see the expected 2 second delay from `bluetoothd` between the last operation (Write Response) and and the MGMT Disconnect command.

However, there is an 18 second delay between the MGMT command to the kernel and actually sending the HCI command.

dlech:
OK, the disconnect is actually coming from a 20 second timeout in `hci_le_create_conn_sync()` in the Linux kernel.

I think the kernel is expecting `HCI_EV_LE_ENHANCED_CONN_COMPLETE` instead of `HCI_EV_LE_CONN_COMPLETE` because either `HCI_LE_LL_PRIVACY` or `HCI_LE_EXT_ADV` is supported on the VHCI device. So we should be able to tweak bumble to give the expected response.

tests/integration/test_client.py
```diff
@@ -0,0 +1,363 @@
...
+
+async def test_read_gatt_char(bumble_peripheral: Device):
+    """Reading a GATT characteristic is possible."""
+    add_default_advertising_data(bumble_peripheral)
+    READ_SERVICE_UUID = "2908f536-3fab-43c9-a7b2-80b6fdaae99b"
+    READ_CHARACTERISITC_UUID = "1a1af049-9c23-4b69-b763-e1096674ed18"
+    virtual_characteristic = Characteristic(
+        READ_CHARACTERISITC_UUID,
+        Characteristic.Properties.READ,
+        Characteristic.READABLE,
```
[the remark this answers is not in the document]
timrid: I think `Characteristic.Permission.READABLE` is more readable than `Attribute.Permission.READABLE`

tests/integration/test_client_properties.py
```diff
@@ -0,0 +1,36 @@
...
+async def test_mtu_size(bumble_peripheral: Device):
+    """Check if the mtu size can be optained."""
+    await configure_and_power_on_bumble_peripheral(bumble_peripheral)
+
+    device = await find_ble_device(bumble_peripheral)
+
+    async with BleakClient(device) as client:
+        # The mtu_size is different between different platforms. So it is not possible
+        # to get a reliable value. So we only check that it is an int.
+        assert isinstance(client.mtu_size, int)
```
dlech:
```suggestion
        assert 23 <= client.mtu_size <= 517
```

This would test that it is a reasonable number.

tests/integration/test_client_properties.py
```diff
@@ -0,0 +1,36 @@
...
+    reason="'get_rssi()' is only available on CoreBluetooth",
+)
+async def test_get_rssi(bumble_peripheral: Device):
+    """Getting RSSI from client is possible."""
+    await configure_and_power_on_bumble_peripheral(bumble_peripheral)
+
+    device = await find_ble_device(bumble_peripheral)
+
+    async with BleakClient(device) as client:
+        rssi = await client._backend.get_rssi()  # type: ignore
```
dlech:
```suggestion
        rssi = await client._backend.get_rssi()  # pyright: ignore[reportPrivateUsage]
```

Makes it more obvious what exactly we are ignoring.

tests/integration/test_client_properties.py
dlech: This raises `UserWarning` on Linux. The test should verify that.

I also created https://github.com/hbldh/bleak/pull/1898 to that CI will fail if we miss handling any warnings like this.

tests/integration/test_client_descriptors.py
```diff
@@ -0,0 +1,89 @@
...
+        )
+        assert characteristic
+
+        descriptor = characteristic.get_descriptor(
+            GATT_CHARACTERISTIC_USER_DESCRIPTION_DESCRIPTOR.to_hex_str()
+        )
+        assert descriptor
+
+        await client.write_gatt_descriptor(descriptor, b"Description")
+        assert virtual_descriptor.value == b"Description"  # type: ignore
```
dlech: Can we be more specific about the type ignore? e.g. this seems like something that should be fixed in Bumble, so we should have a comment to that effect.

PR #1901, opened 2025-12-30
CoreBluetooth: refactor catching closed event loop errors in delegates

timrid:
This is a small refactoring of the CoreBluetooth implementation of catching runtime errors in the delegates. It removes duplicated code.

This will also simplify the integration tests to get a 100% code coverage of the CoreBluetooth implementation.

### DIFF
--- a/bleak/_compat.py
+++ b/bleak/_compat.py
@@ -10,12 +10,14 @@ if sys.version_info < (3, 11):
     from async_timeout import timeout as timeout
     from typing_extensions import Never as Never
     from typing_extensions import Self as Self
+    from typing_extensions import TypeVarTuple as TypeVarTuple
     from typing_extensions import Unpack as Unpack
     from typing_extensions import assert_never as assert_never
 else:
     from asyncio import timeout as timeout  # noqa: F401
     from typing import Never as Never  # noqa: F401
     from typing import Self as Self  # noqa: F401
+    from typing import TypeVarTuple as TypeVarTuple  # noqa: F401
     from typing import Unpack as Unpack  # noqa: F401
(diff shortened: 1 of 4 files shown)


bleak/backends/corebluetooth/utils.py
```diff
@@ -68,3 +74,19 @@ def to_optional_int(value: Optional[NSNumber]) -> Optional[int]:
         return None
 
     return int(value)
+
+
+_Ts = TypeVarTuple("_Ts")
+
+
+def call_soon_threadsafe(
```
dlech: This could be useful in the WinRT backend as well since it has a similar situation with event callbacks.

So perhaps we should start a new `bleak.backends._util` module for this?

bleak/backends/corebluetooth/utils.py
```diff
@@ -68,3 +74,19 @@ def to_optional_int(value: Optional[NSNumber]) -> Optional[int]:
...
+    event_loop: asyncio.AbstractEventLoop,
+    callback: Callable[[Unpack[_Ts]], object],
```
dlech:
```suggestion
    callback: Callable[[Unpack[_Ts]], Any],
```

bleak/backends/corebluetooth/utils.py
```diff
@@ -1,5 +1,9 @@
+import asyncio
+import logging
 import sys
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Callable
```
dlech:
```suggestion
from collections.abc import Callable
from typing import TYPE_CHECKING
```

timrid:
I have addressed all points. The PR is ready for a new review.

PR #1902, opened 2025-12-30
CoreBluetooth: Track code coverage of objc delegates in integration tests

timrid:
Currently the code coverage of the pyobjc delegates is not tracked, because they are executed in threads that are not created by python. This adds a patch, so that the coverage is now tracked.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -110,14 +110,13 @@ _TReturn = TypeVar("_TReturn")
 
 def enable_coverage(func: Callable[_P, _TReturn]) -> Callable[_P, _TReturn]:
     """
-    Enable coverage tracing on a non-Python-created thread.
+    Enable coverage tracing on a non-Python-created thread, if not already enabled.
     (https://github.com/nedbat/coveragepy/issues/686)
     """
 
     @functools.wraps(func)
     def wrapped(*args: _P.args, **kwargs: _P.kwargs) -> _TReturn:
-        trace_hook = threading.gettrace()
-        if trace_hook:
+        if sys.gettrace() is None and (trace_hook := threading.gettrace()):
             sys.settrace(trace_hook)
         return func(*args, **kwargs)
 


tests/integration/conftest.py
```diff
@@ -98,3 +100,72 @@ async def find_ble_device(bumble_peripheral: Device) -> BLEDevice:
         raise RuntimeError("failed to discover device, is Bumble working?")
 
     return device
+
+
+def enable_coverage(fn: Callable[..., Any]) -> Callable[..., Any]:
+    # Enable coverage tracing on this non-Python-created thread
+    # (https://github.com/nedbat/coveragepy/issues/686).
+
+    def wrapped(*args: Any, **kwargs: Any) -> Any:
```
dlech: Should we be using `functools.wraps`?

dlech (commented):
Interesting. I thought there would have been some kind of magic hook already in the Python runtime for this. I guess we will want to do something similar for WinRT event handlers as well.

tests/integration/conftest.py
```diff
@@ -98,3 +100,72 @@ async def find_ble_device(bumble_peripheral: Device) -> BLEDevice:
...
+    if sys.platform != "darwin":
+        # Patching CoreBluetooth is only necessary on macOS
+        return
+
+    from bleak.backends.corebluetooth.CentralManagerDelegate import (
+        ObjcCentralManagerDelegate,
+    )
+    from bleak.backends.corebluetooth.PeripheralDelegate import ObjcPeripheralDelegate
+
+    delegates_to_patch: list[tuple[type[Any], list[Callable[..., Any]]]] = [
```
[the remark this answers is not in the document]
timrid: Alternatively, we can add an `@objc_method` decorator and decorate all of these functions with it. Then we only have to adapt this `@objc_method` decorator in the unit tests. Another advantage is that such a decorator is required in `rubicon-objc` anyway (see, for example, here https://rubicon-objc.beeware.org/en/stable/tutorial/tutorial-2/).

I have update it like this.

tests/integration/conftest.py
```diff
@@ -98,3 +101,36 @@ async def find_ble_device(bumble_peripheral: Device) -> BLEDevice:
...
+    by default. This patch wraps the delegate methods to enable coverage tracing
+    in those threads.
+    """
+    if sys.platform != "darwin":
+        # Patching CoreBluetooth is only necessary on macOS
+        return
+
+    from bleak.backends.corebluetooth import utils
+
+    utils.objc_method = lambda func: enable_coverage(func)
```
dlech: Also, what if we have an objc method that isn't a delegate callback? Does it hurt to have the extra coverage hook on it?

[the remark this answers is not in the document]
timrid: Monkeypatching would mean that this would be reversed after the fixture, but then all relevant modules have already been imported, so reversing it would be pointless.

I think that the whole fixture is unnecessary here and could actually be problematic. That's why I now simply overwrite the decorator in conftest.py. pytest ensures that conftest.py is always executed before the `test_###.py` files are imported. This ensures that the decorator is correct as long as we do not import the CoreBluetooth backend modules in conftest.py itself.

dlech: > It is always called from the main thread and not from the dispatch queue

Interesting... I wonder how it does this without a `NSRunLoop` on the main thread.

How did you determine it is called on the main thread? What is the call stack?

bleak/_compat.py
```diff
@@ -25,3 +25,8 @@
     from typing_extensions import override as override
 else:
     from typing import override as override  # noqa: F401
+
+if sys.version_info < (3, 13):
+    from typing_extensions import ParamSpec as ParamSpec
```
dlech: Why? `ParamSpec` was added in Python 3.10.

tests/integration/conftest.py
dlech: The line number isn't shown for `start_scan()`. I assume it is line 303 where `scanForPeripheralsWithServices_options_()` is called.

I wonder if this always happens or there is a case where it is deferred because scanning can't start right away for some reason.

timrid: Your assumption is right.

I added a log message with the thread-id and ran all integration tests. It is always the same thread id. But just because that's the case in the integration tests, it's no guarantee that it will always be the case. So I can also add the `@external_thread_callback` decorator here. It definitely doesn't cause any problems if adding it without requiring it. Or what do you think?

dlech: I tried that and we do get the callback on a dummy thread, so I guess it does use the dispatch queue even for this. So yes, it looks like we need the decorator there.

Is the threading hook the same as the main thread hook? Asking since this can be called on the main thread, it can write over the existing trace hook.

timrid:
I have updated this PR, please review again.

PR #1904, opened 2026-01-01
tests/CoreBluetooth: add integration tests for `use_cached` parameter

timrid:
This adds integration tests for the `use_cached` parameter in CoreBluetooth when reading characteristics and descriptors

### DIFF
--- a/tests/integration/test_client_characteristics.py
+++ b/tests/integration/test_client_characteristics.py
@@ -34,6 +34,33 @@ async def test_read_gatt_char(bumble_peripheral: Device):
         assert data == b"DATA"
 
 
+async def test_read_gatt_char_use_cached(bumble_peripheral: Device):
+    """Reading a cached GATT characteristic is possible."""
+    READ_SERVICE_UUID = "2908f536-3fab-43c9-a7b2-80b6fdaae99b"
+    READ_CHARACTERISITC_UUID = "1a1af049-9c23-4b69-b763-e1096674ed18"
+    virtual_characteristic = Characteristic(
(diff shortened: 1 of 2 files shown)


tests/integration/test_client_characteristics.py
```diff
@@ -34,6 +35,36 @@ async def test_read_gatt_char(bumble_peripheral: Device):
         assert data == b"DATA"
 
 
+@pytest.mark.skipif(
+    sys.platform != "darwin", reason="Test only applies to CoreBluetooth"
```
dlech: This is not specific to macOS, it should be enabled on all platforms.

tests/integration/test_client_descriptors.py
```diff
@@ -53,6 +56,55 @@ async def test_read_gatt_descriptor(bumble_peripheral: Device):
         assert data == b"Description"
 
 
+@pytest.mark.skipif(
+    sys.platform != "darwin", reason="Test only applies to CoreBluetooth"
```
dlech: ditto

tests/integration/test_client_characteristics.py
dlech: done

dlech: I tested Windows, so if you want to re-test Mac, I'll merge this.

timrid: Wow, that was fast 👍 

I also verified the integration test on windows, without problems.

timrid: Also macOS is no problem 👍

PR #1905, opened 2026-01-01
tests/CoreBluetooth: Test that scanner handles closed event loop gracefully

timrid:
This adds an integration test, to test an stopped event loop while scanning is active.

### DIFF
--- a/tests/integration/test_scanner.py
+++ b/tests/integration/test_scanner.py
@@ -1,6 +1,9 @@
 import asyncio
 import contextlib
+import logging
 import sys
+import threading
+import time
 
 import pytest
 from bumble import data_types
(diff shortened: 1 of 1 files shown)


timrid:
This is the only way I found to cover this [line of code](https://github.com/hbldh/bleak/blob/b6a90a2cee2736d58faedb79fde75e674ab55e1c/bleak/backends/_utils.py#L24) in a complete bleak setup. These are the alternatives to cover this line in a reduced setup. Any preferences?

- reduce the test to only test `try_call_soon_threadsafe` but also create a event loop and stop it.
- reduce the test to only test `try_call_soon_threadsafe` but mock the `call_soon_threadsafe` to raise a RuntimeError
- add a `# pragma: no cover` to this line.

timrid:
All right. Then I will close this PR and do that in a separate PR, as there are more places where a `no cover` needs to be used.

PR #1906, opened 2026-01-01
tests/CoreBluetooth: Add an test for the bluetooth availability check

timrid:
This adds a test for the bluetooth availability check.

Unfortunately, it is not possible to perform these as real integration tests, as there are no options to trigger these settings programmatically in some cases. Therefore, the values are only mocked in these tests.

### DIFF
--- a/tests/integration/test_cb_bluetooth_availability.py
+++ b/tests/integration/test_cb_bluetooth_availability.py
@@ -21,10 +21,7 @@ from CoreBluetooth import (
 
 from bleak import BleakScanner
 from bleak.backends.corebluetooth.scanner import BleakScannerCoreBluetooth
-from bleak.exc import (
-    BleakBluetoothNotAvailableError,
-    BleakBluetoothNotAvailableReason,
-)
+from bleak.exc import BleakBluetoothNotAvailableError, BleakBluetoothNotAvailableReason
 
 
 @pytest.mark.parametrize(


tests/integration/test_cb_bluetooth_availability.py
```diff
@@ -0,0 +1,120 @@
+import sys
+
+import pytest
+
+if sys.platform != "darwin":
+    pytest.skip("CoreBluetooth only tests", allow_module_level=True)
```
dlech:
```suggestion
    pytest.skip("CoreBluetooth only tests", allow_module_level=True)
    # unreachable, but makes the type checkers happy
    assert False
```

For some reason, the type checkers don't see that `skip()` is NoReturn.

tests/integration/test_cb_bluetooth_availability.py
```diff
@@ -0,0 +1,120 @@
...
+            CBManagerStateResetting,
+            UNCHANGED,
+            "Connection to the Bluetooth system service was lost",
+            BleakBluetoothNotAvailableReason.UNKNOWN,
+            id="resetting",
+        ),
+        pytest.param(
+            CBManagerStateUnknown,
+            UNCHANGED,
+            "state is unknown",
```
dlech: looks like we have a spelling error in the actual error message

tests/integration/test_cb_bluetooth_availability.py
```diff
@@ -0,0 +1,120 @@
...
+
+UNCHANGED = object()
+
+
+@pytest.mark.parametrize(
+    "state,authorization,expected_msg,expected_reason",
+    [
+        pytest.param(
+            CBManagerStateUnsupported,
+            UNCHANGED,
```
dlech: How about just `None` instead of introducing `UNCHANGED`?

tests/integration/test_cb_bluetooth_availability.py
```diff
@@ -0,0 +1,120 @@
...
+    state: int,
+    authorization: int | object,
+    expected_msg: str,
+    expected_reason: BleakBluetoothNotAvailableReason,
+):
+    """An Exception is raised, when bluetooth is not available."""
+
+    scanner = BleakScanner()
+
+    # Unfortunately it is not possible to modify the bluetooth state on a macOS pro
```
dlech: I was wondering if we could use AppleScript or something like that to change the settings, but just mocking is probably more reliable.

timrid: If that's even possible, it would be much more complicated. That can still be done, but for now, these tests are definitely better than none at all.

tests/integration/test_cb_bluetooth_availability.py
timrid: Done

tests/integration/test_cb_bluetooth_availability.py
timrid: I corrected the spelling error.

PR #1907, opened 2026-01-01
tests/CoreBluetooth: Add a tests for discovery by address using `use_bdaddr`

timrid:
Add a integration tests on CoreBluetooth for discovery by address using `use_bdaddr`.

### DIFF
--- a/tests/integration/test_scanner.py
+++ b/tests/integration/test_scanner.py
@@ -35,6 +35,16 @@ async def test_discover(bumble_peripheral: Device):
         assert filtered_devices[0][0].address == str(bumble_peripheral.static_address)
 
 
+async def test_find_by_address(bumble_peripheral: Device):
+    """Scanner is finding the device by address."""
+    await configure_and_power_on_bumble_peripheral(bumble_peripheral)
+
+    device = await BleakScanner.find_device_by_address(
+        bumble_peripheral.static_address.to_string(), cb={"use_bdaddr": True}
+    )
+    assert device is not None
+
+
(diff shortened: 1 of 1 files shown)


tests/integration/test_scanner.py
```diff
@@ -29,10 +28,21 @@ async def test_discover(bumble_peripheral: Device):
     )
 
     assert len(filtered_devices) == 1
-    assert filtered_devices[0][1].local_name == bumble_peripheral.name
-    if sys.platform != "darwin":
-        # accessing the address is not possible on macOS
-        assert filtered_devices[0][0].address == str(bumble_peripheral.static_address)
+
+
+async def test_discover_by_address(bumble_peripheral: Device):
```
dlech: We already have a test that calls `find_device_by_address()` that should get coverage for this.

Since these tests are so slow, I would rather not add tests for every little thing.

timrid: I am more in favor of explicitly testing individual aspects of the code rather than implicitly testing other parts of the code in a test for which the test is not intended. That's why I created https://github.com/hbldh/bleak/pull/1908 to remove `find_device_by_address` from the other test.

If the test times are a real problem, it might be better to use [subtests](https://docs.pytest.org/en/stable/how-to/subtests.html) (that is included in pytest since v9.0) to explicitly test different things in one test.

dlech: I would still prefer a test for `BleakScanner.find_device_by_address()` over this.

PR #1909, opened 2026-01-02
test/integration: Read binary descriptor

timrid:
Add a integration test to verify that reading binary descriptors is possible.

CoreBluetooth distinguishes between binary and string descriptors. One returns an NSData and the other returns an NSString.

### DIFF
--- a/tests/integration/test_client_descriptors.py
+++ b/tests/integration/test_client_descriptors.py
@@ -3,13 +3,10 @@ from collections.abc import AsyncGenerator
 
 import pytest
 import pytest_asyncio
+from bumble import gatt
+from bumble.core import UUID
 from bumble.device import Device
-from bumble.gatt import (
-    GATT_CHARACTERISTIC_USER_DESCRIPTION_DESCRIPTOR,
-    Characteristic,
-    Descriptor,
-    Service,
-)
+from bumble.gatt import Characteristic, Descriptor, Service
 from bumble.transport.common import Transport
 
 from bleak import BleakClient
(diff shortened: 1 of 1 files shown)


tests/integration/test_client_descriptors.py
```diff
@@ -100,14 +109,15 @@ async def descr_test_peripheral(
...
+            readable_string_descr=readable_string_descr,
+            readable_binary_descr=readable_binary_descr,
             writable_descr=writable_descr,
         )
 
 
 @pytest.mark.asyncio(loop_scope="module")
-async def test_read_gatt_descriptor(descr_test_peripheral: DescrTestPeripheral):
-    """Reading a GATT descriptor is possible."""
+async def test_read_gatt_string_descriptor(descr_test_peripheral: DescrTestPeripheral):
```
dlech: While we are here, this test could be made more robust by including a multi-byte UTF-8 character to make sure it survives being changed to a NSString and back.

tests/integration/test_client_descriptors.py
```diff
@@ -12,15 +14,34 @@
...
+@pytest.mark.parametrize(
+    "descriptor_uuid,descriptor_value",
+    [
+        pytest.param(
+            GATT_CHARACTERISTIC_USER_DESCRIPTION_DESCRIPTOR,
+            b"User Description",
+            id="string_descriptor",
+        ),
+        pytest.param(
+            CUSTOM_BINARY_DESCRIPTOR_UUID,
```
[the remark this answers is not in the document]
timrid: You are right, some of them are returned as NSNumber (e.g. https://developer.apple.com/documentation/corebluetooth/cbuuidcharacteristicextendedpropertiesstring). I fixed it in a separat commit in this PR.

tests/integration/test_client_descriptors.py
```diff
@@ -41,35 +38,117 @@ async def bumble_peripheral(hci_transport: Transport) -> Device:
...
 
 
+STANDARD_DESCRIPTORS: list[tuple[UUID, bytes]] = [
+    (
+        gatt.GATT_CHARACTERISTIC_EXTENDED_PROPERTIES_DESCRIPTOR,
+        b"\x03\x00",  # Reliable Write + Writable Auxiliaries
+    ),
+    (
+        gatt.GATT_CHARACTERISTIC_USER_DESCRIPTION_DESCRIPTOR,
+        "🚀 Description".encode("utf-8"),  # Test Description
```
dlech:
```suggestion
        "🚀 Description".encode(),  # Test Description
```

"utf-8" is the default value

tests/integration/test_client_descriptors.py
```diff
@@ -100,28 +178,34 @@ async def descr_test_peripheral(
...
     assert characteristic
 
-    descriptor = characteristic.get_descriptor(
-        GATT_CHARACTERISTIC_USER_DESCRIPTION_DESCRIPTOR.to_hex_str()
-    )
+    descriptor = characteristic.get_descriptor(descr_uuid.to_hex_str())
     assert descriptor
 
-    descr_test_peripheral.readable_descr.value = b"Description"
+    descr_test_peripheral.readable_descr[descr_uuid].value = descr_data
```
dlech: A comment here to remind us that other tests may have modified the value would be helpful.

bleak/backends/corebluetooth/client.py
```diff
@@ -317,12 +321,19 @@ async def read_gatt_descriptor(
...
         )
-        if isinstance(
-            output, str
-        ):  # Sometimes a `pyobjc_unicode`or `__NSCFString` is returned and they can be used as regular Python strings.
+
+        # Most descriptors are returned as NSData (raw bytes), but some of them
+        # are returned as NSString or NSNumber.
+        # See: https://developer.apple.com/documentation/corebluetooth/characteristic-descriptors
+        if descriptor.uuid in DESCRIPTOR_TYPE_NSNUMBER:
+            output = cast(NSNumber, output)
```
dlech: We could avoid the casts by making type discriminator functions to use in the if statements.

bleak/backends/corebluetooth/utils.py
```diff
@@ -68,3 +74,30 @@ def to_optional_int(value: Optional[NSNumber]) -> Optional[int]:
...
+    if descriptor_uuid in _DESCRIPTOR_TYPE_NSNUMBER:
+        return True
+    return False
+
+
+def is_descriptor_nsstring(value: Any, descriptor_uuid: str) -> TypeGuard[NSString]:
+    """Check if descriptor value is returned as NSString by CoreBluetooth."""
+    if descriptor_uuid in _DESCRIPTOR_TYPE_NSSTRING:
+        return True
+    return False
```
dlech:
```suggestion
    return descriptor_uuid in _DESCRIPTOR_TYPE_NSSTRING
```

timrid:
From a code coverage perspective, this is not necessary, but it could make sense in order to ensure functionality.

dlech:
Yes, I expect some descriptors, like CCCD, might raise an error since it should be managed by notifications. So it will be interesting to see which OSes allow it and which don't.

PR #1910, opened 2026-01-02
tests/integration: Speedup characteristic and descriptor tests by a module level peripheral

timrid:
The integration tests are slow, because every time a new peripheral is created and connected to. But for some tests like characteristic and descriptor tests it is not necessary to always create a new peripheral. So to speed things up a module level peripheral fixture is created to share a connected BleakClient/BumblePeripheral between the tests.

### DIFF
--- a/tests/integration/test_client_characteristics.py
+++ b/tests/integration/test_client_characteristics.py
@@ -1,182 +1,219 @@
 import asyncio
+import dataclasses
+from collections.abc import AsyncGenerator
 
 import pytest
+import pytest_asyncio
 from bumble.device import Connection, Device
 from bumble.gatt import Characteristic, CharacteristicValue, Service
+from bumble.transport.common import Transport
 
 from bleak import BleakClient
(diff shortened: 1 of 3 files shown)


tests/integration/test_client_characteristics.py
```diff
@@ -1,182 +1,214 @@
...
+            notify_characteristic=notify_characteristic,
         )
-        assert virtual_characteristic.value == b"DATA"
 
 
-async def test_write_gatt_char_no_response(bumble_peripheral: Device):
-    """Writing a GATT characteristic is possible."""
-    WRITE_WITHOUT_RESPONSE_SERVICE_UUID = "79a92bad-31b4-4a70-885c-e704ae2c6363"
-    WRITE_WITHOUT_RESPONSE_CHARACTERISITC_UUID = "2a07f6ea-6401-45d7-838c-0f459a7edb7f"
+@pytest.mark.asyncio(loop_scope="module")
```
dlech: Can we use `pytestmark` to avoid repeating this on every test?

PR #1912, opened 2026-01-03
tests/CoreBluetooth: Add a test, that verifies that the `isScanning` observer is working correctly.

timrid:
Normally in the integration tests, the `CBCentralManager.isScanning()` flag always is set correctly after calling `scanForPeripheralsWithServices_options_` or `stopScan`. But there may be cases, where the `isScanning` flag is not updated immediately and we have to wait until the `isScanning` observer is called. This test ensures that the observer is called correctly in such cases.

Maybe more of such backend specific test quirks are required, so I also renamed the test file.

### DIFF
--- a/tests/integration/test_quirks_corebluetooth.py
+++ b/tests/integration/test_quirks_corebluetooth.py
@@ -20,8 +20,22 @@ from CoreBluetooth import (
 )
 
 from bleak import BleakScanner
+from bleak.backends.corebluetooth.CentralManagerDelegate import CentralManagerDelegate
 from bleak.backends.corebluetooth.scanner import BleakScannerCoreBluetooth
 from bleak.exc import BleakBluetoothNotAvailableError, BleakBluetoothNotAvailableReason
+
+
+def get_central_manager_delegate(
+    scanner: BleakScanner,
+) -> CentralManagerDelegate:
+    """Get the private CentralManagerDelegate Object from the scanner."""
+    backend = scanner._backend  # pyright: ignore[reportPrivateUsage]
+    assert isinstance(
+        backend,
+        BleakScannerCoreBluetooth,
+    )
+    central_manager_delegate = backend._manager  # pyright: ignore[reportPrivateUsage]
+    return central_manager_delegate
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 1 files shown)


tests/integration/test_quirks_corebluetooth.py
```diff
@@ -115,3 +123,31 @@ async def test_bluetooth_availability(
...
     assert exc_info.value.reason == expected_reason
+
+
+async def test_isScanning_observer(
+    monkeypatch: pytest.MonkeyPatch,
+):
+    """
+    Normally in the integration tests, the 'CBCentralManager.isScanning()' flag always
+    is set correctly after calling 'scanForPeripheralsWithServices_options_' or 'stopScan'.
+    But there may be cases, where the 'isScanning' flag is not updated immediately and we have
```
dlech: Do we know this is actually true? 

It could very well be that calling `scanForPeripheralsWithServices_options_()` always sets `isScanning` to true during the call. The fact that we observed the observer callback on the main thread make me suspect that this is the case.

tests/integration/test_quirks_corebluetooth.py
```diff
@@ -115,3 +123,31 @@ async def test_bluetooth_availability(
...
+    central_manager_delegate = get_central_manager_delegate(scanner)
+    mock_manager = Mock(wraps=central_manager_delegate.central_manager)
+    monkeypatch.setattr(
+        central_manager_delegate,
+        "central_manager",
+        mock_manager,
+    )
+
+    # Simulate that isScanning is alway False, so that the future from the observer is awaited
+    mock_manager.isScanning.return_value = False
```
dlech: If this can actually happen in real life, it would be better to actually reproduce those conditions rather than mocking something that may not even be possible.

tests/integration/test_quirks_corebluetooth.py
timrid: Nop, I am not sure. I was just expecting it, because it is implemented like this.

Any idea how we can verify this for sure? And if we are sure we should change the actual bleak implementation to not wait on these futures?

timrid:
The code that should be tested was completely removed in https://github.com/hbldh/bleak/pull/1913, so this is closed.

PR #1916, opened 2026-01-06
tests/CoreBluetooth: Add a test, that verifies that the `didFailToConnectPeripheral` delegate is working correctly.

timrid:
This PR adds a test, that verifies that the `didFailToConnectPeripheral` delegate is working correctly.

Unfortunately, I haven't been able to find a way to actually reproduce this error using Bumble with the nRF dongle. 

From the [Apple docs](https://developer.apple.com/documentation/corebluetooth/cbcentralmanagerdelegate/centralmanager(_:didfailtoconnect:error:)?language=objc):
> Because connection attempts don’t time out, a failed connection usually indicates a transient issue, in which case you may attempt connecting to the peripheral again.

I don't think that it's possible to create such an "transient issue" deterministically with a standard HCI like in the nRF Firmware.

### DIFF
--- a/tests/integration/test_quirks_corebluetooth.py
+++ b/tests/integration/test_quirks_corebluetooth.py
@@ -7,8 +7,10 @@ if sys.platform != "darwin":
     # unreachable, but makes the type checkers happy
     assert False
 
+from typing import Any, Optional, cast
 from unittest.mock import Mock
 
+from bumble.device import Device
 from CoreBluetooth import (
     CBManagerAuthorizationDenied,
     CBManagerAuthorizationRestricted,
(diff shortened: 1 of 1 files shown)


dlech:
> I don't think that it's possible to create such an "transient issue" deterministically with a standard HCI like in the nRF Firmware.

Can we perhaps call `HCI_LE_Add_Device_To_Filter_Accept_List_Command` on the peripheral with a different address to cause the central to fail to connect?

timrid:
> Can we perhaps call `HCI_LE_Add_Device_To_Filter_Accept_List_Command` on the peripheral with a different address to cause the central to fail to connect?

Unfortunately, this only results in a `TimeoutError` in my tests and we already have a test to cover a `TimeoutError`.

miguel-neto-andrade (commented):
I ran the test's approach against real hardware on macOS (Apple Silicon, BlueNRG-based peripheral), since the bumble fixture can't produce this error deterministically: found the device with a real scan, patched `central_manager` with the `Mock(wraps=...)` + `side_effect` exactly as in this PR, and attempted a real `BleakClient` connect. The simulated `didFailToConnectPeripheral` path works as intended — it raises `BleakError: failed to connect: <Mock id=...>` — and a real connect to the same peripheral afterwards still succeeds, so the mock doesn't corrupt the delegate state. The mechanics are also sound: the connect future is registered before `connectPeripheral_options_` is called, so invoking the ObjC delegate method synchronously from the side effect is race-free.

A few observations from that run:

1. `pytest.raises(Exception, match="failed to connect")` could be tightened to `BleakError` — that's what `did_fail_to_connect_peripheral` raises, and `Exception` would also match e.g. an `AssertionError` containing the same text.

2. The `localizedDescription="Simulated connection error"` kwarg on the error mock is never used — `did_fail_to_connect_peripheral` formats the error with `f"failed to connect: {error}"`, so the message ends up containing the Mock's repr rather than the simulated description. Passing a real `NSError` (`NSError.errorWithDomain_code_userInfo_(...)`) would make the simulated error flow through `str()` the same way a real CoreBluetooth error does.

3. Possibly out of scope, but the test surfaces what looks like a small pre-existing leak: `CentralManagerDelegate.connect()` only cleans up `self._disconnect_callbacks[peripheral.identifier()]` on the `TimeoutError` path, so after a `didFailToConnectPeripheral` failure a stale entry remains (I confirmed this on hardware: 1 entry left after the simulated failure). A retry overwrites it, so impact is low, but the test could assert clean state and/or `connect()` could clean up in a `finally`.

Issue #1773, opened 2025-05-29
Add specialized exception for GATT Protocol Error


dlech:
### Discussed in https://github.com/hbldh/bleak/discussions/1772

<div type='discussions-op-text'>

<sup>Originally posted by **PimWagemans09** May 29, 2025</sup>
currently if i get an error like 
bleak.exc.BleakError: Could not write value b'\x02' to characteristic 0011: Protocol Error 0x81: Unknown
i just catch the error and print it but i want to get the error code "0x81" and process it 
is there any way to do this?</div>

dlech:
> was just curious as to how I can recreate this

Do you mean how to force hardware to generate and error? Just run the service explorer example and look for a characteristic that doesn't have the `read` property and try to read it or something like that.


> is the formatting for these types of error roughly the same on all platforms

No, I would not expect so. On Windows, there is actually somewhat structured errors. If a GATT request fails, in the `*Result`, there is a [Status ](https://learn.microsoft.com/en-us/uwp/api/windows.devices.bluetooth.genericattributeprofile.gattcommunicationstatus?view=winrt-26100) and a [ProtocolError ](https://learn.microsoft.com/en-us/uwp/api/windows.devices.bluetooth.genericattributeprofile.gattprotocolerror?view=winrt-26100)(only if Status is ProtocolError).

So it would be nice to model a custom exception after that. I.e. the exception type should have properties to get the actual error numbers as integers so that users don't have to scrape it from a string.

On Linux/BlueZ, I don't think we have any choice but to scrape the string from the D-Bus error's extra data. On Mac, I don't know what sort of NSError we get and it isn't documented AFAICT, so we will probably just have to find a way to cause and error and see what happens.

The ProtocolError number should be the same cross-platform though because those numbers are defined by the Bluetooth spec. So we can add an enum for those.

mashurr:
Macos:

ERROR: Failed to read characteristic 2049: Error Domain=CBATTErrorDomain Code=2 "Reading is not permitted." UserInfo={NSLocalizedDescription=Reading is not permitted.}
ERROR: Failed to write characteristic 2049: Error Domain=CBATTErrorDomain Code=3 "Writing is not permitted." UserInfo={NSLocalizedDescription=Writing is not permitted.}

PR #1917, opened 2026-01-06
test/integration: Write all standard descriptors

timrid:
This PR adds a write test for all standard descriptors. Only the CCCD is not writeable on CoreBluetooth and BlueZ. On Windows it is writeable.

### DIFF
--- a/tests/integration/test_client_descriptors.py
+++ b/tests/integration/test_client_descriptors.py
@@ -48,81 +48,99 @@ class DescrTestPeripheral:
     bleak_client: BleakClient
     bumble_peripheral: Device
     readable_descr: dict[UUID, Descriptor]
-    writable_descr: Descriptor
+    writable_descr: dict[UUID, Descriptor]
 
 
-STANDARD_DESCRIPTORS: list[tuple[UUID, bytes]] = [
+STANDARD_DESCRIPTORS: list[tuple[UUID, bytes, bytes]] = [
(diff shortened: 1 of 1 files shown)


tests/integration/test_client_descriptors.py
```diff
@@ -248,25 +279,54 @@ async def test_read_gatt_descriptor_use_cached(
...
+    descr_test_peripheral.writable_descr[descr_uuid].value = descr_data_1
+
+    if (
+        descr_uuid == gatt.GATT_CLIENT_CHARACTERISTIC_CONFIGURATION_DESCRIPTOR
+        and descr_test_peripheral.bleak_client.backend_id
+        in (BleakBackend.CORE_BLUETOOTH, BleakBackend.BLUEZ_DBUS)
+    ):
+        # On some backends the CCCD is not writable, because it is manage by
+        # notifications/indications APIs. So we check that writing fails.
+        with pytest.raises(Exception):
```
dlech: It would be nice if this raised a common exception with a helpful error message.

Or at least let's add a TODO here and make an issue for it if we don't have time to do it right now.

I guess it wouldn't be quite the same as https://github.com/hbldh/bleak/issues/1773 since this is coming from the OS rather than from GATT.

We could even raise an exception before even calling into the backend so that it is the same on all platforms even if the OS would permit it.

PR #1922, opened 2026-01-10
Add Tox configuration

timrid:
While working on Bleak, I noticed that pytest, mypy, flake8 or isort errors occurred very frequently in the CI, which I could actually have detected locally. But then I would have had to run various tools manually, which was always quite time-consuming.

So this PR adds a tox configuration. To run pytests on all supported python versions, lint, check types and building the docs just run: `tox`

But you can also selectively run separat steps:

- pytests on Python 3.10:
`tox -e py310`

- linting (isort, black, flake8):
`tox -e lint`

- linting (isort, black, flake8) without changing anything (mainly for CI usage):
`tox -e lint-checkonly`

- check types:
`tox -e typecheck`

- Building the docs:
`tox -e docs`

### DIFF
--- a/.github/alpine-vm/start-alpine-vm.sh
+++ b/.github/alpine-vm/start-alpine-vm.sh
@@ -21,7 +21,7 @@ VM_BUILD_DIR="$REPO_ROOT/.alpine-vm-build"
 VM_IMAGE="$VM_BUILD_DIR/alpine.qcow2"
 VM_SSH_PORT="2222"
 VM_CONSOLE_LOG="vm-console.log"
-VM_PACKAGES="bluez poetry openssh uv"
+VM_PACKAGES="bluez openssh uv"
 
 # Colors for output
 GREEN='\033[0;32m'
--- a/.github/alpine-vm/vm-ssh.sh
+++ b/.github/alpine-vm/vm-ssh.sh
@@ -9,4 +9,4 @@ ssh -o StrictHostKeyChecking=no \
     -o LogLevel=ERROR \
     -p 2222 \
     builder@localhost \
-    "cd ~/repo && $*"
+    "sh -l -c 'cd ~/repo && $*'"
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -18,28 +18,31 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
         python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
+    env:
+      FORCE_COLOR: "1"
     steps:
       - uses: actions/checkout@v4
-      - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v5
-        id: py
+
+      - name: Install uv and set up Python ${{ matrix.python-version }}
+        uses: astral-sh/setup-uv@v7
         with:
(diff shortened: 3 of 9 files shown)


timrid:
I completely forgot to describe my actual main motivation for introducing tox in the PR description:

When working on the rubicon-objc integration in https://github.com/hbldh/bleak/pull/1853, I regularly have to run integration tests with pyobjc *and* rubicon-objc. It was quite tedious to always do this manually, which is why I started using tox for the pytests (the integration tests can be run via `tox -- --bleak-hci-transport=serial:/dev/tty.usbmodem1101`). The fact that this also makes it very easy to test each individual Python version was an added bonus for me.

Then [you suggested](https://github.com/hbldh/bleak/pull/1853#issuecomment-3566964258) that the typings for pyobjc and rubicon-objc should be completely separated from each other. As a result, I always had to call mypy and pyright twice, once with the type hints for pyobjc and once for rubicon-objc. That's why I also automated this via tox as well.

Since the tox configuration has nothing to do with rubicon-objc and also offers advantages on its own, I thought it would be appropriate to handle this in a separate PR and not in https://github.com/hbldh/bleak/pull/1853. In the next few days, I will update https://github.com/hbldh/bleak/pull/1853 based on this PR with the additional tox configurations for rubicon-objc (and many more changes), so that perhaps what I mean will become a little clearer.

> I'm not sure tox is the best tool for this. It is pretty rare that I would ever need to run tests for multiple python versions locally.

Perhaps these cases are relatively rare, but in CI, only the unit tests are run for all Python versions. Running integration tests under all Python versions always involves additional effort by creating multiple environments manually, which can be greatly simplified with tox.

> It sounds like what you really looking for is a pre-commit hook that runs the auto-fixers and linters that can't autofix before committing.

Adding lint and docs was just an extra goodie to have a unified tool that does everything at once. Pre-commit hooks would of course also be suitable for linting (black, isort, flake8) and possibly also for type checks (mypy, pyright). However, I would not consider the execution of pytests to be a pre-commit.

dlech (commented):
I've been putting off reviewing this because I don't have a good reason to say that we don't need something like this, yet for some reason, I seem to have an aversion to `tox`. After thinking about it a bit, I think it is because it just doesn't seem very user-friendly. So I looked a bit at alternatives and https://poethepoet.natn.io/ looks like it could do the same job in a nicer way (well, at least in a way that I find easier to understand). Could we try that instead?

timrid:
> So I looked a bit at alternatives and https://poethepoet.natn.io/ looks like it could do the same job in a nicer way (well, at least in a way that I find easier to understand). Could we try that instead?

I've looked at poethepoet and prepared something locally. One advantage of tox is that you can easily run unit tests with different Python versions. This only works with poethepoet if you use it in combination with uv (see [here](https://poethepoet.natn.io/guides/tox_replacement_guide.html)). But this also means you would need to have poetry, poethepoet, and uv installed for it to work. This leads to requiring quite a few tools (poetry, poethepoet, uv, git, python) to be installed to work as a developer with this repository. This has led me to the question of whether we should generally switch from poetry to uv? This would eliminate the need to install poetry and python globally. What's your opinion on this? Would a general switch from poetry to uv (independent of the poethepoet or tox question) be an option?

timrid:
I followed your advice by using `poethepoet` and created https://github.com/hbldh/bleak/pull/1976 that replaces this PR.

PR #1925, opened 2026-01-13
backend/CoreBluetooth: Fix cyclic references problem causing memory leaks

timrid:
In https://github.com/hbldh/bleak/pull/1842, I accidentally introduced a memory leak because we have cyclic references between `ObjcCentralManagerDelegate` <-> `CentralManagerDelegate` and `ObjcPeripheralDelegate` <-> `PeripheralDelegate`. If these both were pure Python types, this should be resolved by the Python garbage collector, but since `ObjcCentralManagerDelegate` / `ObjcPeripheralDelegate` are ObjC objects, this does not seem to work.

### DIFF
--- a/tests/integration/test_quirks_corebluetooth.py
+++ b/tests/integration/test_quirks_corebluetooth.py
@@ -7,8 +7,11 @@ if sys.platform != "darwin":
     # unreachable, but makes the type checkers happy
     assert False
 
+import gc
+import weakref
 from unittest.mock import Mock
 
+from bumble.device import Device
 from CoreBluetooth import (
     CBManagerAuthorizationDenied,
     CBManagerAuthorizationRestricted,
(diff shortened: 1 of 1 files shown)


bleak/backends/corebluetooth/CentralManagerDelegate.py
```diff
@@ -76,7 +77,7 @@ class ObjcCentralManagerDelegate(NSObject, protocols=[CBCentralManagerDelegate])
     """
 
     def initWithPyDelegate_(
-        self, py_delegate: CentralManagerDelegate
+        self, py_delegate: weakref.ReferenceType["CentralManagerDelegate"]
```
dlech: This strikes me as a little odd that we force the caller to create the weak ref even though the caller doesn't need the weak ref for anything else.

timrid: I did this intentionally because in my (still private) rubicon-objc branch, the `initWithPyDelegate_` method has to be completely omitted, as it is not possible in rubicon-objc to pass Python objects to an ObjC function. It is only possible to assign Python objects to an ObjC-Object. So there,

```python
delegate = ObjcCentralManagerDelegate.alloc().initWithPyDelegate_(
     weakref.ref(self)
)
assert delegate is not None
```

become

```python
delegate = ObjcCentralManagerDelegate.alloc().init()
assert delegate is not None
delegate.py_delegate = weakref.ref(self)
```

And so I have to create the `weakref` outside of `ObjcCentralManagerDelegate`.

PR #1938, opened 2026-02-12
docs: update troubleshooting guide for macOS 26.2

ivanhercaz:
Hi,

I was reading the troubleshooting and following "Capture Bluetooth Traffic" for macOS, I discovered that at least on my macOS 26.2, once the profile is downloaded and "installed", the place in which it appears is "Device Management" instead of "Privacy and Security", so I added a tip for locating it.

image

It is my first contributing to this repository, so feel free to tell me whatever I need to do before continuing with this PR or if it is preferable another format for the tip. Whatever!

Regards

### DIFF
--- a/docs/troubleshooting.rst
+++ b/docs/troubleshooting.rst
@@ -374,10 +374,14 @@ You will need to sign up for an Apple Developer account to obtain this software.
     logging profile from the Apple develop `Profiles and Logs`_ page. Follow the
     instruction provided in the link, then continue with the steps below.
 
-    .. tip:: After installing the ``Bluetooth_macOS.mobileconfig`` file, the
-             profile can be found in *System Settings* under *Privacy and Security*
-             > *Others* > *Profiles*. You have to go there to actually install
-             the profile. Then reboot your computer.
+    .. tip:: After installing the ``Bluetooth_macOS.mobileconfig`` file, 
+            the profile can be found in *System Settings* under *Device
+            Management* > *Downloaded*. There you can double-click to review
+            and install it. You have to go there to actually install the profile.
+            Then reboot your computer.
+            
+            If you are on macOS 14.x, the profile can be found on *System Settings*
+            under *Privacy and Security* > *Others* > *Profiles*.
 
     If you have an older version of macOS, you can skip this step.
 


dlech:
Thanks for taking the time to do this. I noticed this the other day as well but didn't have time to update the docs right away.

Could we perhaps combine the two `.. tip` sections? I think it would be most relevant to put the new location first and then say that on older macOS versions, it is at the old location instead.

docs/troubleshooting.rst
```diff
@@ -374,10 +374,13 @@ You will need to sign up for an Apple Developer account to obtain this software.
     logging profile from the Apple develop `Profiles and Logs`_ page. Follow the
     instruction provided in the link, then continue with the steps below.
 
-    .. tip:: After installing the ``Bluetooth_macOS.mobileconfig`` file, the
-             profile can be found in *System Settings* under *Privacy and Security*
-             > *Others* > *Profiles*. You have to go there to actually install
-             the profile. Then reboot your computer.
+    .. tip:: After installing the ``Bluetooth_macOS.mobileconfig`` file, since
+            macOS 26.2 the profile can be found in *System Settings* under
```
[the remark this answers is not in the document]
ivanhercaz: Sorry for the delay pushing this easy change. I just commit the change and now I think the message is complete. I am going to mark this as resolved!

dlech: This still says 26.2. And we don't need to repeat the version twice in the same sentence.

In fact, I would say "since" or mention the version at all for the current location.

docs/troubleshooting.rst
```diff
@@ -374,10 +374,13 @@ You will need to sign up for an Apple Developer account to obtain this software.
...
+            *Device Management* > *Downloaded* since macOS 15.x. There you can
+            double-click to review and install it. If you are on previous macOS
+            If you are on macOS 14.x, the profile can be found on *System Settings* under *Privacy
```
dlech: Let's put the old location in a separate paragraph so that we don't overwhelm people with too much info at once.

docs/troubleshooting.rst
```diff
@@ -374,10 +374,13 @@ You will need to sign up for an Apple Developer account to obtain this software.
...
+            and Security* > *Others* > *Profiles*. You have to go there to actually
+            install the profile. Then reboot your computer.
```
dlech: And keep the last two sentences with the first paragraph.

docs/troubleshooting.rst
ivanhercaz: Ups! Thanks! I didn't catch it! I will keep it without mentioning versions for the current location.

Issue #883, opened 2022-07-13
Android app crash after lunch


Egidio92:
* bleak version: 0.14.3
* Python version: Python 3.7.13 (Colab)
* Operating System: Samsung SM-G900F Android 6.0.1, API 23

### Description #######################################################

Hello! I am trying to implement an application on an android mobile. This app needs to connect via low-energy bluetooth BLE to a scale (Mi Smart Scale 2) and take the weight. To create the apk of the application I use Colab.
The problem is that when I launch the application it crashes immediately after starting and I cannot read the weight.

### What I Did ########################################################

import asyncio

from kivy.app import App
from kivy.lang.builder import Builder
from kivy.uix.screenmanager import ScreenManager, Screen
from kivy.properties import ObjectProperty

from bleak import BleakClient

kv = '''
WindowManager:
    
    MainWindow:
    BluetoothWindow:
    LoginWindow:
    
        
     
    
<LoginWindow>:
    name: "login"

    email: email
    password: password

    FloatLayout:

        Label:
            text:"Email: "
            font_size: (root.width**2 + root.height**2) / 13**4
            pos_hint: {"x":0.1, "top":0.9}
            size_hint: 0.35, 0.15

        TextInput:
            id: email
            font_size: (root.width**2 + root.height**2) / 13**4
            multiline: False
            pos_hint: {"x": 0.45 , "top":0.9}
            size_hint: 0.4, 0.15

        Label:
            text:"Password: "
            font_size: (root.width**2 + root.height**2) / 13**4
            pos_hint: {"x":0.1, "top":0.7}
            size_hint: 0.35, 0.15

        TextInput:
            id: password
            font_size: (root.width**2 + root.height**2) / 13**4
            multiline: False
            password: True
            pos_hint: {"x": 0.45, "top":0.7}
            size_hint: 0.4, 0.15

        Button:
            pos_hint:{"x":0.2,"y":0.05}
            size_hint: 0.6, 0.2
            font_size: (root.width**2 + root.height**2) / 13**4
            text: "Login"
            on_release:
                app.root.current = "main"
                root.manager.transition.direction = "left"

        Button:
            pos_hint:{"x":0.2,"y":0.3}
            size_hint: 0.6, 0.1
            font_size: (root.width**2 + root.height**2) / 17**4
            text: "Non hai un account? Creane uno"  
            on_release:
     
                
     
        
     
<MainWindow>:
    name: "main"
    BoxLayout:
		orientation: "vertical"
[...]

PR #1944, opened 2026-03-18
Add Android backend using Chaquopy/Briefcase

### LINKED ISSUE #917: Bleak on Android with Chaquopy Android sdk
* bleak version: 0.15.0
* Python version: 3.8.13 (default, May 10 2022, 19:17:45)
* Operating System: Android (Python SDK for Android) https://chaquo.com/chaquopy/

### Description
I am trying to discover active Bluetooth devices using the following script from an Android app. 

```python
import asyncio

async def main():
    devices = await BleakScanner.discover()
    for d in devices:
        print(d)

asyncio.run(main())
````

got the following error: 

Screenshot 2022-08-02 at 12 36 25 PM

### PR DESCRIPTION
**Pull request by timrid**
This PR adds Android support for use with [briefcase](https://briefcase.readthedocs.io/) and [chaquopy](https://github.com/chaquo/chaquopy) as a separate backend.

Fixes #917

**Backend changes**
The existing python-for-android backend served as the foundation for the new backend, but several modifications were made. Some of them are:

- Migrated from python-for-android to chaquopy bindings
- Rewrote the dispatcher to be fully typed
- Added an `ExcessiveUsageChecker` because Android enforces a limit of 5 start/stop scan cycles within 30 seconds. This was necessary to allow the integration tests to run, as they would otherwise exceed this limit
- Implemented a `BroadcastReceiver`, which is part of python-for-android but not of chaquopy

I also noticed [your comment](https://github.com/hbldh/bleak/issues/1621#issuecomment-2363211989) that you would prefer permission handling to be explicit rather than implicit at scan start. I deliberately decided against this, because CoreBluetooth also handles permissions implicitly when starting a scan, and provides no way to request permissions independently without initiating a scan. To keep the backends as consistent as possible, I prefer the same approach on Android.

The old python-for-android backend has been left untouched for reference. Removing it would be better handled in a separate PR.

**Type hints**
For the type hints of the Android API, I created the tool [chaquopy-stubgen](https://github.com/timrid/chaquopy-stubgen) and created [chaquopy-stubs-android](https://pypi.org/project/chaquopy-stubs-android/). This enables type checking with mypy, and in the future with pyright once [microsoft/pyright#11221](https://github.com/microsoft/pyright/pull/11221) is released.

**Integration tests**
To run the integration tests, a briefcase test bed example has been added. This builds a test app that can be run on a real Android device or the Android emulator.

In CI, the integration tests run on the Android emulator, which is started automatically by briefcase. To simulate Bluetooth devices, Bumble is connected to the virtual Bluetooth controller provided by the Android emulator (see [Android Emulator transport](https://google.github.io/bumble/transports/android_emulator.html)). Since this virtual Bluetooth controller is normally only accessible from the host, the port is forwarded into the Android emulator via adb. Additionally, permission and pairing dialogs in the Android emulator must be confirmed — this is automated in CI using `uiautomator2`. All of this is encapsulated in the `run_android_tests_emulator.py` script and can be run with `uv run --python 3.13 testbed/run_android_tests_emulator.py`.

To run the integration tests on a real Android device, use the `run_android_tests_real_device.py` script. This requires an nRF52840 dongle with HCI-UART firmware and automatically forwards the UART to a TCP server that is tunneled to the device via adb. I tested this on a Samsung S23 running Android 13 (API level 33).

**Supported Python versions**
The target is Python 3.13 and later, as this is the first version in which Android is officially supported via [PEP 738](https://peps.python.org/pep-0738/). This is also the first version where `sys.platform` returns `android`, which simplifies backend detection and makes conditional dependencies possible.

Python 3.14 cannot yet be included in CI because some binary dependencies of Bumble are not yet available for Python 3.14. I am still working on this.

**Miscellaneous**
- I hope that, given the integration tests running in CI with over 80% code coverage, it is acceptable for the new Android backend to be listed as "Tier 1".

### CODE REVIEW

**Comment by timrid** [comment]
> * It would be nice to also split out the first couple of tests patches to separate PRs as those stand on their own.

Done in https://github.com/hbldh/bleak/pull/1959 and https://github.com/hbldh/bleak/pull/1958. Now that also https://github.com/hbldh/bleak/pull/1934 is merged, I rebased on the new develop branch.

>   * Is the backend called `android` with the idea to support multiple bindings (similar to `corebluetooth` with PyObjC/Rubicon ObjC)?

Yes, that was my intention. It is theoretically possible to integrate the `p4android` backend back into the new `android` backend, but as [I noted some time ago](https://github.com/hbldh/bleak/issues/917#issuecomment-3139481863) I got very frustrated with P4A. So I will not do that, but maybe someone else.

>   * For the test runner, it would be nice if it was under `tests` instead of `examples.

I think the `tests` folder should be only for tests and not for an testbed app. So I created a new `testbed` folder for the testbed app these. This testbed can then maybe also be used to run the integration tests for iOS.

>   * For the type hints, it would be nice to keep that all under `typings`

Done.

> * For the type stubs, it it looks like this is over 100k lines! Can we trim this down to just the modules we actually use? Or find a way to not vendor it?

I created https://github.com/timrid/chaquopy-stubs to outsource most of the stubs, so we just need to depend on [chaquopy-stubs-android](https://pypi.org/project/chaquopy-stubs-android/). But some of the stubs are special ones that can not be autogenerated so easy. So they need to be kept in `typings`.

File: bleak/backends/android/scanner.py [hunk 1]
```diff
@@ -0,0 +1,273 @@
...
+            self._scan_objs.callback.java
+        )
+
+        try:
+            # Wait a short time to check if "onScanFailed" is called.
+            await asyncio.wait_for(scanfuture, timeout=0.2)
+        except asyncio.exceptions.TimeoutError:
+            # If the scan started successfully but no device is found in the
+            # short waiting period, a timeout occurs. This is not an error.
+            pass
```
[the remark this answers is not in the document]
timrid: I simply copied `flushPendingScanResults` over from the P4A backend. I don't understand why it's needed. We don't use batch mode because we use `setReportDelay(0)`. So I've now removed it entirely.

I’ve also completely removed the raising of unhandled exceptions in the dispatcher and replaced it with a log message. I had simply copied this over from the old P4A backend, but I couldn’t find any indication in the commit history as to why it was needed. Furthermore, I couldn’t reproduce this scenario in my tests, which is why I believe it shouldn’t occur during normal operation and that logging is sufficient.

bleak/backends/android/scanner.py
```diff
@@ -0,0 +1,273 @@
...
+                    BleakBluetoothNotAvailableReason.POWERED_OFF,
+                )
+
+            javascanner = adapter.getBluetoothLeScanner()
+
+            self._scan_objs = ScanObjects(
+                adapter=adapter,
+                javascanner=javascanner,
+                callback=callback,
+            )
```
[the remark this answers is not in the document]
timrid: According to the documentation, `getBluetoothLeScanner()` returns `None` if Bluetooth is turned off. This is actually checked beforehand. However, for the sake of completeness, I’ve also added an exception here. To do this, I also had to [modify](https://github.com/timrid/chaquopy-stubgen/pull/14/changes) chaquopy-stubgen.

bleak/backends/android/scanner.py
```diff
@@ -0,0 +1,273 @@
...
+        detection_callback:
+            Optional function that will be called each time a device is
+            discovered or advertising data has changed.
+        service_uuids:
+            Optional list of service UUIDs to filter on. Only advertisements
+            containing this advertising data will be received. Specifying this
+            also enables scanning while the screen is off on Android.
+        scanning_mode:
+            Set to ``"passive"`` to avoid the ``"active"`` scanning mode.
+    """
```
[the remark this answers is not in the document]
timrid: I removed the python-for-android wording.

bleak/backends/android/client.py
```diff
@@ -0,0 +1,611 @@
...
+            dispatch_func=dispatch_func(
+                device.connectGatt,
+                context,
+                False,
+                callbacks.java,
+                BluetoothDevice.TRANSPORT_LE,
+            ),
+            callback_api=OnConnectionStateChangeCallback(),
+            dispatch_result_indicates_status=False,
+        )
```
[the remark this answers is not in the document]
timrid: I've also added a `check_for_permissions` function to BleakClient in case you use BleakClient without having used BleakScanner first.

bleak/backends/android/client.py
```diff
@@ -0,0 +1,611 @@
...
+
+        timeout = kwargs.get("timeout", self._timeout)
+
+        adapter = BluetoothAdapter.getDefaultAdapter()
+        assert adapter is not None
+        if adapter.getState() != BluetoothAdapter.STATE_ON:
+            raise BleakBluetoothNotAvailableError(
+                "Bluetooth is not turned on",
+                BleakBluetoothNotAvailableReason.POWERED_OFF,
+            )
```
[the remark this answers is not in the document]
timrid: I have added an explicit exception.

bleak/backends/android/client.py
```diff
@@ -0,0 +1,611 @@
...
+            if conn_result.new_state != BluetoothProfile.STATE_CONNECTED:
+                raise BleakError(
+                    f"Unexpected connection state: {conn_result.new_state}"
+                )
+            logger.debug(f"{OnConnectionStateChangeCallback()} succeeded")
+        except BaseException:
+            # If connecting is canceled, times out, or fails, we need to disconnect to clean up the gatt connection.
+            gatt.disconnect()
+            gatt.close()
+            raise
```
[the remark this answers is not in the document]
timrid: I added an explicit error message if `gatt` is `None` and [update chaquopy-stubgen](https://github.com/timrid/chaquopy-stubgen/pull/15/changes) to reflect the connectGatt can return `None`.

bleak/backends/android/client.py
```diff
@@ -0,0 +1,611 @@
...
+        else:
+            raise BleakError(
+                f"Characteristic {characteristic.uuid} does not support Notify or Indicate"
+            )
+
+        await self.write_gatt_descriptor(
+            cccd,
+            cccd_value,
+        )
+
```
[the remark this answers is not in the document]
timrid: I added a try/except cleanup.

bleak/backends/android/permissions.py
```diff
@@ -0,0 +1,126 @@
...
+    ):  # pragma: no cover  # Can not be tested in CI, because netsim requires min. API 31
+        return [
+            Manifest.permission.ACCESS_FINE_LOCATION,
+            Manifest.permission.ACCESS_BACKGROUND_LOCATION,  # optional: only if scanning BLE devices in background
+        ]
+    elif api_level > 30:
+        return [
+            Manifest.permission.BLUETOOTH_SCAN,
+            Manifest.permission.BLUETOOTH_CONNECT,
+        ]
```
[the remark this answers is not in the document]
timrid: I removed ACCESS_BACKGROUND_LOCATION and added a comment explaining why it is not listed.

bleak/backends/android/dispatcher.py
```diff
@@ -0,0 +1,190 @@
...
+        )
+
+        callback_result = await state
+
+        logger.debug(f"{callback_api} succeeded {callback_result}")
+
+        if dispatch_result_indicates_status:
+            return callback_result
+        else:
+            return (dispatch_result, callback_result)
```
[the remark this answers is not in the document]
timrid: You can actually just delete the future from `self.futures` after you've awaited it. However, you can't just delete the state, since you might need it later. I don't think this is a problem, though, since the keys in both dictionaries are *frozen* dataclass instances and there are only a limited number of `CallbackApi` objects. For `CallbackApi` objects with the `uuid` parameter, there is exactly one entry in the dictionary for each UUID read. Since in reality you only read a limited number of UUIDs, this shouldn’t be a problem.

bleak/backends/android/dispatcher.py
```diff
@@ -0,0 +1,190 @@
...
+                    if not namedfuture[1].done()
+                ]
+                if len(namedfutures):
+                    # send it on existing requests
+                    for name, future in namedfutures:
+                        warnings.warn(f"Redirecting error without home to {name}")
+                        future.set_exception(exception)
+                else:
+                    # send it on the event thread
+                    raise exception
```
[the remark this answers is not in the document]
timrid: Errors are now only logged.

tests/integration/test_issue_1885.py
```diff
@@ -20,6 +22,9 @@
 TEST_CHARACTERISTIC_UUID = "e809cb2f-34e3-42a1-ba92-22db2495cd6a"
 
 
+@pytest.mark.skipif(
+    sys.platform == "android", reason="On Android this tests times out."
+)
```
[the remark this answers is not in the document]
timrid: Fixed it.

testbed/README.rst
```diff
@@ -0,0 +1,3 @@
+## Briefcase Testbed Application
+
+This is a briefcase testbed application for running the integration tests the Android backend. For more details see `tests/integration/README.rst <tests/integration/README.rst>`_.
```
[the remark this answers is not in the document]
timrid: Fixed it.

docs/backends/android-beeware.rst
```diff
@@ -0,0 +1,85 @@
+Android backend (BeeWare)
+=========================
+
+This backend is intendet for Android apps using the `BeeWare <https://beeware.org/>`_ toolchain:
```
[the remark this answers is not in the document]
timrid: Fixed it.

timrid:
I've integrated Copilot's comments into the code and added a few points of my own.

The CI is currently generating the error `Unable to download pixel_7_pro device skin; is your computer offline?`, but this has nothing to do with the recent changes. I hope this is just a temporary glitch.

bleak/backends/android/client.py
```diff
@@ -0,0 +1,654 @@
...
+                bytes(BluetoothGattDescriptor.ENABLE_NOTIFICATION_VALUE),
+            )
+        elif props & BluetoothGattCharacteristic.PROPERTY_INDICATE:
+            cccd_value = bytes(BluetoothGattDescriptor.ENABLE_INDICATION_VALUE)
+        elif props & BluetoothGattCharacteristic.PROPERTY_NOTIFY:
+            cccd_value = bytes(BluetoothGattDescriptor.ENABLE_NOTIFICATION_VALUE)
+        else:
+            raise BleakError(
+                f"Characteristic {characteristic.uuid} does not support Notify or Indicate"
+            )
```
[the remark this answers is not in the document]
timrid: Removed the `bitwise_or` case.

bleak/backends/android/permissions.py
```diff
@@ -0,0 +1,131 @@
...
+        if all(grant_results):
+            loop.call_soon_threadsafe(
+                permission_acknowledged.set_result,
+                grant_results,
+            )
+        else:
+            loop.call_soon_threadsafe(  # pragma: no cover  # Difficult to test user denial in CI. This would require a second run of the test suite with different user interaction.
+                permission_acknowledged.set_exception,
+                BleakBluetoothNotAvailableError(
+                    f"User denied access to {permissions}",
```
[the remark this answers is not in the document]
timrid: Fixed.

testbed/pyproject.toml
```diff
@@ -0,0 +1,93 @@
...
+    # So we just copy the bleak files (see 'sources').
+    # "..",
+    
+    # This bugfix https://github.com/beeware/toga/pull/4083 is needed. 
+    # The current version Toga v0.5.3 does not include this fix.
+    # It is needed so that grpcio can be used on Android.
+    # "git+https://github.com/beeware/toga.git@main",
+    "git+https://github.com/beeware/toga.git#egg=travertino&subdirectory=travertino",
+    "git+https://github.com/beeware/toga.git#egg=toga-android&subdirectory=android",
+    "git+https://github.com/beeware/toga.git#egg=toga-core&subdirectory=core",
```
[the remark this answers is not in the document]
timrid: Fixed.

bleak/backends/android/client_callback.py
```diff
@@ -0,0 +1,260 @@
...
+            # On API level 32 (Android 12) and below
+            value = characteristic.getValue()
+        else:
+            # On API level 33 (Android 13) and above
+            value = args[0]  # pragma: no cover  # (CI is running on API level below 33)
+
+        conn_objs = self._client._conn_objs  # pyright: ignore[reportPrivateUsage]
+        if conn_objs is not None:
+            self._loop.call_soon_threadsafe(
+                conn_objs.subscriptions[handle],
```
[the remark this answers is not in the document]
timrid: Fixed.

bleak/backends/android/scanner.py
```diff
@@ -0,0 +1,274 @@
...
+            dispatch_result_indicates_status=False,
+        )
+
+        try:
+            # Wait a short time to check if "onScanFailed" is called.
+            await asyncio.wait_for(scanfuture, timeout=0.2)
+        except asyncio.exceptions.TimeoutError:
+            # If the scan started successfully but no device is found in the
+            # short waiting period, a timeout occurs. This is not an error.
+            pass
```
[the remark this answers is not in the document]
timrid: Fixed.

timrid:
I've fixed all new Copilot comments.

The CI failure was [fixed in `briefcase`](https://github.com/beeware/briefcase/pull/2788) but has not been released yet. So I use the current git main branch to get the CI working again.

testbed/src/bleak_testbed/__main__.py
```diff
@@ -0,0 +1,4 @@
+from bleak_testbed.app import main
```
dlech, line 1: These testbed files could use some comments/docstrings at the start of each file to explain why they are here.

pyrightconfig-android.json
```diff
@@ -0,0 +1,16 @@
+{
+  "pythonPlatform": "Android",
+  "typeCheckingMode": "strict",
+  "exclude": [
+    "**/.venv/**",
+    "**/kivy/**",
+    "**/recipes/**",
+    "**/testbed/**",
```
dlech, line 8: Should testbed be still be excluded here?

testbed/README.rst
dlech, line 3: Let's try to wrap to approx 80 characters per line.

testbed/run_android_tests_emulator.py
```diff
@@ -0,0 +1,312 @@
...
+                len(parts) >= 2
+                and parts[1] == "device"
+                and parts[0].startswith("emulator-")
+            ):
+                serial = parts[0]
+                try:
+                    name_output = adb.call_adb(["-s", serial, "emu", "avd", "name"])
+                    if name_output.splitlines()[0].strip() == avd_name:
+                        return serial
+                except Exception:
```
dlech, line 86: Let's add some comments explaining why it is safe to ignore exceptions here and below.

Or log something to help with debugging.

docs/backends/android-beeware.rst
```diff
@@ -0,0 +1,91 @@
...
+
+Backend Specific Quirks
+-----------------------
+
+On Android, no more than 5 start/stop scanning operations are allowed per 30 seconds! See also 
+`this issue comment <https://github.com/NordicSemiconductor/Android-Scanner-Compat-Library/issues/18#issuecomment-402412139>`_ or
+`this PR in the Android OS <https://android-review.googlesource.com/c/platform/packages/apps/Bluetooth/+/215844>`_.
+If this limit is exceeded in a normal Android application, scanning simply won't work without 
+producing an error. Therefore, Bleak automatically tracks the starting times of scans 
+and waits the necessary time before a new scan can be started.
```
dlech, line 73: Not sure how I feel about this feature. Seems like it could have unintended consequences. Like if someone connects BleakScanner directly to UI that lets user start/stop scanning, it seems like it would make the UI unresponsive if they pressed the start/stop button a bunch. It would say that it was starting scanning, but would actually be much delayed with no user feedback.

I would be more inclined to just raise an error and let the application deal with it.

docs/backends/android-p4a.rst
```diff
@@ -1,5 +1,8 @@
-Android backend
-===============
+Android backend (Python-for-Android)
+====================================
+
+**Warning: This backend is not tested regularly and may contain bugs! If you want to use a
```
dlech, line 4: Why not using `.. warning:: `?

docs/backends/android-p4a.rst
dlech, line 1: Renaming the `android.rst` file without replacing it will break existing web links.

Not sure how much I care if that happens, but want to make sure we are intentional about it if we do that.

mypy-android.ini
```diff
@@ -0,0 +1,14 @@
+# Use this with 
+#   mypy --config-file=mypy-android.ini
+
+[mypy]
+python_version = 3.10
```
dlech, line 5:
```suggestion
python_version = 3.13
```

tests/integration/test_client_pairing.py
```diff
@@ -26,4 +26,25 @@ async def test_pairing_unavailable(bumble_peripheral: Device):
         await client.unpair()
 
 
+@pytest.mark.skipif(
+    get_default_backend() != BleakBackend.ANDROID,
+    reason="pairing is only possible on Android backend",
```
dlech, line 31: Some other backends have pairing implemented, so this statement seems misleading.

tests/integration/test_issue_1885.py
```diff
@@ -20,6 +22,7 @@
 TEST_CHARACTERISTIC_UUID = "e809cb2f-34e3-42a1-ba92-22db2495cd6a"
 
 
+@pytest.mark.skipif(sys.platform == "android", reason="On Android this test times out.")
```
dlech, line 25: Could this be a bug (race condition) in the implementation?

Based on user feedback, there are quite a few devices with this quirk, so could be worth trying to fix.

The code looks correct though, so I'm not sure why it wouldn't be working.

bleak/backends/android/client_callback.py
```diff
@@ -0,0 +1,264 @@
...
+        *args: Any,
+    ):
+        logger.debug("onCharacteristicChanged")
+        handle = characteristic.getInstanceId()
+        if len(args) == 0:
+            # On API level 32 (Android 12) and below
+            value = characteristic.getValue()
+        else:
+            # On API level 33 (Android 13) and above
+            value = args[0]  # pragma: no cover  # (CI is running on API level below 33)
```
dlech, line 165: Is the CI Android version tied to the CI Ubuntu version?

bleak/backends/android/client.py
```diff
@@ -0,0 +1,646 @@
...
+            assert isinstance(java_service, BluetoothGattService)
+            if (
+                self._requested_services is not None
+                and java_service.getUuid() not in self._requested_services
+            ):
+                continue
+
+            service = BleakGATTService(
+                java_service,
+                java_service.getInstanceId(),
```
dlech, line 365: This is inheriting a long-known bug that the instance ID is not globally unique (unless Android changed the implementation and not the docs)?

We need to make a helper function that turns the service (and characteristic) UUIDs into an integer and combine that with the instance ID to get a globally unique number.

The number doesn't have to be stable between runs, so we could do something like:

```python
from itertools import count

...

_attribute_handle_generator = count()
_attribute_handles: dict[tuple[str, int], int] = {}

def attribute_to_handle(attribute: BluetoothGattService | BluetoothGattCharacteristic) -> int:
    key = str(attribute.getUuid()), attribute.getInstanceId()
    handle = _attribute_handles.get(key)

    if handle is None:
        _attribute_handles[key] = handle = next(_attribute_handle_generator)

    return handle
```

Not sure if `getInstanceId()` for characteristics would be globally unique per characteristic UUID or if it is unique per parent service.

In any case, it would be good to add a test with multiple services and characteristics with the same UUID on a single device to make sure we are able to access each one correctly.

bleak/backends/android/client.py
```diff
@@ -0,0 +1,646 @@
...
+                )
+                services.add_characteristic(characteristic)
+
+                for descriptor_index, java_descriptor in enumerate(
+                    java_characteristic.getDescriptors().toArray()
+                ):
+                    assert isinstance(java_descriptor, BluetoothGattDescriptor)
+                    descriptor = BleakGATTDescriptor(
+                        java_descriptor,
+                        characteristic.handle + 1 + descriptor_index,
```
dlech, line 388: We'll also need to consider how this fits into how we handle `getInstanceId()`.

It is probably still OK since we never use handle to determine the attribute type. All descriptor handles for a single characteristic should still be unique for that characteristic.

bleak/backends/android/client_callback.py
```diff
@@ -0,0 +1,264 @@
...
+    )
+    @external_thread_callback
+    def onCharacteristicChanged(
+        self,
+        gatt: BluetoothGatt,
+        characteristic: BluetoothGattCharacteristic,
+        *args: Any,
+    ):
+        logger.debug("onCharacteristicChanged")
+        handle = characteristic.getInstanceId()
```
dlech, line 159: As in the other comment about `getInstanceId()`, this and others will need to use the same `attribute_to_handle()` function.

.github/workflows/build_and_test.yml
```diff
@@ -106,3 +106,80 @@ jobs:
...
+          activate-environment: true
+
+      - name: Setup Environment
+        run: |
+          # Use GitHub's preinstalled JDK 17 for Android builds
+          echo JAVA_HOME="${JAVA_HOME_17_X64:-$JAVA_HOME_17_arm64}" | tee -a ${GITHUB_ENV}
+          # Enable KVM permissions for the emulator
+          echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' \
+            | sudo tee /etc/udev/rules.d/99-kvm4all.rules
+          sudo udevadm control --reload-rules
```
dlech, line 140: Modern systems have file watchers that does this automatically, so I've never needed to manually reload rules.

Maybe there could be a race condition though in a script, so I guess it doesn't hurt.

bleak/backends/android/client.py
```diff
@@ -0,0 +1,646 @@
...
+        payload = jarray(jbyte)(bytes(data))
+
+        if Build.VERSION.SDK_INT >= 33:
+
+            def _do_write_char():  # pragma: no cover  # (CI is running on API level below 33)
+                # On API level 33 (Android 13) and above writeCharacteristic returns int (GATT_SUCCESS=0 on success).
+                # Convert to bool so dispatch_result_indicates_status=True works correctly.
+                return (
+                    gatt.writeCharacteristic(characteristic.obj, payload, write_type)
+                    == 0
```
dlech, line 505: Ideally, we would be capturing the return value and including it in the raised exception to help the user find out what went wrong. At the very least, we should debug log it.

bleak/backends/android/client.py
```diff
@@ -0,0 +1,646 @@
...
+
+        gatt = self._conn_objs.gatt
+        payload = jarray(jbyte)(bytes(data))
+
+        if Build.VERSION.SDK_INT >= 33:
+
+            def _do_write_desc():  # pragma: no cover  # (CI is running on API level below 33)
+                # On API level 33 (Android 13) and above writeDescriptor returns int (GATT_SUCCESS=0 on success).
+                # Convert to bool so dispatch_result_indicates_status=True works correctly.
+                return gatt.writeDescriptor(descriptor.obj, payload) == 0
```
dlech, line 550: Another case to capture the error.

bleak/backends/android/dispatcher.py
```diff
@@ -0,0 +1,181 @@
...
+        self,
+        failure: Exception | None,
+        callback_api: CallbackApi[CallbackResultT],
+        callback_result: CallbackResultT,
+    ):
+        self._loop.call_soon_threadsafe(
+            self._result_state_unthreadsafe, failure, callback_api, callback_result
+        )
+
+    def _result_state_unthreadsafe(
```
dlech, line 156:
```suggestion
    def _result_state(
```

I think not-threadsafe can be generally assumed.

bleak/backends/android/client_callback.py
```diff
@@ -0,0 +1,264 @@
...
+    def onCharacteristicWrite(
+        self,
+        gatt: BluetoothGatt,
+        characteristic: BluetoothGattCharacteristic,
+        status: int,
+    ):
+        logger.debug(f"onCharacteristicWrite {status=}")
+        handle = characteristic.getInstanceId()
+        self.dispatcher.result_state_threadsafe(
+            BleakGATTProtocolError(int(status)) if status != GATT_SUCCESS else None,
```
dlech, line 218:
```suggestion
            BleakGATTProtocolError(status) if status != GATT_SUCCESS else None,
```

Or type hint of `status` is wrong.

dlech (commented):
This is huge, so will probably come back with more comments later. Here are a few hours worth...

.github/workflows/build_and_test.yml
dlech: Or since this is CI, just get rid of the udev stuff and just run `sudo chmod` directly.

docs/backends/android-p4a.rst
timrid: When the plan is to remove the P4A backend fully I could rename `android-beeware.rst` to just `android.rst` as this is then the only android backend.

tests/integration/test_client_pairing.py
timrid: I added a hint, that *testing* is only implemented for the android backend.

testbed/README.rst
timrid: Done

docs/backends/android-p4a.rst
timrid: Done

mypy-android.ini
timrid: Done

bleak/backends/android/client_callback.py
timrid: The Android API level currently depends on the default Android emulator image that is used by briefcase. Currently the [default of briefcase](https://github.com/beeware/briefcase/blob/0332fe66e22d094d938dc24ceb4ba9431ce76558/src/briefcase/integrations/android_sdk.py#L187)  `system-images;android-31;default;<EMULATOR_ABI>` is used. So the integration tests are running on API level 31.

Today I did some testing of the other API Levels. API levels 32 and 33 are also working with a minor adjustments. API level 34, 35 and 36 make some trouble on pairing test, everything else is working. So I guess I can create a CI matrix for all working API levels and this `#pragma: no cover` can be removed. But for API level 30 I had no success, because I could not connect to the `netsim` bluetooth emulator of the android emulator. Maybe older Android versions don't support `netsim`.

I will investigate further...

timrid: I now have the integration tests running locally on all versions between 31 and 36. However, on versions 35 and 36, I frequently encounter [an error](https://github.com/beeware/briefcase/issues/2809) in CI that I can't reproduce locally. To get rid of the `pragma: no cover`, we don’t need to test all API levels anyway, just <33 and >=33. That’s why I’ve only enabled API levels 31 and 33 in the CI. These give us the necessary test coverage.

testbed/run_android_tests_emulator.py
timrid: I am not sure anymore why they are necessary, because everything works without them. So I removed them completely. If this should make problems, we can add the try-except blocks again with comments why they are needed.

.github/workflows/build_and_test.yml
timrid: I just followed the recommendations from Github: https://github.blog/changelog/2024-04-02-github-actions-hardware-accelerated-android-virtualization-now-available/

I added the link also to the source code, so that the source of these commands is clear.

bleak/backends/android/client_callback.py
dlech: I think supporting just the last 5 years of releases is fine, so I wouldn't spend much time trying to make API level 30 working.

.github/workflows/build_and_test.yml
dlech: Just because they recommended it doesn't mean they understood it. :wink:

timrid:
I accidentally synced my branch to the develop via the Github UI, which caused the PR to be closed... That was by accident. But now I am not sure how to reopen this PR. Sorry for the inconvenience...

timrid:
I finally got it reopened. Sorry for the noise.

tests/integration/test_issue_1885.py
timrid: That seemed to be an issue with `notify_subscribers` (plural). If you use `notify_subscriber` (singular) and include the connection, it works on Android as well. I'm not sure what the problem with `notify_subscribers` was, but now it works on Android too.

bleak/backends/android/client.py
timrid: I refactored the `dispatcher` a bit, so that it is simplified and it is also possible to raise an exception in the `dispatch_func`. So it is possible to raise an error message here that includes the return status.

bleak/backends/android/client.py
timrid: Same answer as in https://github.com/hbldh/bleak/pull/1944#discussion_r3177203555

bleak/backends/android/dispatcher.py
timrid: I changed the method name.

bleak/backends/android/client_callback.py
timrid: The type hint is correct, so the explicit conversion to `int` is not necessary.

This is also [documented in the chaquopy docs](https://chaquo.com/chaquopy/doc/17.0/python.html#notes):
> When a Python method is called from Java, the parameters and return value are converted as described in [data types](https://chaquo.com/chaquopy/doc/17.0/python.html#data-types) above.

testbed/src/bleak_testbed/__main__.py
timrid: Actually the files in `testbed/src/bleak_testbed` are not needed, because we only use the `--test` mode of `briefcase`. So I removed these files. But there has to be at least a folder called `testbed/src/bleak_testbed` or `briefcase` will raise a error.

For the `testbed/tests/bleak_testbed.py` I added a comment.

bleak/backends/android/client.py
timrid: You’re probably referring to the bug at https://github.com/hbldh/bleak/issues/883#issuecomment-1183468846? That bug was found in Android 6 with API 23.

So before I tried to fix the problem, I wanted to be able to reproduce it using an integration test. So I created some tests, but the problem doesn't occur for me in the emulator or on a real device with API 31. That's why I took a closer look at the issue.

I had Claude search the Android source code, and it appears that the behavior changed in Android 8 (API 26). From then on, the instance ID is supposed to be globally unique. [Here](https://android.googlesource.com/platform/packages/apps/Bluetooth/+/418f1c1e242313ab8b83eb9e38ffeee52e6e9445) is the corresponding commit that implemented the change. This means the bug only affects Android 7 (API 25) and earlier.

I also took another look at the documentation for [BluetoothGattCharacteristic.getInstanceId](https://developer.android.com/reference/android/bluetooth/BluetoothGattCharacteristic#getInstanceId()):

> Returns the instance ID for this characteristic.
>
> If a remote device offers multiple characteristics with the same UUID, the instance ID is used to distinguish between characteristics.

As I understand it, this does not explicitly state whether the instance ID is globally unique or not. So the docs does not really help here.

Of course, we can make the change as you suggested, but we have no way to test it in the CI. The integration tests in the Android emulator only run on API 31 and above, and I don't have a real device with an older Android version to test whether the change actually works correctly. Given that Android 7 is from 2016 and is therefore already 10 years old, the question arises as to whether this actually needs to be fixed.

docs/backends/android.rst
```diff
@@ -0,0 +1,91 @@
...
+
+Backend Specific Quirks
+-----------------------
+
+On Android, no more than 5 start/stop scanning operations are allowed per 30 seconds! See also 
+`this issue comment <https://github.com/NordicSemiconductor/Android-Scanner-Compat-Library/issues/18#issuecomment-402412139>`_ or
+`this PR in the Android OS <https://android-review.googlesource.com/c/platform/packages/apps/Bluetooth/+/215844>`_.
+If this limit is exceeded in a normal Android application, scanning simply won't work without 
+producing an error. Therefore, Bleak automatically tracks the starting times of scans 
+and waits the necessary time before a new scan can be started.
```
timrid: The problem with throwing an exception is that the integration tests in CI will then stop running...

That was actually my original problem, which is how I became aware of this issue in the first place: the integration tests just got stuck. The logs only showed

```
D/BluetoothLeScanner: onScannerRegistered() - status=6 scannerId=-1 mScannerId=0
```

and the scan didn’t work. There were no other error message or exception besides the log.

That’s why I implemented it so that it blocks automatically. 

However, it might also be possible to make this behavior configurable via an Android-specific parameter. The default would then be to throw the exception, and the parameter could be used to enable automatic waiting. This parameter would then be used in the integration tests.

What do you think?

bleak/backends/android/client.py
dlech: Thanks for the investigation. I guess that explains why no one has complained about that for a while even though we never fixed it.

If we have a test with duplicate service UUIDs and duplicate characteristic UUIDs and it passes, we are good.

And we can probably close that issue. :smile:

PR #1958, opened 2026-03-25
Integration tests: Correct RSSI check

timrid:
According to the bluetooth spec the RSSI can vary between -127 and 20.

In virtual bluetooth devices of the Android Emulator an RSSI of 20 is returned.


dlech:
Thanks!

PR #1959, opened 2026-03-25
tests: Update bumble to v0.0.226

timrid:
This updates bumble to v0.0.226. This is mainly necessary for my work on the [new Android backend](https://github.com/hbldh/bleak/pull/1944) and I need [these changes](https://github.com/google/bumble/pull/869) in bumble.

After the update I tried the integration tests with macOS and Windows locally without problems. Linux is tested via the CI.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,7 +57,7 @@ test = [
     "pytest>=8.2.1,<9.0.0",
     "pytest-asyncio>=1.2.0,<2.0.0",
     "pytest-cov>=7.0.0,<8.0.0",
-    "bumble==0.0.220",
+    "bumble==0.0.226",
 ]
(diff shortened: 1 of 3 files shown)


dlech:
Can we also drop the type ignore comments on `notify_subscribers()` calls? I fixed that in Bumble a while back, so hopefully type hints should be correct now.

timrid:
> Can we also drop the type ignore comments on `notify_subscribers()` calls? I fixed that in Bumble a while back, so hopefully type hints should be correct now.

Unfortunately, no. The stubs are still incomplete because only `Attribute` is specified. I guess correct would be `Attribute[Any]`.

image

dlech:
Thanks!

Issue #320, opened 2020-10-05
A way to detect BT module not being present or disabled at all


SteKme:
* bleak version: 0.8.0
* Python version: 3.7x64
* Operating System: Windows 10

### Description
Discovery service - BleakScanner.discover() - returns [] no matter if 
** there are no devices the PC can connect to (which is fine, indeed), 
** but also when the BT module of the PC is turned off (my case, notebook's BT module off in the right hand command control center of Windows)
** ? or the BT module not present at all (can't check if this applies as well at the moment).

I can't seem to find any way to detect the BT module of the PC not being present / disabled in the Windows. Would it be perhaps possible to throw an exception or something in this discover() when there is no BTLE hardware connected to the PC at all? Any other working solution is fine, of course.

Thank You so much!

hbldh:
Currently, Bleak has no capability to check if your computer's Bluetooth adapter is on or off. There are API methods that can check it, and in one case it seems it can also switch them on and off (BlueZ).

- BlueZ: https://kernel.googlesource.com/pub/scm/bluetooth/bluez/+/refs/tags/5.52/doc/adapter-api.txt#243
- Windows: https://stackoverflow.com/a/48794678 and https://docs.microsoft.com/en-us/uwp/api/windows.devices.bluetooth.bluetoothadapter.getradioasync?view=winrt-19041#Windows_Devices_Bluetooth_BluetoothAdapter_GetRadioAsync
- Core Bluetooth: This might be used as a hint: https://stackoverflow.com/a/57953922

It is not a priority task for me I am afraid. You are more than welcome to do the implementation and make a PR, because I approve of the functionality being present.

dlech:
We are actually already monitoring this in CoreBluetooth:

https://github.com/hbldh/bleak/blob/9ff5ed8ccbb0493e2994a0bbfe9ebf26810878fb/bleak/backends/corebluetooth/CentralManagerDelegate.py#L178-L195

But the state is not saved anywhere.

We should also revisit the use of `wait_for_powered_on()` in CoreBluetooth when we look at this. According to the CoreBluetooth docs, we should be waiting for this to be resolved before calling any other APIs, but I don't think we are doing that in all cases. (Note to self: we can probably use a `threading.Event` instead of `asyncio.Event` and make the `__init__()` method blocking when we create a new central manager - it shouldn't block for long.)

0ge:
[...]

PR #1966, opened 2026-04-09
Add BleakAdapter with get_connected_devices() to retrieve already connected devices

Vodur:
## Summary                                                                                                                                                                                                                        
  
  Adds a new `BleakAdapter` class with an async `get_connected_devices()` method to query the OS for BLE devices that are already connected, without requiring an active scan. The returned `BLEDevice` objects can be passed         directly to `BleakClient` to use the existing OS-level connection.                                                                                                                                                                
                                                                                                                                                                                                                                    
  Resolves part of #1060.

  - **Windows**: Uses `BluetoothLEDevice.get_device_selector_from_connection_status()` with `DeviceInformation.find_all_async()` to enumerate connected BLE devices.
  - **macOS**: Uses `CBCentralManager.retrieveConnectedPeripheralsWithServices()` filtered by service UUID. `BleakClient.connect()` now also tries `retrievePeripheralsWithIdentifiers:` before falling back to scanning, so        
  already-known peripherals connect without a scan.
  - **Linux (BlueZ)**: Adds `BlueZManager.get_connected_devices()` to query D-Bus for connected BLE devices on the selected adapter, filtered by service UUID. Adapter selection uses a typed `BlueZAdapterArgs` dict (e.g.
  `bluez={"adapter": "hci1"}`).
  - Other backends raise `NotImplementedError`.

  Adapter instances are obtained via the async `BleakAdapter.get()` factory.

  ## Motivation

  There is currently no way to retrieve already-connected BLE devices from the OS without scanning. Many applications need to interact with devices the OS has already connected, and an active scan is wasteful and adds latency.  
  This is also the foundation for adapter state and adapter-level callbacks tracked in #1060.

  ## Commits

  1. **`backends/winrt: simplify BleakClientWinRT.__init__ to use address string`** — independent refactor: derive the integer address from `BLEDevice.address` instead of unpacking `RawAdvData`.
  2. **`feat: add BleakAdapter.get_connected_devices() to retrieve already-connected BLE devices`** — main feature, all platform implementations, public facade, example, typings, changelog.
  3. **`typings/corebluetooth: fix retrievePeripheralsWithIdentifiers parameter type`** — standalone typing fix (`identifiers: NSArray[NSUUID]` instead of `serviceUUIDs: NSArray[CBUUID]`).
  4. **`feat(macos): try retrievePeripheralsWithIdentifiers before scanning`** — uses the typing fix; lets `BleakClient` connect to already-known peripherals without scanning.

  ## Test plan

  - [x] `python examples/connected_devices.py` lists connected devices and their services
  - [x] `python examples/connected_devices.py --service-uuids 180F` filters by service
  - [x] `python examples/connected_devices.py --details` also prints characteristics and descriptors
  - [x] Manually tested on Windows 11, macOS, and Ubuntu with real connected devices
  - [x] `uv run pyright` and `uv run mypy -p bleak -p tests -p examples` pass
  - [x] `black`, `isort`, `flake8` pass


dlech:
Thanks for getting this started.

We'll also want automated tests for this and a changelog entry. And ideally Linux support.

I've just had a quick look for now. I was wondering why we should use IOBluetooth instead of `retrievePeripheralsWithIdentifiers:`. Does the latter remember all devices that have ever been connected even if they were not paired/bonded?

Vodur:
> Thanks for getting this started.
> 
> We'll also want automated tests for this and a changelog entry. And ideally Linux support.
> 
> I've just had a quick look for now. I was wondering why we should use IOBluetooth instead of `retrievePeripheralsWithIdentifiers:`. Does the latter remember all devices that have ever been connected even if they were not paired/bonded?

Hey David,
I started this locally to help with my project, and then I thought why not share with everyone.

Regarding the question about IOBluetooth vs retrievePeripheralsWithIdentifiers:. The latter requires you to already know the CoreBluetooth UUID of the device — it's a lookup by ID, not an enumeration API. CoreBluetooth has no way to list all paired/bonded devices. The only APIs it offers are:
  - scanForPeripheralsWithServices: — active scan (requires advertising) 
  - retrieveConnectedPeripheralsWithServices: — only currently-connected peripherals
  - retrievePeripheralsWithIdentifiers: — lookup by known UUID (not enumeration)

IOBluetooth is the only macOS framework that can enumerate all paired Bluetooth devices. We then use retrieveConnectedPeripheralsWithServices + the undocumented retrieveAddressForPeripheral: to resolve CoreBluetooth UUIDs for connected devices, so the returned BLEDevice objects work directly with BleakClient.

I'll work on adding:
   1. Linux (BlueZ) backend support
   2. Automated tests
   3. Changelog entry

dlech:
It could also be easier to work on this if we split it up a bit. For example, we could just start with enumerating and "connecting" to already connected devices. Then handle already paired devices separately.

Vodur:
That makes sense — I'll split this up.

Phase 1: Enumerate already-connected BLE devices and allow connecting to them without scanning. This avoids the IOBluetooth dependency on macOS since
retrieveConnectedPeripheralsWithServices: handles the connected case natively in CoreBluetooth.

Phase 2: Extend to paired-but-disconnected devices, which is where IOBluetooth becomes necessary on macOS (since CoreBluetooth has no enumeration API for paired devices).

I'll rework this PR to just cover phase 1. Should I close this one and open a fresh PR, or force-push the reworked version here?

dlech:
> Should I close this one and open a fresh PR, or force-push the reworked version here?

I this case I'm OK with either way.

Vodur:
> > Should I close this one and open a fresh PR, or force-push the reworked version here?
> 
> I this case I'm OK with either way.

@dlech Just one thing to to be straight forward, tested on Windows and can test on macOS on Sunday, would appreciate if someone could verify the Linux implementation.

dlech:
> would appreciate if someone could verify the Linux implementation.

If there are tests, then CI should be checking it already. :smile: 

But sure, I can help with Linux if needed.

tests/test_connected_devices.py
```diff
@@ -0,0 +1,63 @@
...
+        BLEDevice("AA:BB:CC:DD:EE:02", "Keyboard K780", {}),
+        BLEDevice("AA:BB:CC:DD:EE:03", "MX Master 3S", {}),
+        BLEDevice("AA:BB:CC:DD:EE:04", None, {}),
+    ]
+
+
+@pytest.mark.asyncio
+async def test_get_connected_devices(sample_devices):
+    StubScanner._devices = sample_devices
+    devices = await BleakScanner.get_connected_devices(backend=StubScanner)
```
dlech: If we are using a stub scanner, then we aren't testing the actual backend implementation.

This should be an integration test that uses actual hardware and backend.

CHANGELOG.rst
```diff
@@ -10,6 +10,14 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 `Unreleased`_
 =============
 
+Added
+-----
+* Added ``BleakScanner.get_connected_devices()`` to retrieve BLE devices that are already connected to the system without scanning.
+
+Changed
+-------
+* ``BleakClientCoreBluetooth`` now tries ``retrievePeripheralsWithIdentifiers:`` before falling back to scanning when connecting by CoreBluetooth UUID address.
```
dlech, line 19: Since we have two separate changes, we should split that into two commits.

CHANGELOG.rst
```diff
@@ -10,6 +10,14 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 `Unreleased`_
 =============
 
+Added
+-----
+* Added ``BleakScanner.get_connected_devices()`` to retrieve BLE devices that are already connected to the system without scanning.
```
dlech: Since this doesn't actually do scanning, I'm wondering if this should be attached to something besides `BleakScanner`.

See https://github.com/hbldh/bleak/issues/1060 for a discussion on this.

bleak/backends/corebluetooth/scanner.py
```diff
@@ -178,3 +179,34 @@ def callback(
...
+    async def get_connected_devices(cls, **kwargs: Any) -> list[BLEDevice]:
+        service_uuids: Optional[list[str]] = kwargs.get("service_uuids")
+
+        manager = CentralManagerDelegate()
+        await manager.wait_until_ready()
+
+        if service_uuids:
+            cb_uuids = [CBUUID.UUIDWithString_(u) for u in service_uuids]
+        else:
+            cb_uuids = [CBUUID.UUIDWithString_("1800")]
```
dlech: It doesn't allow None?

bleak/backends/corebluetooth/scanner.py
```diff
@@ -178,3 +179,34 @@ def callback(
...
+
+        seen: set[str] = set()
+        devices: list[BLEDevice] = []
+
+        for cb_uuid in cb_uuids:
+            peripherals = (
+                manager.central_manager.retrieveConnectedPeripheralsWithServices_(
```
dlech: Why can't we just pass `cb_uuids` directly instead of calling this multiple times?

bleak/backends/bluezdbus/scanner.py
```diff
@@ -237,3 +239,32 @@ def _handle_device_removed(self, device_path: str) -> None:
...
+    @override
+    async def get_connected_devices(cls, **kwargs: Any) -> list[BLEDevice]:
+        manager = await get_global_bluez_manager()
+
+        bluez_args = kwargs.get("bluez", {})
+        adapter = bluez_args.get("adapter")
+        adapter_path = f"/org/bluez/{adapter}" if adapter else None
+
+        if adapter_path is None and manager._adapters:
+            adapter_path = next(iter(sorted(manager._adapters)))
```
dlech: IIRC, we already have a "get preferred adapter" method somewhere.

tests/test_connected_devices.py
Vodur: True, you're right.
Integration testing this is tricky since it queries OS-level connection state, which Bumble doesn't simulate. Happy to hear suggestions - should this be a manual test for now, or is there a way to inject connected device state in the test infrastructure?
I can test manually on Windows, MacOS and Linux if this will suffice.

CHANGELOG.rst
Vodur: I'll split into two commits: one for `get_connected_devices()` and one for `retrievePeripheralsWithIdentifiers:` optimization in the client.

CHANGELOG.rst
Vodur: I read #1060. I've put it on `BleakScanner` since it's the closest existing class to "device discovery", I'm happy to move it.
Would a standalone function like `bleak.get_connected_devices()` be better, or would you prefer for me to wait until there is a design for `BleakAdapter`?

bleak/backends/corebluetooth/scanner.py
Vodur: I'm not sure if nil is accepted — I defaulted to Generic Access (0x1800) to be safe. If `nil` returns all connected peripherals, I'll remove the default.
Can test this on Sunday.

bleak/backends/corebluetooth/scanner.py
Vodur: No reason to loop, will fix to pass the full array in one call.

bleak/backends/bluezdbus/scanner.py
Vodur: I found the method, I'll use `manager.get_full_adapter()` instead of accessing `_adapters` directly.

Vodur:
Tested on macOS — `None` crashes with `NSInternalInconsistencyException: Invalid parameter not satisfying: serviceUUIDs != nil`. Empty array returns nothing. Even `0x1800` (Generic Access) returned nothing for a connected MX Master 2S, while `0x180F` (Battery Service) found it. So `service_uuids` is now required on macOS and raises `BleakError` if not provided.

Vodur:
@dlech I tested on all 3 platforms, did a manual test, all had the same mouse connected, the MX Master 2S, and it repeated finding it on the OS query level and connecting to it and getting services and characteristics, only on macOS `service_uuids` was passed as a parameter:
`devices = await BleakScanner.get_connected_devices(service_uuids=["180F"])`
or
`devices = await BleakScanner.get_connected_devices(service_uuids=["180F", "180A", "FFF0"])`

dlech:
> So `service_uuids` is now required on macOS and raises `BleakError` if not provided.

I think it would be better if it was the same on all platforms (so required on all platforms). And `ValueError` would probably make more sense. `BleakError` is generally for runtime errors rather than invalid arguments.

Making the default "Generic Attribute Profile" (hex 1801) would be a good idea too as this would filter out Bluetooth classic devices (particularly on Linux).

bleak/backends/bluezdbus/manager.py
```diff
@@ -840,6 +840,30 @@ def is_paired(self, device_path: str) -> bool:
         except KeyError:
             return False
 
+    def get_connected_devices(
```
dlech, line 843: Should we pass the list of service UUIDs here so that we don't have to filter twice?

bleak/backends/corebluetooth/client.py
```diff
@@ -95,16 +95,33 @@ async def connect(self, pair: bool, **kwargs: Any) -> None:
             logger.debug("Explicit pairing is not available in CoreBluetooth.")
 
         timeout = kwargs.get("timeout", self._timeout)
+        if self._peripheral is None:
+            manager = CentralManagerDelegate()
+            await manager.wait_until_ready()
+            nsuuid = NSUUID.alloc().initWithUUIDString_(self.address)
```
dlech:
```suggestion
            nsuuid = NSUUID.UUIDWithString_(self.address)
```

bleak/backends/corebluetooth/client.py
```diff
@@ -95,16 +95,33 @@ async def connect(self, pair: bool, **kwargs: Any) -> None:
...
             )
 
             if device:
                 self._peripheral, self._central_manager_delegate = device.details
             else:
                 raise BleakDeviceNotFoundError(
-                    self.address, f"Device with address {self.address} was not found"
+                    self.address,
+                    f"Device with address {self.address} was not found",
                 )
```
dlech, line 121: These look like unrelated formatting changes.

bleak/__init__.py
```diff
@@ -417,6 +417,50 @@ async def find_device_by_filter(
             except asyncio.TimeoutError:
                 return None
 
+    @classmethod
+    async def get_connected_devices(
+        cls,
+        *,
+        service_uuids: Optional[list[str]] = None,
```
```suggestion
        service_uuids: Iterable[str]] = ["1801"],
        *,
```

As mentioned in other comment, better to be required and have a useful default.

Also doesn't need to be a keyword-only argument.

And `Iterable` makes things a bit more flexible for those using type checking.

bleak/__init__.py
```diff
@@ -417,6 +417,50 @@ async def find_device_by_filter(
...
+        backend: Optional[type[BaseBleakScanner]] = None,
+        **kwargs: Any,
```
dlech, line 440: Need to document what `kwargs` is for or drop it if it isn't needed.

bleak/__init__.py
```diff
@@ -417,6 +417,50 @@ async def find_device_by_filter(
...
+    ) -> list[BLEDevice]:
+        """Retrieve BLE devices that are currently connected to the system.
+
+        This does not require starting a scan. It queries the operating system
+        for devices that are already connected at the OS level.
+
+        The returned :class:`BLEDevice` objects can be passed directly to
+        :class:`BleakClient` for connecting without scanning.
```
```suggestion
        :class:`BleakClient` for getting the existing connection without scanning.
```

examples/connected_devices.py
```diff
@@ -0,0 +1,72 @@
...
+    print(f"Found {len(devices)} connected device(s):\n")
+    for d in devices:
+        print(f"  {d.name or '(unknown)'} [{d.address}]")
+
+    if args.connect:
+        target = devices[0]
+        print(f"\nConnecting to {target.name} [{target.address}]...")
+
+        async with BleakClient(target) as client:
+            print(f"Connected: {client.is_connected}")
```
dlech: Connected is always going to be true, so not particularly interesting to print.

examples/connected_devices.py
```diff
@@ -0,0 +1,72 @@
...
+            print(f"MTU size: {client.mtu_size}")
```
dlech: We try to discourage people from using this property since it doesn't work correctly on Linux. So probably not a good idea to use it in an example.

examples/connected_devices.py
```diff
@@ -0,0 +1,72 @@
...
+
+            print("\nServices:")
+            for service in client.services:
+                print(f"\n  [Service] {service.uuid} (Handle: {service.handle})")
+                for char in service.characteristics:
```
dlech: To keep it short, maybe just print the services?

Could add a `--details` option to optional print more.

examples/connected_devices.py
```diff
@@ -0,0 +1,72 @@
...
+        print("No connected devices found.")
+        return
+
+    print(f"Found {len(devices)} connected device(s):\n")
+    for d in devices:
+        print(f"  {d.name or '(unknown)'} [{d.address}]")
+
+    if args.connect:
+        target = devices[0]
+        print(f"\nConnecting to {target.name} [{target.address}]...")
```
dlech: It is a bit confusing to say "Connecting" here since the device is already connected to the OS. Maybe we could come up with something that better reflects what actually happens?

examples/connected_devices.py
```diff
@@ -0,0 +1,72 @@
...
+            print("\nServices:")
+            for service in client.services:
+                print(f"\n  [Service] {service.uuid} (Handle: {service.handle})")
+                for char in service.characteristics:
+                    props = ", ".join(char.properties)
+                    print(f"    [Char] {char.uuid} ({props})")
+                    for desc in char.descriptors:
+                        print(f"      [Desc] {desc.uuid}")
+
+        print("\nDisconnected.")
```
dlech: Saying "Disconnected" here is confusing here for the same reason that the device should actually stay connected to the OS AFAIU.

bleak/backends/scanner.py
```diff
@@ -277,6 +277,23 @@ async def stop(self) -> None:
         """Stop scanning for devices"""
         raise NotImplementedError()
 
+    @classmethod
+    async def get_connected_devices(cls, **kwargs: Any) -> list[BLEDevice]:
```
dlech: This should have `service_uuids` as an explicit argument.

bleak/backends/corebluetooth/scanner.py
```diff
@@ -178,3 +179,33 @@ def callback(
...
         await self._manager.stop_scan()
         self._manager.callbacks.pop(id(self), None)
+
+    @classmethod
+    @override
+    async def get_connected_devices(cls, **kwargs: Any) -> list[BLEDevice]:
+        service_uuids: Optional[list[str]] = kwargs.get("service_uuids")
+
+        if not service_uuids:
+            raise BleakError(
```
dlech: As mentioned elsewhere, this argument check can be moved out of the backend so the requirement is the same on all platforms.

bleak/backends/winrt/scanner.py
```diff
@@ -334,3 +340,26 @@ async def stop(self) -> None:
...
+        device_info_collection = await DeviceInformation.find_all_async_aqs_filter(
+            selector
+        )
+
+        devices: list[BLEDevice] = []
+        for i in range(device_info_collection.size):
+            device_info = device_info_collection.get_at(i)
+            ble_device = await BluetoothLEDevice.from_id_async(device_info.id)
+            address = _format_bdaddr(ble_device.bluetooth_address)
+            name = device_info.name if device_info.name else None
```
dlech:
```suggestion
            name = device_info.name
```

Seems redundant. Or add a comment if this is handling converting empty string to None.

bleak/backends/winrt/scanner.py
```diff
@@ -334,3 +340,26 @@ async def stop(self) -> None:
...
+            BluetoothConnectionStatus.CONNECTED
+        )
+        device_info_collection = await DeviceInformation.find_all_async_aqs_filter(
+            selector
+        )
+
+        devices: list[BLEDevice] = []
+        for i in range(device_info_collection.size):
+            device_info = device_info_collection.get_at(i)
+            ble_device = await BluetoothLEDevice.from_id_async(device_info.id)
```
dlech: Can we not get the address and name from the `device_info`?

bleak/backends/winrt/client.py
```diff
@@ -157,10 +158,13 @@ def __init__(
...
-            args = data.adv or data.scan
-            assert args
-            self._device_info = args.bluetooth_address
+            details = address_or_ble_device.details
+            if isinstance(details, RawAdvData):
+                args = details.adv or details.scan
+                assert args
+                self._device_info = args.bluetooth_address
+            else:
+                self._device_info = _address_to_int(address_or_ble_device.address)
```
dlech:
```suggestion
            self._device_info = _address_to_int(address_or_ble_device.address)
```

Seems like we could just do this and avoid the type check.

CHANGELOG.rst
dlech: I'll have to think about this some more.

We could start a new `BleakAdapter` class anyway even if this is the only method for now.

typings/Foundation/__init__.pyi
```diff
@@ -65,6 +66,9 @@ class NSArray(NSObject, Sequence[T]):
     def __getitem__(self, index: slice) -> Sequence[T]: ...
     def __len__(self) -> int: ...
     def initWithArray_(self, array: Sequence[Any]) -> Self: ...
+    @classmethod
+    def arrayWithObject_(cls, obj: Any) -> NSArray[Any]: ...
```
dlech:
```suggestion
    def arrayWithObject_(cls, obj: T) -> NSArray[T]: ...
```

Vodur:
> Can we not get the address and name from the `device_info`?

I kept `BluetoothLEDevice.from_id_async()` since `DeviceInformation` does not expose `bluetooth_address` directly without parsing the device ID string or using fragile property keys. Happy to switch if there is a preferred approach.

The WinRT AQS selector for connection status does not combine with service UUIDs in a single query. Filtering would require calling `get_gatt_services_async()` per device. Since `BluetoothLEDevice` already excludes Bluetooth Classic, I left service_uuids as a no-op on Windows (documented in the code). Let me know if you want strict filtering even at the cost of an extra async call per device.

typings/Foundation/__init__.pyi
```diff
@@ -39,7 +39,8 @@ class NSDictionary(NSObject, Mapping[_KT, _VT_co]):
 
 class NSUUID(NSObject):
     @classmethod
-    def UUIDWithString_(cls, uuidString: str) -> NSUUID: ...
+    def UUIDWithString_(cls, uuidString: str) -> Optional[NSUUID]: ...
```
dlech:
```suggestion
    def UUIDWithString_(cls, uuidString: str) -> NSUUID: ...
```

I don't think this can fail. Instead, remove the `None` check at the call site.

bleak/backends/corebluetooth/client.py
```diff
@@ -95,6 +95,20 @@ async def connect(self, pair: bool, **kwargs: Any) -> None:
             logger.debug("Explicit pairing is not available in CoreBluetooth.")
 
         timeout = kwargs.get("timeout", self._timeout)
+        if self._peripheral is None:
+            manager = CentralManagerDelegate()
+            await manager.wait_until_ready()
+            nsuuid = NSUUID.UUIDWithString_(self.address)
+            if nsuuid is not None:
+                peripherals = (
+                    manager.central_manager.retrievePeripheralsWithIdentifiers_(
```
dlech: `typings/CoreBluetooth/__init__.pyi` needs to be fixed to use `NSUUID` instead of `CBUUID`.

Vodur:
This macos-latest CI is rough haha.

dlech:
You can run the checks locally to speed things up.

```
uv tool install pyright
uv run pyright
uv tool install mypy
uv run mypy -p bleak -p tests -p examples
```

Vodur:
So regarding codecov, I'll probably need some assistance @dlech.
If it's because of integration tests, so we can't actually have any since this needs to test hardware, I did the manual tests on the same MX Master 2S which worked on all Windows, MacOS and Linux.
I also had Bluetooth earphones connected for the test, but they are a Bluetooth Classic device so it wasn't even listed, which is great.

Vodur:
> You can run the checks locally to speed things up.
> 
> 
> 
> ```
> 
> uv tool install pyright
> 
> uv run pyright
> 
> uv tool install mypy
> 
> uv run mypy -p bleak -p tests -p examples
> 
> ```

Thank you! This really helped me out, hopefully for future assistance with Bleak!

bleak/backends/winrt/scanner.py
Vodur: This code moved to `bleak/backends/winrt/adapter.py`. Kept `from_id_async` since `DeviceInformation` doesn't expose `bluetooth_address` directly without fragile property key lookups. Happy to change if you have a preferred approach.

bleak/backends/bluezdbus/adapter.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def __init__(self, manager: BlueZManager, adapter_path: str):
+        self._manager = manager
+        self._adapter_path = adapter_path
+
+    @classmethod
+    @override
+    async def default_adapter(cls, **kwargs: Any) -> Self:
+        manager = await get_global_bluez_manager()
+        adapter = kwargs.get("adapter")
```
dlech: This needs a Bluez-specific dict like `BlueZClientArgs`.

bleak/backends/bluezdbus/adapter.py
```diff
@@ -0,0 +1,47 @@
...
+class BleakAdapterBlueZDBus(BaseBleakAdapter):
+    """The native Linux Bleak BLE Adapter."""
+
+    def __init__(self, manager: BlueZManager, adapter_path: str):
+        self._manager = manager
+        self._adapter_path = adapter_path
+
+    @classmethod
+    @override
+    async def default_adapter(cls, **kwargs: Any) -> Self:
```
dlech: If we allow specifying the adapter, then it isn't really `default`. Perhaps we should pick a different name?

bleak/backends/bluezdbus/adapter.py
```diff
@@ -0,0 +1,47 @@
...
+from bleak.backends.adapter import BaseBleakAdapter
+from bleak.backends.bluezdbus.manager import BlueZManager, get_global_bluez_manager
+from bleak.backends.device import BLEDevice
+
+
+class BleakAdapterBlueZDBus(BaseBleakAdapter):
+    """The native Linux Bleak BLE Adapter."""
+
+    def __init__(self, manager: BlueZManager, adapter_path: str):
+        self._manager = manager
```
dlech: If D-Bus or BlueZ restarts, we need to get a new manager object, so caching it like this doesn't work. Just call `manager = await get_global_bluez_manager()` everywhere we need it.

bleak/backends/bluezdbus/adapter.py
```diff
@@ -0,0 +1,47 @@
...
+
+    @override
+    async def get_connected_devices(
+        self, service_uuids: Iterable[str]
+    ) -> list[BLEDevice]:
+        devices: list[BLEDevice] = []
+        for path, props in self._manager.get_connected_devices(
+            self._adapter_path, list(service_uuids)
+        ):
+            address = props.get("Address", "")
```
dlech:
```suggestion
            address = props["Address"]
```

"Address" is always present.

bleak/backends/bluezdbus/adapter.py
```diff
@@ -0,0 +1,47 @@
...
+            alias = props.get("Alias", address)
```
dlech:
```suggestion
            alias = props["Alias"]
```

"Alias" is always present.

bleak/backends/bluezdbus/adapter.py
```diff
@@ -0,0 +1,47 @@
...
+            name = None if alias == address.replace(":", "-") else alias
```
dlech: Maybe worth factoring this out into a function to share code with the same logic in `BleakScannerBlueZDBus._handle_advertising_data()`?

bleak/backends/bluezdbus/adapter.py
```diff
@@ -0,0 +1,47 @@
...
+        )
+        return cls(manager, adapter_path)
+
+    @override
+    async def get_connected_devices(
+        self, service_uuids: Iterable[str]
+    ) -> list[BLEDevice]:
+        devices: list[BLEDevice] = []
+        for path, props in self._manager.get_connected_devices(
+            self._adapter_path, list(service_uuids)
```
dlech:
```suggestion
            self._adapter_path, service_uuids
```

Should not need to turn this into a list. Just make `BlueZManager.get_connected_devices()` accept an `Iterable[str]`.

bleak/backends/bluezdbus/manager.py
```diff
@@ -840,6 +840,41 @@ def is_paired(self, device_path: str) -> bool:
...
+        self,
+        adapter_path: str,
+        service_uuids: Optional[list[str]] = None,
```
dlech:
```suggestion
        service_uuids: Iterable[str],
```

It looks like we always pass an arg, so probably don't need to make it optional.

bleak/backends/bluezdbus/manager.py
```diff
@@ -840,6 +840,41 @@ def is_paired(self, device_path: str) -> bool:
...
+        Returns:
+            A list of (device_path, device_properties) tuples.
+        """
+        uuid_set = {u.lower() for u in service_uuids} if service_uuids else None
+        result: list[tuple[str, dict[str, Any]]] = []
+        for path, interfaces in self._properties.items():
+            props = interfaces.get(defs.DEVICE_INTERFACE)
+            if props is None:
+                continue
+            if not props.get("Connected", False):
```
dlech:
```suggestion
            if not props["Connected"]:
```

Should always be present.

bleak/backends/bluezdbus/manager.py
```diff
@@ -840,6 +840,41 @@ def is_paired(self, device_path: str) -> bool:
...
+                continue
+            if props.get("Adapter") != adapter_path:
```
dlech:
```suggestion
            if props["Adapter"] != adapter_path:
```

It would make more sense to check this before checking "Connected" too.

bleak/backends/bluezdbus/manager.py
```diff
@@ -840,6 +840,41 @@ def is_paired(self, device_path: str) -> bool:
...
+                continue
+            if uuid_set is not None:
```
dlech: Ideally, this would be done at a higher level and not allow `uuid_set` to be `None` here... I think we discussed this on macOS as well for different reasons so the same logic applies everywhere.

We should at least be checking for the GATT service to ensure this is a compatible device. Otherwise it would return all Bluetooth devices, not just BLE.

bleak/backends/bluezdbus/manager.py
```diff
@@ -840,6 +840,41 @@ def is_paired(self, device_path: str) -> bool:
...
+                device_uuids = {u.lower() for u in props.get("UUIDs", [])}
+                if not (device_uuids & uuid_set):
```
dlech: What is the logic of macOS? Does it match _any_ UUID in the list or does it require _all_ UUIDs to be present?

We need to make sure this logic matches and add a comment explaining the reasoning behind it.

bleak/backends/bluezdbus/manager.py
```diff
@@ -840,6 +840,41 @@ def is_paired(self, device_path: str) -> bool:
...
+                    continue
+            result.append((path, props))
+        return result
```
dlech, line 887: For readability, I appreciate blank lines before and after any indented block (if, for, while, with, try statements).

bleak/backends/corebluetooth/adapter.py
```diff
@@ -0,0 +1,55 @@
...
+    @override
+    async def get_connected_devices(
+        self, service_uuids: Iterable[str]
+    ) -> list[BLEDevice]:
+        cb_uuids = cast(
+            NSArray[CBUUID],
+            NSArray.alloc().initWithArray_(
+                [CBUUID.UUIDWithString_(u) for u in service_uuids]
+            ),
+        )
```
dlech:
```suggestion
        cb_uuids = NSArray.arrayWithArray_(
            [CBUUID.UUIDWithString_(u) for u in service_uuids]
        )
```

It seems like we should be able to fix the type hints to make this work.

If we really can't, then...

```suggestion
        cb_uuids: NSArray[CBUUID] = NSArray.arrayWithArray_(
            [CBUUID.UUIDWithString_(u) for u in service_uuids]
        )
```

bleak/backends/corebluetooth/client.py
```diff
@@ -95,6 +95,19 @@ async def connect(self, pair: bool, **kwargs: Any) -> None:
             logger.debug("Explicit pairing is not available in CoreBluetooth.")
 
         timeout = kwargs.get("timeout", self._timeout)
+        if self._peripheral is None:
```
dlech, line 103: This could use some comments to explain why we are doing this.

bleak/backends/winrt/adapter.py
```diff
@@ -0,0 +1,53 @@
...
+from winrt.windows.devices.bluetooth import BluetoothConnectionStatus, BluetoothLEDevice
+from winrt.windows.devices.enumeration import DeviceInformation
+
+from bleak._compat import Self, override
+from bleak.backends.adapter import BaseBleakAdapter
+from bleak.backends.device import BLEDevice
+from bleak.backends.winrt.util import assert_mta
+
+
+def _format_bdaddr(a: int) -> str:
```
dlech: Let's move the existing implementation from `scanner.py` to `util.py` and use that instead of duplicating it.

bleak/backends/winrt/client.py
```diff
@@ -157,10 +158,7 @@ def __init__(
 
         # Backend specific. WinRT objects.
         if isinstance(address_or_ble_device, BLEDevice):
-            data: RawAdvData = address_or_ble_device.details
-            args = data.adv or data.scan
-            assert args
-            self._device_info = args.bluetooth_address
+            self._device_info = _address_to_int(address_or_ble_device.address)
```
dlech, line 161: This change stands on it's own, so we could split it into a separate commit and merge that already.

bleak/backends/adapter.py
```diff
@@ -0,0 +1,78 @@
...
+
+class BaseBleakAdapter:
+    """Interface for Bleak Bluetooth adapter operations.
+
+    Provides operations that do not require active scanning, such as
+    retrieving devices that are already connected to the system.
+
+    Instances should be obtained via :meth:`default_adapter`.
+
+    .. versionadded:: 3.1
```
dlech:
```suggestion
    .. versionadded:: unreleased
```

Otherwise people might be confused by the docs if we don't make a release right away.

bleak/__init__.py
```diff
@@ -418,6 +419,81 @@ async def find_device_by_filter(
...
+    async def default_adapter(
+        cls,
+        *,
+        backend: Optional[type[BaseBleakAdapter]] = None,
+        **kwargs: Any,
+    ) -> Self:
+        """Get the default Bluetooth adapter for the current platform.
+
+        Currently all platforms other than Linux only allow one Bluetooth
+        adapter. On Linux, the adapter can be selected via the ``adapter``
```
dlech: As in another comment, `adapter` needs to be encapsulated in a backend-specific dict.

bleak/__init__.py
```diff
@@ -418,6 +419,81 @@ async def find_device_by_filter(
...
+                devices and excludes Bluetooth Classic devices.
+
+        Returns:
+            A list of :class:`BLEDevice` objects representing connected devices.
+
+        Raises:
+            ValueError: if ``service_uuids`` is empty.
+        """
+        service_uuids_list = [normalize_uuid_str(u) for u in service_uuids]
+        if not service_uuids_list:
```
dlech, line 499: Would be more logical to do the check for empty first.

typings/CoreBluetooth/__init__.pyi
```diff
@@ -52,7 +52,7 @@ class CBCentralManager(CBManager):
         self, serviceUUIDs: NSArray[CBUUID]
     ) -> NSArray[CBPeripheral]: ...
     def retrievePeripheralsWithIdentifiers_(
-        self, serviceUUIDs: NSArray[CBUUID]
+        self, identifiers: NSArray[NSUUID]
```
dlech, line 55: This is another one that can be a separate commit and be merged already.

typings/Foundation/__init__.pyi
```diff
@@ -40,6 +40,7 @@ class NSDictionary(NSObject, Mapping[_KT, _VT_co]):
 class NSUUID(NSObject):
     @classmethod
     def UUIDWithString_(cls, uuidString: str) -> NSUUID: ...
+    def initWithUUIDString_(self, uuidString: str) -> Optional[Self]: ...
```
dlech: Would prefer to just use `UUIDString` and not add `initWithUUIDString_`.

bleak/backends/bluezdbus/adapter.py
Vodur: Added `BlueZAdapterArgs TypedDict` in `bleak/args/bluez.py` and replaced `kwargs.get("adapter")` with explicit `bluez: BlueZAdapterArgs = {}` parameter. Mirrors the `BlueZClientArgs` pattern. The `BleakAdapter` facade in `bleak/__init__.py` also takes `bluez: BlueZAdapterArgs = {}` and passes it through.

bleak/backends/bluezdbus/adapter.py
Vodur: Renamed to `get()` everywhere - `BaseBleakAdapter.get()`, `BleakAdapterBlueZDBus.get()`, `BleakAdapterCoreBluetooth.get()`, `BleakAdapterWinRT.get()`, and the public `BleakAdapter.get()` facade. Usage is now `adapter = await BleakAdapter.get()` or `BleakAdapter.get(bluez={"adapter": "hci1"})`.

bleak/backends/bluezdbus/adapter.py
Vodur: Done - Removed manager from `__init__`, the class now only stores `adapter_path`. Both `get()` and `get_connected_devices()` call `await get_global_bluez_manager()` independently so a restart of D-Bus/BlueZ gets the fresh manager.

bleak/backends/bluezdbus/adapter.py
Vodur: Using `props["Address"]` directly.

bleak/backends/bluezdbus/adapter.py
Vodur: Using `props["Alias"]` directly.

bleak/backends/bluezdbus/adapter.py
Vodur: Extracted `device_name_from_props()` helper to `bleak/backends/bluezdbus/utils.py`. Both the new adapter and the existing `BleakScannerBlueZDBus._handle_advertising_data()` now use it.

bleak/backends/bluezdbus/adapter.py
Vodur: `BlueZManager.get_connected_devices()` now accepts `Iterable[str]` (required, not optional). Dropped the `list()` cast in the adapter. Made it required since the only caller is the adapter, which always passes a non-empty value.

bleak/backends/bluezdbus/manager.py
Vodur: Made `service_uuids` required (`Iterable[str]`). Since `BleakAdapter` always validates non-empty in the facade, optional-with-fallback was redundant.

bleak/backends/bluezdbus/manager.py
Vodur: `props["Connected"]` direct access.

bleak/backends/bluezdbus/manager.py
Vodur: `props["Adapter"]` direct access, and moved the adapter check before the connected check.

bleak/backends/bluezdbus/manager.py
Vodur: Addressed in the previous comment by making `service_uuids` required.

For the BLE-only filter: added `if path not in self._service_map: continue`. BlueZ only adds entries to `_service_map` when it exposes `org.bluez.GattService1` interfaces, which only happens for BLE devices after service discovery. Classic-only devices are excluded.

bleak/backends/bluezdbus/manager.py
Vodur: Logic matches, both use "any of" semantics. macOS `retrieveConnectedPeripheralsWithServices:` per Apple docs returns peripherals "containing any of the specified services". Linux uses set intersection (`device_uuids & uuid_set`), which is non-empty if at least one service matches. Updated the docstring to call this out explicitly and reference the macOS API.

bleak/backends/bluezdbus/manager.py
Vodur: Applied the blank-lines-around-blocks style consistently across all files in #1966.

bleak/backends/corebluetooth/adapter.py
Vodur: Added `arrayWithArray_` to the Foundation typings as a generic class method (same `_U` TypeVar pattern as `arrayWithObject_`). The adapter now uses `NSArray.arrayWithArray_(...)` directly without `cast`, and pyright correctly infers `NSArray[CBUUID]`.

bleak/backends/winrt/adapter.py
Vodur: Moved `_format_bdaddr` from `bleak/backends/winrt/scanner.py` to `bleak/backends/winrt/util.py` as `format_bdaddr` (public, since it's now shared). Both the scanner and adapter import from there.
No duplication.

bleak/backends/winrt/client.py
Vodur: Split into a separate commit. The WinRT `__init__` simplification is now its own commit at the bottom of the PR (independent of the `BleakAdapter` feature).

bleak/backends/adapter.py
Vodur: Done - Replaced all `.. versionadded:: 3.1` with `.. versionadded:: unreleased`.

bleak/__init__.py
Vodur: Already addressed with `BlueZAdapterArgs`.

bleak/__init__.py
Vodur: Check happens first now: convert iterable to list, raise `ValueError` if empty, then normalize.

typings/CoreBluetooth/__init__.pyi
Vodur: Split into a separate commit `typings/corebluetooth: fix retrievePeripheralsWithIdentifiers parameter type`. Sits before the macOS commit so it can be merged independently.

typings/Foundation/__init__.pyi
Vodur: Removed `initWithUUIDString_` from the Foundation typings. Wasn't used anywhere.

bleak/backends/bluezdbus/manager.py
```diff
@@ -839,6 +839,54 @@ def is_paired(self, device_path: str) -> bool:
...
+        self,
+        adapter_path: str,
+        service_uuids: Iterable[str],
+    ) -> list[tuple[str, dict[str, Any]]]:
+        """
+        Gets D-Bus object paths and properties for all connected BLE devices
+        on an adapter that advertise at least one of the given service UUIDs.
+
+        The "any of" matching semantics match the macOS
+        ``retrieveConnectedPeripheralsWithServices:`` API.
```
dlech: This comment would make more sense right above the `if not (device_uuids & uuid_set):` line.

bleak/backends/bluezdbus/manager.py
```diff
@@ -839,6 +839,54 @@ def is_paired(self, device_path: str) -> bool:
...
+                matches if it advertises at least one of these services.
+
+        Returns:
+            A list of (device_path, device_properties) tuples.
+        """
+        uuid_set = {u.lower() for u in service_uuids}
+        result: list[tuple[str, Device1]] = []
+
+        for path, interfaces in self._properties.items():
+            props = interfaces.get(defs.DEVICE_INTERFACE)
```
dlech:
```suggestion
            props: Device1 | None = interfaces.get(defs.DEVICE_INTERFACE)
```

dlech (changes requested):

bleak/__init__.py
```diff
@@ -418,6 +424,87 @@ async def find_device_by_filter(
...
+        for devices that are already connected at the OS level.
+
+        The returned :class:`BLEDevice` objects can be passed directly to
+        :class:`BleakClient` for getting the existing connection without scanning.
+
+        Args:
+            service_uuids:
+                Service UUIDs to filter on. Defaults to the Generic
+                Attribute Profile (``0x1801``), which is mandatory for BLE
+                devices and excludes Bluetooth Classic devices.
```
dlech: Technically, one can have GATT on Bluetooth classic. Not sure if mentioning that would make things more confusing or not.

Vodur: I did mention, I softened it up a bit, where dropping the inaccurate Classic-exclusion claim while still explaining why this default makes sense.

bleak/backends/bluezdbus/manager.py
```diff
@@ -839,6 +839,53 @@ def is_paired(self, device_path: str) -> bool:
...
+                matches if it advertises at least one of these services.
+
+        Returns:
+            A list of (device_path, device_properties) tuples.
+        """
+        uuid_set = {u.lower() for u in service_uuids}
+        result: list[tuple[str, Device1]] = []
+
+        for path, interfaces in self._properties.items():
+            props: Device1 | None = interfaces.get(defs.DEVICE_INTERFACE)
```
dlech:
```suggestion
            props = cast(Device1 | None, interfaces.get(defs.DEVICE_INTERFACE))
```

I guess it has to be this way.

bleak/backends/bluezdbus/manager.py
Vodur: Locally running pyright raises no errors, interesting...

Vodur: I can try:
`cast(Device1, raw_props)`, where `raw_props = interfaces.get(defs.DEVICE_INTERFACE)`.

dlech:
FYI, a lot of the changes got squashed into the wrong commit.

Vodur:
> FYI, a lot of the changes got squashed into the wrong commit.

You're right, I'll rearrange and push again.

bleak/backends/corebluetooth/client.py
```diff
@@ -95,6 +95,26 @@ async def connect(self, pair: bool, **kwargs: Any) -> None:
...
+            nsuuid = NSUUID.UUIDWithString_(self.address)
+            peripherals = (
+                new_manager.central_manager.retrievePeripheralsWithIdentifiers_(
+                    NSArray.arrayWithObject_(nsuuid)
+                )
+            )
+
+            if peripherals and len(peripherals) > 0:
+                self._peripheral = peripherals.objectAtIndex_(0)
+                self._central_manager_delegate = new_manager
```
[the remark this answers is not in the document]
dlech: I don't think it is possible to hit this condition.

bleak/backends/bluezdbus/manager.py
```diff
@@ -839,6 +839,53 @@ def is_paired(self, device_path: str) -> bool:
...
+
+            if props["Adapter"] != adapter_path:
+                continue
+
+            if not props["Connected"]:
+                continue
+
+            if path not in self._service_map:
+                continue
+
```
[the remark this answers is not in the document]
dlech: We are looking for already connected devices, so I think keeping this check makes sense. Although, I suppose it probably is a bit redundant as the other checks should catch all cases.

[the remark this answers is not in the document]
Vodur: `_service_map` not actually redundant - it's the BLE-only filter. The other checks don't rule out Classic Bluetooth.
`path in self._service_map` is the only thing that distinguishes BLE from Classic - `org.bluez.GattService1` interfaces are exposed only for BLE devices that have completed service discovery.

bleak/backends/adapter.py
```diff
@@ -0,0 +1,78 @@
...
+
+            return (BleakAdapterCoreBluetooth, backend)  # type: ignore
+
+        case BleakBackend.WIN_RT:
+            from bleak.backends.winrt.adapter import BleakAdapterWinRT  # type: ignore
+
+            return (BleakAdapterWinRT, backend)  # type: ignore
+
+        case _:
+            raise BleakError(f"Unsupported backend: {backend}")
```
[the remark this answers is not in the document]
Vodur: Done - extended `tests/test_platform_detection.py` to also cover `get_platform_adapter_backend_type()` for all three platforms (BlueZDBus, CoreBluetooth, WinRT). Mirrors the existing tests for client and scanner.

bleak/backends/bluezdbus/manager.py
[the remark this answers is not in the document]
dlech: If we always search for at least one UUID though, that UUID should be a GATT service and have the same effect. 

Anyway, I'm OK to still leave in the check.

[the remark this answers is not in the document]
Vodur: Fair point, agreed that for proper GATT service UUIDs the intersection check would catch it too.
Keeping the `_service_map` check as a defensive filter for cases where the user might pass a UUID that overlaps with a Classic profile.

Anything else outstanding before merge?

dlech:
> Anything else outstanding before merge?

I would like to do a bug fix release with #1973 first before merging any major new features like this.

Then we can consider to merge this even though there is sill more to be done. 

Before a release including this feature though, we need to:
- Update BlueZ backend to connect to already connected devices and leave them connected when "disconnecting". This will require Bleak to add a flag to keep track of this since BlueZ doesn't do reference counting of requested connections. https://github.com/bluez/bluez/issues/89
- Add integration tests and do lots of testing to make sure things are working the same on all platforms.
- Consider if we want `BleakAdapter.get()` to act like a singleton. (I know from experience that a lot of people will write `BleakAdapter.get()` a lot of times instead of storing the result in a variable.)

Vodur:
> > Anything else outstanding before merge?
> 
> I would like to do a bug fix release with #1973 first before merging any major new features like this.
> 
> Then we can consider to merge this even though there is sill more to be done.
> 
> Before a release including this feature though, we need to:
> 
> * Update BlueZ backend to connect to already connected devices and leave them connected when "disconnecting". This will require Bleak to add a flag to keep track of this since BlueZ doesn't do reference counting of requested connections. [Device connection reference counting bluez/bluez#89](https://github.com/bluez/bluez/issues/89)
> * Add integration tests and do lots of testing to make sure things are working the same on all platforms.
> * Consider if we want `BleakAdapter.get()` to act like a singleton. (I know from experience that a lot of people will write `BleakAdapter.get()` a lot of times instead of storing the result in a variable.)

Sounds good - happy to wait for #1973.

On the three follow-up items:
1. BlueZ keep-connected - I'll look into [Device connection reference counting bluez/bluez#89](https://github.com/bluez/bluez/issues/89). Sounds like we need a flag on `BleakClientBlueZDBus` like `_owns_connection` (true if we initiated, false if we got the device from `get_connected_devices()`), so `disconnect()` only calls BlueZ's `Disconnect()` when we own it. Want this in this PR or a follow-up?
2. Integration tests - happy to manually test all three platforms again with real devices once you have a sense of the test scenarios you'd want covered. If you want automated tests, I'd appreciate guidance - Bumble doesn't simulate OS-level connection state, so I'm not sure how to script "device is already connected" in CI.
3. Singleton for `BleakAdapter.get()` - I think this is a good idea, but per-adapter rather than global, since on Linux `bluez={"adapter": "hci1"}` should give a different instance than `hci0`. So a cache keyed on `(backend, adapter_path)`. Subsequent `await BleakAdapter.get(...)` returns the existing instance. Want me to add this in this PR or a follow-up?

dlech:
> 2\. Bumble doesn't simulate OS-level connection state,

We only use Bumble to simulate the peripheral while still using the OS Bluetooth stack. So I don't think this would be too difficult.

```python
async with BleakClient(...) as outer_client:
    # We don't actually want to actually support nested clients
    # like this in Bleak, but for the purposes of testing an
    # already connected device, this is the simplest way...

    adapter = await BleakAdapter.get()
    devices = await adapter.get_connected_devices()
    match = next(d for d in devices if device...)

    async with BleakClient(match) as inner_client:
        # Not sure if we need to do anything much here...
        await asyncio.sleep(1)


    assert outer_client.connected, "inner client disconnecting should not disconnect outer client"
```

dlech:
> 3\. I think this is a good idea, but per-adapter rather than global,

Yes, on Linux, per adapter, on other OSes, global.

Also needs to be per-event loop. Some backends may capture the running event loop, so we would need a new instance in that case as well.

dlech:
I would prefer splitting things out into separate PRs so that this doesn't get too big.

Vodur:
Thanks for the clarifications!

**Singleton design** - got it: per-(adapter, event-loop) cache on Linux, per-event-loop cache elsewhere. So the cache key is `(PlatformBleakAdapter, adapter_path_or_None, loop)`. Two `await BleakAdapter.get()` calls from the same event loop return the same instance.

Regarding scope, agreed, splitting into separate PRs keeps this one focused. So this PR stays as-is.

- Integration test (using your nested-BleakClient pattern with Bumble - that's much simpler than I thought, thanks).

- Singleton caching for `BleakAdapter.get()`.

- BlueZ keep-connected behavior (bluez/bluez#89)

Also pushed `docs/api/adapter.rst` and added it to the docs toctree, plus added myself to `AUTHORS.rst` (per `CONTRIBUTING.rst`).

bleak/__init__.py
JPHutchins: It seems like the bluez param could be renamed to "adapter_args" where "adapter_args" is a type union of adapter-specific args collections, replacing the need for type erased kwargs.

bleak/backends/adapter.py
JPHutchins, line 78: Possible to assert_never instead for a compile time instead of run time error?

JPHutchins (approved):
Great work, very exciting to get this feature!

bleak/__init__.py
Vodur: A union arg here would be an outlier. But you're right that `**kwargs: Any` is type-erased and unnecessary - only `bluez=` is forwarded.

bleak/backends/adapter.py
Vodur: Good idea on `assert_never` - but I held off applying it here because `bleak/backends/client.py` and `bleak/backends/scanner.py` use the same `case _: raise BleakError(...)` pattern, and switching just `adapter.py` to `assert_never` would make this PR diverge from the established convention. Worth doing across all three for consistency.

@dlech preference? Apply `assert_never` everywhere as a follow-up PR afterthis, or fold it into this PR while we're here?

Vodur:
Added singleton caching to `BleakAdapter.get()`. Repeated calls from the same event loop with the same `bluez` adapter selection return the same `BleakAdapter` instance. The cache uses a `WeakKeyDictionary` keyed on the event loop, so entries are released when the loop is GC'd.
Concurrent calls converge on a single instance via `dict.setdefault`.
Six unit tests added in `tests/test_adapter_singleton.py`.

bleak/backends/adapter.py
dlech: There is more than enough going on here already. I would prefer to consider it separately.

dlech:
> the cache uses a `WeakKeyDictionary` keyed on the event loop, so entries are released when the loop is GC'd.

This probably doesn't work at least for some backends since the value (adapter instance) can have a reference to the loop. See https://github.com/hbldh/bleak/commit/1be914cca671ae4b44efd3848af4e4aee96c52b6 where we ran into this before.



> Concurrent calls converge on a single instance via `dict.setdefault`.

This also unnecessarily creates a new instance when it is not needed. I would not do that. Since any thing between awaits is atomic in asyncio anyway, not using multiple statements instead of `setdefault()` will converge just the same.

Vodur:
Both points are right - 1be914c shows the exact pattern (dict + lazy prune of closed loops) and the cycle problem with `WeakKeyDictionary`. And yeah, `setdefault` is wasted work given asyncio atomicity. I'll switch this PR to track #1979 and post any follow-ups (singleton with the closed-loop pattern, adapter state property for #320) as separate PRs after that lands. Let me know if you'd prefer I close this in favor of #1979.

dlech:
Normally, I would say keep this PR open, but the discussion has gone in so many directions, it will probably better to start fresh with a new PR for each featured added on top of #1979 (which I just merged).

PR #1973, opened 2026-04-25
backends/bluez: change default to use StartNotify

dlech:
In commit 9b559c52 ("backends/bluezdbus: allow forcing StartNotify over AcquireNotify"), we change the default behavior for the BlueZ backend to use "AcquireNotify" instead of "StartNotify". In practice, this caused issues with a number of devices, so we are changing the default back to "StartNotify". The parameter still exists to allow users to use "AcquireNotify" if needed, but it is now opt-in instead of the default, so users will need to update their code if they want to use "AcquireNotify" from now on.

It also turns out there was an additional bug that if you did opt into using "StartNotify" and tried to call `client.stop_notify()`, it would silently do nothing and not actually stop notifications. A second commit has been added to fix that.

FYI: @lkempf, @patman15, @Raupinger

Closes: https://github.com/hbldh/bleak/issues/1951

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,10 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 `Unreleased`_
 =============
 
+Changed
+-------
+* Changed default value of ``BlueZNotifyArgs.use_start_notify`` to ``False``. Fixes #1951.
+
 `3.0.1`_ (2026-03-25)
 =====================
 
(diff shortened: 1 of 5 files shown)


patman15:
@dlech 
From the release notes:
> * Changed default value of ``BlueZNotifyArgs.use_start_notify`` to ``False``. Fixes #1951.

Isn't the result of this change exactly the opposite? The default is now `True` ?

dlech:
Oops, yes, the typo should be fixed now.

PR #1974, opened 2026-04-27
backends: set BluetoothLEAdvertisementWatcher.AllowExtendedAdvertisements  when availalble only

martinschweingruber:
Set allow_extended_advertisements when available only. Older windows version before 19041 do not support the option. In this case Bleak fails with "property is not available in this version of Windows".

https://learn.microsoft.com/en-us/uwp/api/windows.devices.bluetooth.advertisement.bluetoothleadvertisementwatcher.allowextendedadvertisements

### DIFF
--- a/bleak/backends/winrt/scanner.py
+++ b/bleak/backends/winrt/scanner.py
@@ -269,7 +269,12 @@ class BleakScannerWinRT(BaseBleakScanner):
         self.watcher = BluetoothLEAdvertisementWatcher()
         self.watcher.scanning_mode = self._scanning_mode
         # BlueZ and CoreBluetooth don't allow controlling this and always enabled it, so do the same here
-        self.watcher.allow_extended_advertisements = True
+        try:
+            self.watcher.allow_extended_advertisements = True
+        except AttributeError:
+            logger.warning(
+                "Extended advertisements are not available in this OS Version."
+            )
 
         event_loop = asyncio.get_running_loop()
         self._stopped_event = asyncio.Event()


bleak/backends/winrt/scanner.py
```diff
@@ -269,7 +269,11 @@ async def start(self) -> None:
         self.watcher = BluetoothLEAdvertisementWatcher()
         self.watcher.scanning_mode = self._scanning_mode
         # BlueZ and CoreBluetooth don't allow controlling this and always enabled it, so do the same here
-        self.watcher.allow_extended_advertisements = True
+        try:
+            self.watcher.allow_extended_advertisements = True
+        except AttributeError as e:
+            if "property is not available in this version of Windows" not in str(e):
+                raise
```
dlech: Checking the string is not a good idea. Windows can translate error messages. I don't think it could be caused by anything else anyway.

Instead, I would unconditionally log a warning here that explains what is going on.

CHANGELOG.rst
```diff
@@ -16,6 +16,7 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 Fixed
 -----
 * Fixed ``AttributeError`` in ``start_notify()`` and ``stop_notify()`` on Android. Fixes #1834.
+* Fixed ``AttributeError`` on Windows older than 19041.
```
dlech: This version of Windows 10 has been out of support for almost 5 years. I don't really care to keep supporting versions that old.

If there is a really good reason that someone is still running this version and really needs the latest version of Bleak, then I will take the patch, but I would leave it out of the changelog because I don't want to advertise that we actually support unsupported software.

bleak/backends/winrt/scanner.py
martinschweingruber: thx, for that hint

CHANGELOG.rst
martinschweingruber: Haha, understand that.
I have some restrictions in medical context, but can easily patch it myself!

dlech: I consider that a reasonably good reason to consider accepting the change. :smile:

PR #1975, opened 2026-04-29
backends/bluez: don't disconnect devices we did not connect

Vodur:
## Summary                                                                                                                                                                                 
   
When a `BleakClient` is constructed for a device that is already connected (e.g. connected by another process), `connect()` already correctly skips the BlueZ `Connect` call. However, `disconnect()` would still tear down the connection, since BlueZ does not do reference counting for connection requests.

This adds an `_owns_connection` flag to `BleakClientBlueZDBus` so that `disconnect()` only calls BlueZ `Disconnect` when this instance actually initiated the connection. Local state (`_cleanup_all()`) is still cleared either way.

Discussed in #1966 and tracked in https://github.com/bluez/bluez/issues/89.
FYI: @dlech

## Test plan
- [ ] Manual test on Linux: outer `BleakClient` connects, inner `BleakClient` "connects" to same device, inner exits, outer is still connected, outer exits, BlueZ disconnects.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,10 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 `Unreleased`_
 =============
 
+Fixed
+-----
+* ``BleakClientBlueZDBus.disconnect()`` no longer disconnects from the device after "connecting" to an already connected device. Merged #1975.
+
 `3.0.2`_ (2026-05-02)
 =====================
 
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -102,6 +102,12 @@ class BleakClientBlueZDBus(BaseBleakClient):
         # used to override mtu_size property
         self._mtu_size: Optional[int] = None
 
+        # True if this BleakClient instance initiated the BlueZ connection.
+        # When False, the device was already connected at the time connect()
+        # was called, so disconnect() will not call BlueZ Disconnect.
+        # See https://github.com/bluez/bluez/issues/89
+        self._owns_connection = False
+
     # Connectivity methods
 
     @override
@@ -210,7 +216,10 @@ class BleakClientBlueZDBus(BaseBleakClient):
                         # if connection was successful but _get_services() raises (e.g.
                         # because task was cancelled), then we still need to disconnect
                         # before passing on the exception.
-                        if self._bus:
+                        # If the device was already connected when connect() was called,
+                        # this BleakClient does not own the connection and must not
+                        # tear it down.
+                        if self._bus and self._owns_connection:
                             # If disconnected callback already fired, this will be a no-op
                             # since self._bus will be None and the _cleanup_all call will
                             # have already disconnected.
@@ -250,8 +259,10 @@ class BleakClientBlueZDBus(BaseBleakClient):
                             'skipping calling "Connect" since %s is already connected',
                             self._device_path,
                         )
+                        self._owns_connection = False
                     else:
                         logger.debug("Connecting to BlueZ path %s", self._device_path)
+                        self._owns_connection = True
 
                         # Calling pair will fail if we are already paired, so
                         # in that case we just call Connect.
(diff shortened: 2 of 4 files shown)


dlech:
This needs code coverage.

Vodur:
Handled it.
The remaining partial is on `if self._bus and self._owns_connection:` in the connect-time error cleanup. Its False branch only fires if `_get_services()` raises after the inner client finds the device already connected - I didn't see a clean way to exercise that without mocking.

tests/integration/test_client_connecting.py
```diff
@@ -63,6 +64,31 @@ async def test_is_connected(bumble_peripheral: Device):
...
+
+    async with BleakClient(device) as outer_client:
+        # Nested BleakClients are not a supported usage pattern, but it
+        # is the simplest way to exercise the already-connected code path.
+        async with BleakClient(device) as inner_client:
+            assert inner_client.is_connected
+
+        assert (
+            outer_client.is_connected
+        ), "inner client disconnect should not disconnect outer client"
```
dlech, line 93:
```suggestion
    assert (
        outer_client.is_connected
    ), "inner client disconnect should not disconnect outer client"
```

Otherwise it is not a valid check because the inner context manager has not exited yet.

tests/integration/test_client_connecting.py
```diff
@@ -63,6 +64,31 @@ async def test_is_connected(bumble_peripheral: Device):
...
+    """
+    await configure_and_power_on_bumble_peripheral(bumble_peripheral)
+
+    device = await find_ble_device(bumble_peripheral)
+
+    async with BleakClient(device) as outer_client:
+        # Nested BleakClients are not a supported usage pattern, but it
+        # is the simplest way to exercise the already-connected code path.
+        async with BleakClient(device) as inner_client:
+            assert inner_client.is_connected
```
dlech, line 89:
```suggestion
            # If this is false, it is a bug in Bleak. Real usage never needs
            # to check `is_connected` after connecting because it cannot be
            # anything but true. If connecting fails, an exception is raise
            # before we get here.
            assert inner_client.is_connected
```

There were some early Bleak examples that had a check like this and people have been copying it ever since and it annoys me whenever I see it. 😛

CHANGELOG.rst
```diff
@@ -10,6 +10,10 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 `Unreleased`_
 =============
 
+Fixed
+-----
+* ``BleakClientBlueZDBus.disconnect()`` no longer disconnects from the device when the connection was not initiated by this :class:`BleakClient` instance (i.e. the device was already connected at the time :meth:`connect` was called). See https://github.com/bluez/bluez/issues/89.
```
dlech:
```suggestion
* ``BleakClientBlueZDBus.disconnect()`` no longer disconnects from the device after "connecting" to an already connected device. Merged #1975.
```

I like to try to keep these as brief as possible.

bleak/backends/bluezdbus/client.py
dlech, line 254: This brings up an interesting point I had not considered. If the device is already connected because an app crashed, then this change in behavior will leave a device connected when it was not intended to.

We might need to rethink what `_owns_connection` means. That is, we might only want to set the flag when the user looked up the device from the list of already connected devices.

Otherwise, this is going to be a breaking change. And we would need to add a BlueZ-specific flag to force disconnection to be able to get the previous behavior.

tests/integration/test_client_connecting.py
Vodur: I left the `outer_client.is_connected` assert at the original 8-space indent (inside the outer `async with`). De-indenting to 4 spaces puts it after the outer client has also disconnected, so the assert would always fail.

dlech: Oh right. I guess I am a bit cross-eyed this morning.

bleak/backends/bluezdbus/client.py
Vodur: You know, I really thought that #1966 should be base for this, would be easier to actually "own" connected device.
I agree the trigger should be "user explicitly opted into an existing connection", not "device happened to be already connected". I'd like to gate `_owns_connection = False` on a marker carried via `BLEDevice.details` (set by `BleakAdapter.get_connected_devices()`) - but that means this PR depends on #1966.

Want me to hold #1975 until #1966 lands, or would you prefer a different opt-in mechanism (e.g., a constructor flag) to keep the PRs independent?

dlech:
Did you mean to close this one? We still need this in some form.

Vodur:
> Did you mean to close this one? We still need this in some form.

I wanted to form this anew, since basing it on `BleakAdapter` would be better as we discussed.

dlech:
OK, I thought it would still be 80% the same code.

PR #1976, opened 2026-05-02
Add poethepoet config

timrid:
This is an alternative to https://github.com/hbldh/bleak/pull/1922 that uses `poethepoet` instead of `tox` as [you suggested](https://github.com/hbldh/bleak/pull/1922#pullrequestreview-3688851735).

Some examples of the new commands:

- `uv run poe test-py310`: Run pytest on Python 3.10
- `uv run poe test-py310 --bleak-hci-transport=serial:/dev/tty.usbmodem1101`: Run integration tests on Python 3.10
- `uv run poe test-all`: Run pytest on all Python versions
- `uv run poe test-all --bleak-hci-transport=serial:/dev/tty.usbmodem1101`: Run integration tests on all Python versions
- `uv run poe lint`: Run code linters (isort, black, flake8)
- `uv run poe lint --checkonly`: Run code linters (isort, black, flake8) without changing anything (mainly for CI usage)
- `uv run poe typecheck`: Run type checkers (mypy, pyright)
- `uv run poe docs`: Building the docs 

This is only the groundwork for all current functions of bleak. I plan to expand the poethepoet config in https://github.com/hbldh/bleak/pull/1944 to add "cross-platform" type checking for the new Android backend. And additionaly in https://github.com/hbldh/bleak/pull/1853 to add type checking for multiple backends (`pyobjc` and `rubicon-objc`).

It took some time to find a good way to pass the `--bleak-hci-transport` or `--bleak-bluez-vhci` options to the pytests. Finally I created https://github.com/nat-n/poethepoet/pull/380 that simplifies this problem an which is now released in poethepoeth [v0.45.0](https://github.com/nat-n/poethepoet/releases/tag/v0.45.0).

### DIFF
--- a/.github/alpine-vm/start-alpine-vm.sh
+++ b/.github/alpine-vm/start-alpine-vm.sh
@@ -21,7 +21,7 @@ VM_BUILD_DIR="$REPO_ROOT/.alpine-vm-build"
 VM_IMAGE="$VM_BUILD_DIR/alpine.qcow2"
 VM_SSH_PORT="2222"
 VM_CONSOLE_LOG="vm-console.log"
-VM_PACKAGES="bluez openssh uv"
+VM_PACKAGES="bluez openssh uv git"
 
 # Colors for output
 GREEN='\033[0;32m'
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -17,7 +17,9 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
+        python-version: ["310", "311", "312", "313", "314"]
+    env:
+      FORCE_COLOR: "1"
     steps:
       - uses: actions/checkout@v4
 
@@ -28,7 +30,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Test with pytest
-        run: uv run pytest -v --cov-report=xml --junitxml=junit.xml -o junit_family=legacy
+        run: uv run poe test-py${{ matrix.python-version }} -v --cov-report=xml --junitxml=junit.xml -o junit_family=legacy
 
       - name: Upload coverage reports to Codecov
         uses: codecov/codecov-action@v5
@@ -45,23 +47,13 @@ jobs:
           flags: ${{ matrix.os }}-py${{ matrix.python-version }}
           token: ${{ secrets.CODECOV_TOKEN }}
 
-      # We don't do type checking in the lint job because we have conditionals
-      # on both the platform and the Python version in the code, so we
-      # need to check each matrix combination.
-      #
-      # We want always the latest version of pyright and mypy, to catch
-      # any breakage as soon as we can. That is why they are not dev
-      # dependencies in the pyproject.toml, because that would lock them to
-      # a specific version. Instead we install the latest version as tool.
-      - name: Type checking (pyright)
-        run: |
-          uv tool install pyright
-          uv run pyright
+      - name: Type checking
+        # We don't do this in the lint job because we have conditionals
+        # on both the platform and the Python version in the code, so we
+        # need to check each matrix combination.
 
-      - name: Type checking (mypy)
         run: |
-          uv tool install mypy
-          uv run mypy -p bleak -p tests -p examples
+          uv run poe typecheck
 
   integration_tests_bluez:
     name: "BlueZ integration tests"
@@ -69,7 +61,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
+        python-version: ["310", "311", "312", "313", "314"]
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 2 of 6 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -28,7 +30,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Test with pytest
-        run: uv run pytest -v --cov-report=xml --junitxml=junit.xml -o junit_family=legacy
+        run: uv run poe test-py${{ matrix.python-version }} -v --cov-report=xml --junitxml=junit.xml -o junit_family=legacy
```
dlech: Will this make it more verbose since the commands in `pyproject.toml` also include `-v`?

timrid: You are right. The `-v` is doubled. I removed the `-v` in the .yml.

.github/workflows/build_and_test.yml
```diff
@@ -91,7 +83,7 @@ jobs:
             export FORCE_COLOR=1
             export CI=true
             export GITHUB_ACTIONS=true
-            uv run --python ${{ matrix.python-version }} pytest -v tests --bleak-bluez-vhci -v --cov-report=xml --cov-report=html:../.htmlcov --junitxml=junit.xml -o junit_family=legacy -o cache_dir=../.pytest_cache
+            uv run poe test-py${{ matrix.python-version }} tests --bleak-bluez-vhci --cov-report=xml --cov-report=html:../.htmlcov --junitxml=junit.xml -o junit_family=legacy -o cache_dir=../.pytest_cache
```
dlech:
```suggestion
            uv run poe test-py${{ matrix.python-version }} --bleak-bluez-vhci --cov-report=xml --cov-report=html:../.htmlcov --junitxml=junit.xml -o junit_family=legacy -o cache_dir=../.pytest_cache
```

Isn't `tests` redundant too?

pyproject.toml
```diff
@@ -100,7 +102,76 @@ filterwarnings = [
...
+# a specific version. Instead we install the latest version as tool.
+sequence = [
+    { cmd = "mypy", executor = { with = "mypy" } },
+    { cmd = "pyright", executor = { with = "pyright" } },
+]
+
+[tool.poe.tasks.lint]
+help = "Run code linters (isort, black, flake8)"
+sequence = [
+    { cmd = "isort . docs ${checkonly:+--check --diff}", executor = { group = "lint" } },
```
dlech: This does not work the same as the shell expansion. It does not sort `docs`.

We actually have to run `isort` twice, once with `.` and once with `docs`.

dlech:
```suggestion
    { cmd = "isort . ${checkonly:+--check --diff}", executor = { group = "lint" } },
    { cmd = "isort docs ${checkonly:+--check --diff}", executor = { group = "lint" } },
```

.github/alpine-vm/start-alpine-vm.sh
```diff
@@ -21,7 +21,7 @@ VM_BUILD_DIR="$REPO_ROOT/.alpine-vm-build"
 VM_IMAGE="$VM_BUILD_DIR/alpine.qcow2"
 VM_SSH_PORT="2222"
 VM_CONSOLE_LOG="vm-console.log"
-VM_PACKAGES="bluez openssh uv"
+VM_PACKAGES="bluez openssh uv git"
```
dlech, line 24: Does `poe` invoke `git`?

pyproject.toml
```diff
@@ -100,7 +102,77 @@ filterwarnings = [
...
+[tool.poe.tasks.lint]
+help = "Run code linters (isort, black, flake8)"
+sequence = [
+    { cmd = "isort . ${checkonly:+--check --diff}", executor = { group = "lint" } },
+    { cmd = "isort docs ${checkonly:+--check --diff}", executor = { group = "lint" } },
+    { cmd = "black . ${checkonly:+--check --diff}", executor = { group = "lint" } },
+    { cmd = "flake8 . --count --show-source --statistics", executor = { group = "lint" } },
+]
+args = [{ name = "checkonly", type = "boolean" }]
+executor = { group = "lint" }
```
dlech: If we have this here, do we need the same on each command in the sequence?

.github/alpine-vm/start-alpine-vm.sh
timrid: No, that isn't strictly necessary for `poe`. But during the development of my poe PR, I installed poe directly from a Git branch, and that only works if Git is also installed on the system. Since Git is also listed as a prerequisite in `CONTRIBUTING.rst`, I just left it in there so people don't run into that issue again.

pyproject.toml
timrid: The `executor` option in sequence tasks does not seem to have any effect. In other words, it is not passed on to the child tasks. Therefore, the `executor` must be explicitly specified in each of the cmd tasks. While the poethepoet documentation does not mention this, I discovered it through experimentation.

During testing, I also noticed that with the `--group "test"` executor option, all `dev` dependencies are still installed, since `dev` is installed by default unless you explicitly exclude it using `--no-group "dev"`. To avoid installing all the unnecessary dependencies, I therefore added `--no-group "dev"` everywhere as well.

dlech:
Thanks!

PR #1981, opened 2026-05-05
bleak: cache BleakAdapter instances per event loop

Vodur:
`BleakAdapter.get()` now returns the same `BleakAdapter` instance on repeated calls from the same event loop with the same adapter selection. The cache is keyed by (loop, `PlatformBleakAdapter`, `adapter_path`), so different adapters selected via `bluez={"adapter": ...}` get distinct instances.

Mirrors the pattern in `bleak.backends.bluezdbus.manager._global_instances` introduced in 1be914c.

### DIFF
--- a/bleak/__init__.py
+++ b/bleak/__init__.py
@@ -9,9 +9,19 @@ import logging
 import os
 import sys
 import uuid
-from collections.abc import AsyncGenerator, Awaitable, Callable, Iterable
+from collections.abc import AsyncGenerator, Awaitable, Callable, Hashable, Iterable
 from types import TracebackType
-from typing import Any, Literal, Optional, TypedDict, Union, cast, overload
+from typing import (
+    Any,
+    ClassVar,
+    Literal,
+    Optional,
+    TypedDict,
+    Union,
+    cast,
+    final,
+    overload,
+)
 from warnings import warn
 
 from bleak._compat import Never, Self, Unpack, assert_never
@@ -424,6 +434,7 @@ class BleakScanner:
                 return None
 
 
+@final
 class BleakAdapter:
     """
     Interface for Bleak Bluetooth adapter operations.
@@ -437,6 +448,17 @@ class BleakAdapter:
     .. versionadded:: unreleased
     """
 
+    _instances: ClassVar[
+        dict[
+            tuple[
+                asyncio.AbstractEventLoop,
+                type[BaseBleakAdapter],
+                Hashable,
+            ],
+            "BleakAdapter",
+        ]
+    ] = {}
+
     def __init__(self, backend: BaseBleakAdapter):
         self._backend = backend
 
@@ -446,13 +468,16 @@ class BleakAdapter:
         *,
         bluez: BlueZAdapterArgs = {},
         backend: Optional[type[BaseBleakAdapter]] = None,
-    ) -> Self:
+    ) -> "BleakAdapter":
         """
         Get a Bluetooth adapter for the current platform.
 
         Currently all platforms other than Linux only allow one Bluetooth
         adapter. On Linux, the adapter can be selected via the ``bluez``
         argument with ``{"adapter": "<adapter_name>"}``.
+
+        Repeated calls from the same running event loop with the same
+        adapter selection return the same :class:`BleakAdapter` instance.
 
         Args:
             bluez:
@@ -472,7 +497,23 @@ class BleakAdapter:
             if backend is None
             else (backend, backend.__name__)
         )
-        return cls(await PlatformBleakAdapter.get(bluez=bluez))
+        loop = asyncio.get_running_loop()
+        key = (loop, PlatformBleakAdapter, PlatformBleakAdapter.cache_key(bluez=bluez))
+
+        try:
+            return cls._instances[key]
+        except KeyError:
+            pass
+
+        # Lazily clean up entries for closed loops so the cache does not
+        # accumulate over time. Mirrors the pattern in
+        # ``bleak.backends.bluezdbus.manager._global_instances``.
+        for closed_key in [k for k in cls._instances if k[0].is_closed()]:
+            del cls._instances[closed_key]
+
+        instance = cls(await PlatformBleakAdapter.get(bluez=bluez))
+        cls._instances[key] = instance
+        return instance
 
     async def get_connected_devices(
         self,
--- a/bleak/backends/adapter.py
+++ b/bleak/backends/adapter.py
@@ -1,3 +1,4 @@
+from collections.abc import Hashable
 from typing import Any
 
 from bleak._compat import Self
@@ -30,6 +31,18 @@ class BaseBleakAdapter:
             NotImplementedError: if the current backend does not support this.
         """
         raise NotImplementedError("get is not implemented for this backend")
+
+    @classmethod
+    def cache_key(cls, **kwargs: Any) -> Hashable:
+        """
+        Return a value used to differentiate cached :class:`BleakAdapter`
+        instances for this backend.
+
+        Subclasses override this to differentiate by backend-specific args
+        (e.g. the BlueZ adapter name). The default returns ``None``, which
+        means a single cache entry per (loop, backend type).
+        """
+        return None
 
     async def get_connected_devices(
         self, service_uuids: frozenset[str]
--- a/bleak/backends/bluezdbus/adapter.py
+++ b/bleak/backends/bluezdbus/adapter.py
@@ -5,6 +5,7 @@ if TYPE_CHECKING:
     if sys.platform != "linux":
         assert False, "This backend is only available on Linux"
 
+from collections.abc import Hashable
 from typing import Any
 
 from bleak._compat import Self, override
@@ -29,6 +30,11 @@ class BleakAdapterBlueZDBus(BaseBleakAdapter):
             f"/org/bluez/{adapter}" if adapter else manager.get_default_adapter()
         )
         return cls(adapter_path)
+
+    @classmethod
+    @override
+    def cache_key(cls, *, bluez: BlueZAdapterArgs = {}, **kwargs: Any) -> Hashable:
+        return bluez.get("adapter")
 
     @override
     async def get_connected_devices(
--- a/tests/integration/test_adapter.py
+++ b/tests/integration/test_adapter.py
@@ -73,6 +73,15 @@ async def connected_peripheral(
 
 
 @pytest.mark.asyncio(loop_scope="module")
+@pytest.mark.usefixtures("hci_transport")
+async def test_get_returns_same_instance_on_second_call() -> None:
+    """``BleakAdapter.get()`` returns the same instance on repeated calls."""
+    a = await BleakAdapter.get()
+    b = await BleakAdapter.get()
+    assert a is b
+
+
+@pytest.mark.asyncio(loop_scope="module")
 async def test_get_connected_devices_without_filters_returns_non_empty(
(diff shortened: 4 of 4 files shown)


bleak/__init__.py
```diff
@@ -472,7 +486,24 @@ async def get(
             if backend is None
             else (backend, backend.__name__)
         )
-        return cls(await PlatformBleakAdapter.get(bluez=bluez))
+        loop = asyncio.get_running_loop()
+        adapter_path = bluez.get("adapter")
```
dlech: This isn't going to work cross-platform. It needs to only be called if `PlatformBleakAdapter` is the BlueZ backend type. And we can't import that type here, otherwise it would crash on other platforms.

In practice, I guess it probably won't cause problems if we leave it like this, but logically, it doesn't seem quite right.

Ideally, we would move the backend-specific stuff into the backend itself. But I don't have any good ideas on how to do that without repeating code in each backend.

A hacky workaround could be to use `if PlatformBleakAdapter.__name__ == "BleakAdapterBlueZDBus":`.

tests/test_adapter_singleton.py
```diff
@@ -0,0 +1,119 @@
...
+    b = await BleakAdapter.get(backend=_FakeAdapter)
+    assert a is b
+    assert _FakeAdapter.construct_count == 1
+
+
+@pytest.mark.asyncio
+async def test_get_returns_different_instance_per_bluez_adapter():
+    a = await BleakAdapter.get(backend=_FakeAdapter, bluez={"adapter": "hci0"})
+    b = await BleakAdapter.get(backend=_FakeAdapter, bluez={"adapter": "hci1"})
+    assert a is not b
```
dlech: As mentioned above. Logically, this test should only pass on Linux and fail on other platforms.

tests/test_adapter_singleton.py
```diff
@@ -0,0 +1,119 @@
...
+import asyncio
+from typing import Any
+
+import pytest
+
+from bleak import BleakAdapter
+from bleak.backends.adapter import BaseBleakAdapter
+
+
+class _FakeAdapter(BaseBleakAdapter):
```
dlech: Why using fake adapter? It means we are testing less real code.

tests/test_adapter_singleton.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+class _OtherFakeAdapter(BaseBleakAdapter):
+    @classmethod
+    async def get(cls, **kwargs: Any) -> "_OtherFakeAdapter":
+        return cls()
+
+
+@pytest.fixture(autouse=True)
+def _clear_cache_and_counters():  # pyright: ignore[reportUnusedFunction]
```
dlech:
```suggestion
def clear_cache_and_counters():
```

Drop the `_` to avoid needing to ignore lint errors.

tests/test_adapter_singleton.py
```diff
@@ -0,0 +1,119 @@
...
+    BleakAdapter._instances.clear()  # pyright: ignore[reportPrivateUsage]
```
dlech: Each test should have a new event loop, so I'm not sure we actually need to do this.

tests/test_adapter_singleton.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+@pytest.mark.asyncio
+async def test_get_returns_different_instance_per_backend_type():
+    a = await BleakAdapter.get(backend=_FakeAdapter)
+    b = await BleakAdapter.get(backend=_OtherFakeAdapter)
+    assert a is not b
+
+
+def test_separate_event_loops_get_separate_instances():
```
dlech: As mentioned in another comment, pytest should already be providing a new loop for ever test, so this test should not be necessary.

tests/test_adapter_singleton.py
```diff
@@ -0,0 +1,119 @@
...
+    b = asyncio.new_event_loop()
+    try:
+        second = b.run_until_complete(one())
+    finally:
+        b.close()
+
+    assert first is not second
+
+
+def test_closed_loop_entries_are_pruned_lazily():
```
dlech: We should be able to see in code coverage if this is working or not because the `del` line is either called or not.

Not sure that this test adds much on top of that.

CHANGELOG.rst
```diff
@@ -14,6 +14,10 @@ Added
 -----
 * Added ``BleakAdapter`` class with ``get_connected_devices()`` to retrieve BLE devices that are already connected to the system without scanning.
 
+Changed
+-------
+* ``BleakAdapter.get()`` now returns the same ``BleakAdapter`` instance on repeated calls from the same event loop with the same adapter selection.
```
dlech: We haven't done a release with the previous behavior, so we don't need a changelog entry in this case.

Vodur:
Three CI failures from trying real `BleakAdapter.get()`:

- macOS (`build_desktop`): `BleakBluetoothNotAvailableError: Bluetooth is unsupported` - runner has no BT hardware
- Linux (`build_desktop`): `BleakDBusError: org.bluez was not provided by any .service files` - bluetoothd not installed
- Linux Alpine VM (`integration_tests_bluez`): `BleakBluetoothNotAvailableError: No Bluetooth adapters found` - VHCI is per-fixture, not active when our tests run

Real `BleakAdapter.get()` only works inside integration tests that have already set up their BT environment. So I switched to fake `BaseBleakAdapter` subclasses for unit tests. Real platform behavior is exercised by `tests/integration/test_adapter.py`.

Also added the `cache_key` `classmethod` (default `None`, BlueZ overrides to `bluez["adapter"]`) instead of the name-string check. 


Force-pushed 7fde0a3.

dlech:
I don't understand why we can't put the tests in the integration tests and avoid mocking the `BaseBleakAdapter`.

dlech:
> I don't understand why we can't put the tests in the integration tests and avoid mocking the `BaseBleakAdapter`.

Other than the special case of BlueZ with more than one adapter being used by Bleak at the same time.

Vodur:
Moved both tests to `tests/integration/test_adapter.py` using the existing `hci_transport` fixture - no fakes, real `BleakAdapter.get()`. The differentiation test for the multi-adapter case doesn't need two VHCIs: `BleakAdapterBlueZDBus.get()` doesn't validate the adapter name (just builds the D-Bus path string), so two arbitrary strings via `bluez={"adapter": ...}` are enough to verify the cache-key dispatch.

dlech:
> so two arbitrary strings via `bluez={"adapter": ...}` are enough to verify the cache-key dispatch.

I would expect calling `BleakAdapter.get(bluez={"adapter": "junk"})` should fail with an error that says that the adapter does not exist. So how could that test work? (I didn't look at the code yet.)

Vodur:
> > so two arbitrary strings via `bluez={"adapter": ...}` are enough to verify the cache-key dispatch.
> 
> I would expect calling `BleakAdapter.get(bluez={"adapter": "junk"})` should fail with an error that says that the adapter does not exist. So how could that test work? (I didn't look at the code yet.)

Right - currently `BleakAdapterBlueZDBus.get()` doesn't validate the adapter name (you noted on #1979 that we'd defer fixing it). The test passes because of that, not because cache-key dispatch is being meaningfully exercised.

Easiest path: drop the differentiation test from this PR.

bleak/backends/bluezdbus/adapter.py
```diff
@@ -30,6 +30,11 @@ async def get(cls, *, bluez: BlueZAdapterArgs = {}, **kwargs: Any) -> Self:
         )
         return cls(adapter_path)
 
+    @classmethod
+    @override
+    def cache_key(cls, *, bluez: BlueZAdapterArgs = {}, **kwargs: Any) -> Optional[str]:
+        return bluez.get("adapter")
```
dlech, line 37: I'm not sure this works in all cases.

Consider the following on a system with two Bluetooth adapters:
* Someone calls `BleakAdapter.get()` which gets the "best" adapter. `cache_key` is `None` because no extra args were passed.
* Something changes that affects the "best" adapter algorithm (one adapter is powered down, USB adapter is removed, etc.)
* Now, calling `BleakAdapter.get()` will return an adapter that might not be there any more rather than the "best" adapter.

It hasn't really come up before of what the expected behavior should be when this sort of thing happens. Maybe it is OK the like this for now.

bleak/backends/bluezdbus/adapter.py
```diff
@@ -30,6 +30,11 @@ async def get(cls, *, bluez: BlueZAdapterArgs = {}, **kwargs: Any) -> Self:
         )
         return cls(adapter_path)
 
+    @classmethod
+    @override
+    def cache_key(cls, *, bluez: BlueZAdapterArgs = {}, **kwargs: Any) -> Optional[str]:
```
dlech:
```suggestion
    def cache_key(cls, *, bluez: BlueZAdapterArgs = {}, **kwargs: Any) -> Hashable:
```

Needs to match signature we are overriding.

bleak/backends/adapter.py
```diff
@@ -31,6 +32,18 @@ async def get(cls, **kwargs: Any) -> Self:
         """
         raise NotImplementedError("get is not implemented for this backend")
 
+    @classmethod
+    def cache_key(cls, **kwargs: Any) -> Optional[Hashable]:
```
dlech:
```suggestion
    def cache_key(cls, **kwargs: Any) -> Hashable:
```

`None` is hashable and isn't special in this case.

bleak/__init__.py
```diff
@@ -437,6 +437,17 @@ class BleakAdapter:
     .. versionadded:: unreleased
     """
 
+    _instances: ClassVar[
+        dict[
+            tuple[
+                asyncio.AbstractEventLoop,
+                type[BaseBleakAdapter],
+                Optional[Hashable],
```
```suggestion
                Hashable,
```

bleak/backends/bluezdbus/adapter.py
dlech: Or replace `Hashable` with `str | None` everywhere in this patch.

bleak/__init__.py
```diff
@@ -472,7 +486,23 @@ async def get(
             if backend is None
             else (backend, backend.__name__)
         )
-        return cls(await PlatformBleakAdapter.get(bluez=bluez))
+        loop = asyncio.get_running_loop()
+        key = (loop, PlatformBleakAdapter, PlatformBleakAdapter.cache_key(bluez=bluez))
+
+        try:
+            return cast("Self", cls._instances[key])
```
dlech: Why do we need `cast` here?

Maybe better to make `BleakAdapter` `@final`?

tests/integration/test_adapter.py
```diff
@@ -72,6 +72,16 @@ async def connected_peripheral(
         yield ConnectedPeripheral(client=client, device=device)
 
 
+@pytest.mark.asyncio(loop_scope="module")
+async def test_get_returns_same_instance_on_second_call(
+    hci_transport: Transport,
```
dlech: `hci_transport` is not used.

Vodur:
Tried `@final` first but pyright doesn't narrow `self` to `BleakAdapter` even on a final class, so the cast was still needed. Went with changing the return type from `self` to `"BleakAdapter"` since the class is final - but that's a public-signature change you didn't actually ask for.

dlech:
Sorry it took a while to get back to this. Can you remind me what the motivation of why this change is beneficial in the first place? Usually, we want this sort of explanation in the commit message so that future us don't try to undo it.

PR #1984, opened 2026-05-06
backends/bluez: don't disconnect devices we did not connect

Vodur:
`BleakClientBlueZDBus.disconnect()` now only calls BlueZ Disconnect when this client initiated the connection. The decision is based on a `from_connected_devices` marker that `BleakAdapterBlueZDBus.get_connected_devices()` sets on each returned `BLEDevice`.

If a device is found already connected at connect() time and the user did not opt into the existing connection (no marker), the client still owns the connection so it can clean up an orphan from a crashed previous run. Only when the `BLEDevice` was retrieved via `BleakAdapter.get_connected_devices()` does the new client leave the connection up on close.

See https://github.com/bluez/bluez/issues/89.

### DIFF
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -79,9 +79,16 @@ class BleakClientBlueZDBus(BaseBleakClient):
         if isinstance(address_or_ble_device, BLEDevice):
             self._device_path = address_or_ble_device.details["path"]
             self._device_info = address_or_ble_device.details.get("props")
+            # Set by BleakAdapter.get_connected_devices(). When True, the
+            # user opted into an existing OS-level connection and
+            # disconnect() must not tear it down.
+            self._from_connected_devices = bool(
+                address_or_ble_device.details.get("from_connected_devices")
+            )
         else:
             self._device_path = None
             self._device_info = None
+            self._from_connected_devices = False
 
         self._requested_services = services
 
@@ -211,7 +218,7 @@ class BleakClientBlueZDBus(BaseBleakClient):
                         # if connection was successful but _get_services() raises (e.g.
                         # because task was cancelled), then we still need to disconnect
                         # before passing on the exception.
-                        if self._bus:
+                        if self._bus and not self._from_connected_devices:
                             # If disconnected callback already fired, this will be a no-op
                             # since self._bus will be None and the _cleanup_all call will
                             # have already disconnected.
@@ -406,19 +413,31 @@ class BleakClientBlueZDBus(BaseBleakClient):
             self._disconnecting_event = asyncio.Event()
             try:
                 if self.is_connected:
-                    # Try to disconnect the actual device/peripheral
-                    reply = await self._bus.call(
-                        Message(
-                            destination=defs.BLUEZ_SERVICE,
-                            path=self._device_path,
-                            interface=defs.DEVICE_INTERFACE,
-                            member="Disconnect",
+                    if not self._from_connected_devices:
+                        # Try to disconnect the actual device/peripheral
+                        reply = await self._bus.call(
+                            Message(
+                                destination=defs.BLUEZ_SERVICE,
+                                path=self._device_path,
+                                interface=defs.DEVICE_INTERFACE,
+                                member="Disconnect",
+                            )
                         )
-                    )
-                    assert_reply(reply)
+                        assert_reply(reply)
 
-                    async with async_timeout(10):
-                        await self._disconnecting_event.wait()
+                        async with async_timeout(10):
+                            await self._disconnecting_event.wait()
+                    else:
+                        # Connection was opted into; leave it up. Mirror the
+                        # cleanup that on_connected_changed would do, except
+                        # for _disconnected_callback (the device is not
+                        # actually disconnecting from BlueZ's perspective).
+                        self._is_connected = False
+                        if self._disconnect_monitor_event:
+                            self._disconnect_monitor_event.set()
+                            self._disconnect_monitor_event = None
+                        self._cleanup_all()
+                        self._disconnecting_event.set()
             finally:
                 self._disconnecting_event = None
 
--- a/tests/integration/test_adapter.py
+++ b/tests/integration/test_adapter.py
@@ -1,3 +1,4 @@
+import asyncio
 import dataclasses
 from collections.abc import AsyncGenerator
 
@@ -10,6 +11,7 @@ from bumble.gatt import Characteristic, Service
 from bumble.transport.common import Transport
 
 from bleak import BleakAdapter, BleakClient
+from bleak.backends import BleakBackend, get_default_backend
 from bleak.backends.device import BLEDevice
 from tests.integration.conftest import (
     configure_and_power_on_bumble_peripheral,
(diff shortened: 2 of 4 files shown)


tests/integration/test_adapter.py
```diff
@@ -99,3 +99,22 @@ async def test_get_connected_devices_filters_by_service_uuid(
...
+    disconnect the outer client. See https://github.com/bluez/bluez/issues/89.
+    """
+    adapter = await BleakAdapter.get()
+    connected_devices = await adapter.get_connected_devices([TEST_SERVICE_UUID])
+    inner_target = next(
+        d for d in connected_devices if d.address == connected_peripheral.device.address
+    )
+
+    async with BleakClient(inner_target) as inner_client:
+        assert inner_client.is_connected
```
dlech: I would still like to see a comment here (see https://github.com/hbldh/bleak/pull/1975#discussion_r3161888822) or replace it with a short sleep.

tests/integration/test_adapter.py
```diff
@@ -99,3 +99,22 @@ async def test_get_connected_devices_filters_by_service_uuid(
 
     not_connected = await adapter.get_connected_devices([OTHER_SERVICE_UUID])
     assert not_connected == []
+
+
+@pytest.mark.asyncio(loop_scope="module")
+async def test_inner_client_does_not_disconnect_outer(
+    connected_peripheral: ConnectedPeripheral,
+) -> None:
+    """Inner BleakClient using a device from get_connected_devices() must not
```
dlech: We will also want a similar test that call disconnect on `connected_peripheral.client` inside of the `async with` of the already connected device.

With BlueZ, this will cause `inner_client` to also disconnect (undesirable, but a limitation of BlueZ - we should have comments to explain that). On other platforms though, `inner_client` should stay connected.

tests/integration/test_adapter.py
```diff
@@ -99,3 +99,22 @@ async def test_get_connected_devices_filters_by_service_uuid(
...
+    disconnect the outer client. See https://github.com/bluez/bluez/issues/89.
+    """
+    adapter = await BleakAdapter.get()
+    connected_devices = await adapter.get_connected_devices([TEST_SERVICE_UUID])
+    inner_target = next(
```
dlech: Not sure "inner" naming means much now that there is no corresponding "outer" counterpart. Maybe we could come up with a better name?

bleak/backends/bluezdbus/client.py
```diff
@@ -78,9 +78,14 @@ def __init__(
         if isinstance(address_or_ble_device, BLEDevice):
             self._device_path = address_or_ble_device.details["path"]
             self._device_info = address_or_ble_device.details.get("props")
+            # Set by BleakAdapter.get_connected_devices(); see connect().
+            self._opted_into_existing_connection = bool(
```
```suggestion
            self._from_connected_devices = bool(
```

I would keep the same naming to make the code easier to reason about.

bleak/backends/bluezdbus/client.py
```diff
@@ -78,9 +78,14 @@ def __init__(
...
+                address_or_ble_device.details.get("from_connected_devices", False)
```
```suggestion
                address_or_ble_device.details.get("from_connected_devices")
```

Since we are casting to bool, we can let `None` turn into `False`.

bleak/backends/bluezdbus/client.py
```diff
@@ -90,6 +95,9 @@ def __init__(
         self._remove_device_watcher: Optional[Callable[[], None]] = None
         # private backing for is_connected property
         self._is_connected = False
+        # True if this BleakClient initiated the BlueZ connection. Set in
+        # connect(); gates the BlueZ Disconnect call in disconnect().
+        self._owns_connection = False
```
dlech: This seems redundant with `_opted_into_existing_connection`. Ownership is never going to change for the same instance of a `BleakClient`. So I would expect that we only add one state variable in this patch.

tests/integration/test_adapter.py
```diff
@@ -99,3 +100,55 @@ async def test_get_connected_devices_filters_by_service_uuid(
...
+    connected_peripheral: ConnectedPeripheral,
+) -> None:
+    """When the BleakClient that originally established the connection
+    disconnects, an attached BleakClient (using a device from
+    get_connected_devices()) sees the disconnect on BlueZ.
+
+    BlueZ does not ref-count connections, so any client tearing down the
+    connection takes everyone else with it. Other backends ref-count and
+    would keep the attached client connected; this test runs against
+    BlueZ via VHCI.
```
dlech: This should be a comment in the test.

tests/integration/test_adapter.py
```diff
@@ -99,3 +100,55 @@ async def test_get_connected_devices_filters_by_service_uuid(
...
+        d for d in connected_devices if d.address == connected_peripheral.device.address
+    )
+
+    # NOTE: this test must be the last in the module since it leaves the
+    # fixture's BleakClient in a disconnected state. Reconnecting Bumble
+    # immediately after a cascading disconnect is unreliable.
+    async with BleakClient(target) as client:
+        await connected_peripheral.client.disconnect()
+        await asyncio.sleep(0.5)
+        assert not client.is_connected
```
dlech:
```suggestion
        if get_default_backend() == BleakBackend.BLUEZ_DBUS:
            # BlueZ quirk comment from above goes here.
            assert not client.is_connected
        else:
            assert client.is_connected
```

CHANGELOG.rst
```diff
@@ -17,6 +17,7 @@ Added
 Fixed
 -----
 * Fixed handling empty notification payloads in BlueZ backend when using "AcquireNotify". Fixes #1982.
+* ``BleakClientBlueZDBus.disconnect()`` no longer disconnects from the device when the :class:`BLEDevice` was retrieved from :meth:`BleakAdapter.get_connected_devices`.
```
dlech: There has not been a release with `BleakAdapter` yet, so this doesn't belong in the changelog.

We should document the BlueZ quirk in the docs though.

tests/integration/test_adapter.py
dlech: we will need to manually test this on Windows and Mac to make sure it works.

tests/integration/test_adapter.py
```diff
@@ -99,3 +100,55 @@ async def test_get_connected_devices_filters_by_service_uuid(
...
     assert not_connected == []
+
+
+@pytest.mark.asyncio(loop_scope="module")
+async def test_disconnect_does_not_disconnect_connected_device(
+    connected_peripheral: ConnectedPeripheral,
+) -> None:
+    """A BleakClient using a device from get_connected_devices() must not
+    disconnect the underlying connection on close. See
+    https://github.com/bluez/bluez/issues/89.
```
dlech: Not sure this bug is relevant here. It is only relevant in the BlueZ implementation. This test is the expected behavior for all platforms. And we managed to make the BlueZ work the same, so nothing special to call out here, I don't think.

The bug is relevant in the other test though, since there BlueZ had different behavior from other backends.

tests/integration/test_adapter.py
Vodur: Manually verified the non-BlueZ branch of `test_disconnect_of_originating_client_disconnects_attached_client` against a real BLE peripheral on both supported non-Linux platforms:

Windows 11:

> [4/5] Disconnecting originator while attached is still open...
>       attached.is_connected = True
> [5/5] Verifying expectation for this platform...
> PASS: attached stayed connected (Win/macOS ref-counted handle)

macOS (Darwin 25.3.0):

> [4/5] Disconnecting originator while attached is still open...
>       attached.is_connected = True
> [5/5] Verifying expectation for this platform...
> PASS: attached stayed connected (Win/macOS ref-counted handle)

Both confirm the OS-level handle ref-counting keeps the attached client connected when the originator disconnects. Linux/BlueZ behavior is covered by the integration test.

tests/integration/test_adapter.py
Vodur: Added the comment explaining why `is_connected` is asserted right after the `async with` despite being trivially true - see lines 120-123.

tests/integration/test_adapter.py
Vodur: Added as `test_disconnect_of_originating_client_disconnects_attached_client` (line 129). Asserts `not client.is_connected` on BlueZ and `client.is_connected` on other backends, with a comment explaining the BlueZ ref-counting limitation.

tests/integration/test_adapter.py
Vodur: Dropped the `inner_client`/`outer_client` naming entirely. Tests now use `client` (the attached one), `target` (the `BLEDevice` from `get_connected_devices`), and `connected_peripheral.client` (the fixture's originating client).

bleak/backends/bluezdbus/client.py
Vodur: Applied - `self._from_connected_devices` throughout. Consolidated to a single state variable per your comment below.

bleak/backends/bluezdbus/client.py
Vodur: Applied - `bool(address_or_ble_device.details.get("from_connected_devices"))` in `__init__` (line 85).

bleak/backends/bluezdbus/client.py
Vodur: Consolidated to a single `_from_connected_devices` variable; `_opted_into_existing_connection` removed. The marker is set once in `__init__` based on `BLEDevice.details["from_connected_devices"]` and read only in `disconnect_device()` to skip the BlueZ `Disconnect` call.

tests/integration/test_adapter.py
Vodur: Moved into an inline `NOTE` comment in `test_disconnect_of_originating_client_disconnects_attached_client` (lines 144-146) explaining why it must be last in the module.

CHANGELOG.rst
Vodur: Removed the changelog entry. Added a `.. note::` block in `docs/api/adapter.rst` (lines 53-62) documenting the BlueZ no-ref-counting behavior with a link to https://github.com/bluez/bluez/issues/89.

tests/integration/test_adapter.py
Vodur: Agreed - removed the BlueZ bug reference from `test_disconnect_does_not_disconnect_connected_device` (which is the expected behavior on every platform). The reference now lives only in `test_disconnect_of_originating_client_disconnects_attached_client`, where BlueZ actually diverges.

dlech:
AI review caught some things that look important:

* The attached-client disconnect path skips the disconnected callback. The PR's else branch in disconnect() deliberately does not call _disconnected_callback, arguing the device isn't disconnecting from BlueZ's view. But the other backends do fire it on a client-initiated disconnect: WinRT synthesizes it explicitly at [client.py:469-477](vscode-webview://1i9p9cgoucrasa3fpibgtu6lsqg23rr3e82a59bkuofcf7mtij9v/bleak/backends/winrt/client.py#L469-L477), and CoreBluetooth gets it via the delegate. The callback's contract is "this client is no longer connected", and is_connected does flip to False here. Skipping it makes the BlueZ behavior diverge from the platforms this PR is trying to mimic.
* The crash-safety monitor ignores the flag. _disconnect_monitor at [client.py:333-363](vscode-webview://1i9p9cgoucrasa3fpibgtu6lsqg23rr3e82a59bkuofcf7mtij9v/bleak/backends/bluezdbus/client.py#L333-L363) sends Disconnect on cancellation regardless of ownership. If the event loop dies, an attached client will still tear down a connection some other process owns. Simplest fix is to not spawn the monitor task when the client didn't establish the connection.
* Edge case where the flag becomes wrong. If the BLEDevice came from get_connected_devices() but the link drops before connect() runs, manager.is_connected() is False, so this client calls Connect and now owns the link. The flag still says "don't disconnect", so the connection leaks until some later client cleans up the orphan. Ownership should really be decided at connect time as from_connected_devices and manager.is_connected(path), kept in a session-local variable rather than only the constructor value. This slightly revises your earlier "one state variable" comment, but the two variables mean different things: one is input, the other is the per-session decision.

PR #1986, opened 2026-05-11
bleak: raise BleakBluetoothNotAvailableError from BleakAdapter.get() on Linux and Windows

Vodur:
I would prefer to just consider making raising of `BleakBluetoothNotAvailableError` from `BleakAdapter.get()` consistent on all platforms first.

`BleakAdapter.get()` now raises `BleakBluetoothNotAvailableError` on all platforms when the adapter is not powered on. CoreBluetooth already did this via `wait_until_ready()`; this adds the same behavior to the BlueZ and WinRT backends:

- **BlueZ** — checks that the adapter is present in the manager's cached properties and that `Powered` is set; raises with reason `NO_BLUETOOTH` or `POWERED_OFF`.
- **WinRT** — fetches the Bluetooth radio via `Radio.get_radios_async()` and checks `RadioState`; raises with reason `NO_BLUETOOTH` or `POWERED_OFF`.

### Testing

- BlueZ raise path is covered by an integration test (`test_get_raises_when_powered_off`) that power-cycles the adapter via `bluetoothctl` in the vhci VM.
- The WinRT raise path has no coverage, since there is no Windows integration runner in CI — this is the remaining `codecov/patch` gap.

The `state` property and `subscribe_state_changes()` work (native event sources, no polling) is deferred to a follow-up PR.

Refs #320, #1060.

Notes:
- I quoted dlech's exact words so it's unambiguous we're responding to his prioritization, not paraphrasing.
- The testing section pre-explains the codecov red so he reads it as intentional, not an oversight — folded into the description rather than a separate comment, since you said one-by-one and this keeps it self-contained.
- "deferred to a follow-up PR" sets up the bleak-adapter-state-fullscope work without committing you to open it now.

### DIFF
--- a/bleak/backends/bluezdbus/adapter.py
+++ b/bleak/backends/bluezdbus/adapter.py
@@ -10,8 +10,10 @@ from typing import Any
 from bleak._compat import Self, override
 from bleak.args.bluez import BlueZAdapterArgs
 from bleak.backends.adapter import BaseBleakAdapter
+from bleak.backends.bluezdbus import defs
 from bleak.backends.bluezdbus.manager import get_global_bluez_manager
 from bleak.backends.device import BLEDevice
+from bleak.exc import BleakBluetoothNotAvailableError, BleakBluetoothNotAvailableReason
 
 
 class BleakAdapterBlueZDBus(BaseBleakAdapter):
@@ -25,9 +27,27 @@ class BleakAdapterBlueZDBus(BaseBleakAdapter):
     async def get(cls, *, bluez: BlueZAdapterArgs = {}, **kwargs: Any) -> Self:
         manager = await get_global_bluez_manager()
         adapter = bluez.get("adapter")
-        adapter_path = (
-            f"/org/bluez/{adapter}" if adapter else manager.get_default_adapter()
-        )
+
+        if adapter is None:
+            # get_default_adapter() already raises BleakBluetoothNotAvailableError
+            # if there is no powered BLE-central adapter.
+            return cls(manager.get_default_adapter())
+
+        adapter_path = f"/org/bluez/{adapter}"
+        adapter_props = manager._properties.get(  # pyright: ignore[reportPrivateUsage]
+            adapter_path, {}
+        ).get(defs.ADAPTER_INTERFACE)
+        if adapter_props is None:
+            raise BleakBluetoothNotAvailableError(
+                f"Bluetooth adapter '{adapter_path}' is unavailable",
+                BleakBluetoothNotAvailableReason.NO_BLUETOOTH,
+            )
+        if not adapter_props.get("Powered"):
+            raise BleakBluetoothNotAvailableError(
+                "Bluetooth adapter is not powered on",
+                BleakBluetoothNotAvailableReason.POWERED_OFF,
+            )
+
         return cls(adapter_path)
 
     @override
--- a/bleak/backends/winrt/adapter.py
+++ b/bleak/backends/winrt/adapter.py
@@ -9,6 +9,7 @@ from uuid import UUID
 
 from winrt.system import unbox_string
 from winrt.windows.devices.bluetooth import (
+    BluetoothAdapter,
     BluetoothCacheMode,
     BluetoothConnectionStatus,
     BluetoothDeviceId,
@@ -16,11 +17,13 @@ from winrt.windows.devices.bluetooth import (
 )
 from winrt.windows.devices.bluetooth.genericattributeprofile import GattDeviceService
 from winrt.windows.devices.enumeration import DeviceInformation
+from winrt.windows.devices.radios import RadioState
 
 from bleak._compat import Self, override
 from bleak.backends.adapter import BaseBleakAdapter
 from bleak.backends.device import BLEDevice
 from bleak.backends.winrt.util import assert_mta
+from bleak.exc import BleakBluetoothNotAvailableError, BleakBluetoothNotAvailableReason
 from bleak.uuids import normalize_uuid_16
 
 
@@ -31,6 +34,32 @@ class BleakAdapterWinRT(BaseBleakAdapter):
     @override
     async def get(cls, **kwargs: Any) -> Self:
         await assert_mta()
+
+        # TODO: need to fix return type of get_default_async() in PyWinRT
+        adapter = await BluetoothAdapter.get_default_async()
+        if adapter is None:  # pyright: ignore[reportUnnecessaryComparison]
+            raise BleakBluetoothNotAvailableError(
+                "No Bluetooth adapter found",
+                BleakBluetoothNotAvailableReason.NO_BLUETOOTH,
+            )
+        if not adapter.is_central_role_supported:
+            raise BleakBluetoothNotAvailableError(
+                "BLE 'central' role not supported on this adapter",
+                BleakBluetoothNotAvailableReason.NO_BLE_CENTRAL_ROLE,
+            )
+
+        radio = await adapter.get_radio_async()
+        if radio.state == RadioState.UNKNOWN:
+            raise BleakBluetoothNotAvailableError(
+                "Bluetooth radio state is unknown",
+                BleakBluetoothNotAvailableReason.UNKNOWN,
+            )
+        if radio.state != RadioState.ON:
+            raise BleakBluetoothNotAvailableError(
+                "Bluetooth adapter is not powered on",
+                BleakBluetoothNotAvailableReason.POWERED_OFF,
+            )
+
         return cls()
 
     @override
--- a/bleak/backends/winrt/scanner.py
+++ b/bleak/backends/winrt/scanner.py
@@ -10,7 +10,6 @@ import logging
 from typing import Literal, NamedTuple, Optional
 from uuid import UUID
 
-from winrt.windows.devices.bluetooth import BluetoothAdapter
 from winrt.windows.devices.bluetooth.advertisement import (
     BluetoothLEAdvertisementReceivedEventArgs,
     BluetoothLEAdvertisementType,
@@ -19,7 +18,6 @@ from winrt.windows.devices.bluetooth.advertisement import (
     BluetoothLEAdvertisementWatcherStoppedEventArgs,
     BluetoothLEScanningMode,
 )
-from winrt.windows.devices.radios import RadioState
 from winrt.windows.foundation import EventRegistrationToken
 
 from bleak._compat import override
@@ -29,12 +27,9 @@ from bleak.backends.scanner import (
     AdvertisementDataCallback,
     BaseBleakScanner,
 )
+from bleak.backends.winrt.adapter import BleakAdapterWinRT
 from bleak.backends.winrt.util import assert_mta
-from bleak.exc import (
-    BleakBluetoothNotAvailableError,
-    BleakBluetoothNotAvailableReason,
-    BleakError,
-)
+from bleak.exc import BleakError
 from bleak.uuids import normalize_uuid_str
 
 logger = logging.getLogger(__name__)
@@ -241,26 +236,9 @@ class BleakScannerWinRT(BaseBleakScanner):
         # there is nothing pumping a Windows message loop.
         await assert_mta()
 
-        # TODO: need to fix return type of get_default_async() in PyWinRT
-        adapter = await BluetoothAdapter.get_default_async()
-        if adapter is None:  # pyright: ignore[reportUnnecessaryComparison]
-            raise BleakBluetoothNotAvailableError(
-                "No Bluetooth adapter found",
-                BleakBluetoothNotAvailableReason.NO_BLUETOOTH,
-            )
-
-        if not adapter.is_central_role_supported:
-            raise BleakBluetoothNotAvailableError(
-                "BLE 'central' role not supported on this adapter",
-                BleakBluetoothNotAvailableReason.NO_BLE_CENTRAL_ROLE,
-            )
-
-        radio = await adapter.get_radio_async()
-        if radio.state != RadioState.ON:
-            raise BleakBluetoothNotAvailableError(
-                "Bluetooth radio is not powered on. Turn on Bluetooth and try again.",
-                BleakBluetoothNotAvailableReason.POWERED_OFF,
-            )
+        # Validate that a usable BLE central adapter is available. This raises
+        # BleakBluetoothNotAvailableError if not.
+        await BleakAdapterWinRT.get()
 
         # start with fresh list of discovered devices
         self.seen_devices = {}
--- a/docs/api/adapter.rst
+++ b/docs/api/adapter.rst
@@ -26,6 +26,10 @@ On Linux, a specific adapter can be selected via the ``bluez`` argument::
 
     adapter = await BleakAdapter.get(bluez={"adapter": "hci1"})
 
+:meth:`BleakAdapter.get` raises :class:`~bleak.exc.BleakBluetoothNotAvailableError`
+if the local Bluetooth adapter is not currently powered on or is otherwise
+unavailable.
+
 .. automethod:: bleak.BleakAdapter.get
 
 -----------------------------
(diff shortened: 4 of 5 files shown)


dlech:
I've been hoping to move Bleak to using structured concurrency.

If we add a property that can be read, then people are just going to poll it in a loop with a sleep, which is not good program design. So I would prefer to leave out the `state` property.

And callbacks aren't good program design for async programming either. They were just the only thing available when Bleak was written. So I would prefer to avoid that too.

Structured concurrency might look something like this:

```python
async def main():
    while True:
        try:
            async with BleakAdapter.get():
                # if Bluetooth adapter state changes inside of this
                # context, it will cancel the task
                await do_stuff_with_bluetooth()
        except BleakBluetoothNotAvailableError as ex:
            try_again = await inform_user_of_reason_and_and_ask(ex)
            if try_again:
                continue

        break
```

I think this is going to take more though and trying out different things to figure out if this can actually work as intended. I consider it lower priority that other things we have open right now.

dlech:
> Windows has no native event

It looks like it does.

https://learn.microsoft.com/en-us/uwp/api/windows.devices.radios.radio.statechanged?view=winrt-28000

Vodur:
Thanks - structured concurrency does make sense, and `Radio.StateChanged` exists, so all three platforms have a real event source we can hang the cancellation off:

 | Platform | Native event                                                         |
 |----------|----------------------------------------------------------------------|
 | Linux    | D-Bus `PropertiesChanged` on `org.bluez.Adapter1`                        |
 | macOS    | `CentralManagerDelegate.centralManagerDidUpdateState_` (already wired) |
 | Windows  | `Radio.StateChanged`                                                   |

Building on your sketch, this is roughly how the user side could look:

```python
 from contextlib import asynccontextmanager

 async def main():
     while True:
         try:
             # async ctx mgr: raises immediately if not POWERED_ON;
             # cancels the body if state changes mid-context.
             async with BleakAdapter.acquire() as adapter:
                 await do_stuff_with_bluetooth(adapter)
         except BleakBluetoothNotAvailableError as ex:
             if not await inform_user_and_retry(ex):
                 break

 Implementation-side it'd be something like (rough sketch):

 @asynccontextmanager
 async def acquire(cls, *, bluez: BlueZAdapterArgs = {}):
     backend = await PlatformBleakAdapter.get(bluez=bluez)
     if backend.state != AdapterState.POWERED_ON:
         raise BleakBluetoothNotAvailableError(...)

     async with asyncio.TaskGroup() as tg:
         async def _watch() -> None:
             async for state in backend.state_events():  # platform-native source
                 if state != AdapterState.POWERED_ON:
                     raise BleakBluetoothNotAvailableError(...)
         tg.create_task(_watch())
         try:
             yield cls(backend)
         finally:
             tg.cancel()
```

If you'd rather close this and revisit later, happy to. If the sketch above is roughly the direction you had in mind, I can take a stab at it as a fresh PR (using `state_events()` async-iterators per backend and the native `StateChanged` on Windows).

dlech:
> backend = await PlatformBleakAdapter.get(bluez=bluez)
>      if backend.state != AdapterState.POWERED_ON:
>          raise BleakBluetoothNotAvailableError(...)

Actually, `PlatformBleakAdapter.get()` should already be raising `BleakBluetoothNotAvailableError` if we can't get a powered on adapter. It is probably only implemented in CoreBluetooth right now though.

dlech:
As for the async context manager implementation and async iterator, this can be really tricky to get right. I've been considering using [anyio](https://anyio.readthedocs.io/en/stable/index.html) to help with this as it has solved a lot of these hard-to-get-right things already.

And we can't use `TaskGroup` in the implementation until we drop support for Python 3.10.

Vodur:
Splitting the work: this PR ships building blocks only - `state` property, `get()` raising `BleakBluetoothNotAvailableError` when not powered on (your CoreBluetooth `wait_until_ready` point, now matched on Linux/Windows), and `subscribe_state_changes(callback)` as an `async with`-scoped subscription using native events on each platform: Linux `PropertiesChanged`, macOS `delegate`, Windows `Radio.StateChanged` (which you were right exists - I was wrong about that earlier, sorry).

The cancel-on-state-change wrapper goes in a follow-up where the anyio decision lives. Whether it will be anyio or dropping support for Python 3.10.

Ok to push?

dlech:
I would prefer to just consider making raising of `BleakBluetoothNotAvailableError` from `BleakAdapter.get()` consistent on all platforms first.

All other proposed changes here are lower priority for me.

Vodur:
Only codecov patch is failing here. The WinRT raise paths in `BleakAdapter.get()` have no coverage since there's no Windows integration runner, and I dropped the WinRT unit tests in favor of integration. The BlueZ path is covered by an integration test.

Fine to leave the WinRT paths uncovered, or would you rather I add a Windows integration job?

CHANGELOG.rst
```diff
@@ -14,6 +14,10 @@ Added
 -----
 * Added ``BleakAdapter`` class with ``get_connected_devices()`` to retrieve BLE devices that are already connected to the system without scanning.
 
+Changed
+-------
+* Changed ``BleakAdapter.get()`` to raise ``BleakBluetoothNotAvailableError`` on Linux and Windows when the local Bluetooth adapter is not powered on, matching the existing CoreBluetooth behaviour.
```
dlech: There has still not been a release with `BleakAdapter`, so changelog for anything changing `BleakAdapter` does not make sense. The `Added` section already covers adding this new class.

bleak/backends/winrt/adapter.py
```diff
@@ -31,6 +33,23 @@ class BleakAdapterWinRT(BaseBleakAdapter):
     @override
     async def get(cls, **kwargs: Any) -> Self:
         await assert_mta()
+
+        radio = None
+        for candidate in await Radio.get_radios_async():
+            if candidate.kind == RadioKind.BLUETOOTH:
+                radio = candidate
+                break
+        if radio is None:
```
dlech:
```suggestion
        for radio in await Radio.get_radios_async():
            if radio.kind == RadioKind.BLUETOOTH:
                break
        else:
```

Can be a bit simpler this way.

bleak/backends/winrt/adapter.py
```diff
@@ -31,6 +33,23 @@ class BleakAdapterWinRT(BaseBleakAdapter):
...
+            raise BleakBluetoothNotAvailableError(
+                "No Bluetooth radio available",
+                BleakBluetoothNotAvailableReason.NO_BLUETOOTH,
+            )
+        if radio.state != RadioState.ON:
```
dlech, line 57: I think we could at least map `RadioState.UNKNOWN` to `BleakBluetoothNotAvailableReason.UNKNOWN`.

I don't remember if other OSes made the distinction between hardware switch off (`RadioState.DISCONNECTED`) and software controllable off (`Radio.OFF`). We should probably look at that.

bleak/backends/bluezdbus/adapter.py
```diff
@@ -28,6 +30,21 @@ async def get(cls, *, bluez: BlueZAdapterArgs = {}, **kwargs: Any) -> Self:
         adapter_path = (
             f"/org/bluez/{adapter}" if adapter else manager.get_default_adapter()
         )
+
```
dlech, line 30: It would probably be better to just call `BlueZManager.get_default_adapter()` from here so that we don't have to duplicate the logic.

bleak/backends/winrt/adapter.py
dlech: And we could call `BleakAdapter.get()` in `BleakScannerWinRT.start()` to avoid duplicating the code.

dlech (commented):
This one is hard to test since it requires very conditions on the OS. So I am OK to leave figuring out some good tests for later.

CHANGELOG.rst
Vodur: removed

bleak/backends/winrt/adapter.py
Vodur: `RadioState.UNKNOWN` now maps to `UNKNOWN`, `Off`/`Disabled` map to `POWERED_OFF`.

bleak/backends/bluezdbus/adapter.py
Vodur: fixed

bleak/backends/winrt/adapter.py
Vodur: radio loop is gone, `get()` now uses `BluetoothAdapter.get_default_async()` instead.

bleak/backends/winrt/adapter.py
Vodur: > And we could call `BleakAdapter.get()` in `BleakScannerWinRT.start()` to avoid duplicating the code.

`BleakScannerWinRT.start()` now calls `BleakAdapterWinRT.get()`, and `get()` was reworked to use `BluetoothAdapter` so the central-role and power checks live in one place.

Vodur: > I don't remember if other OSes made the distinction between hardware switch off (`RadioState.DISCONNECTED`) and software controllable off (`Radio.OFF`). We should probably look at that.


- Windows exposes the distinction reliably: `RadioState.Disabled` is the hardware/firmware switch (not programmatically changeable), `Off` is the software toggle.
- macOS `CBManagerState.poweredOff` doesn't carry the cause at all.
- BlueZ only surfaces it through the experimental `PowerState` property (`off-blocked` is rfkill), not the stable `Powered` flag, and even that doesn't split hard-block from soft-block.

dlech: I don't think macs have a hardware switch. And I think `off-blocked` in BlueZ sounds like `RadioState.Disabled` on Windows. Since it is still experimental though, I guess we should just put this in an issue to look at again in the future.

Vodur: opened an issue: #1996

dlech:
Some suggestions from AI review. These sounds like reasonable changes to me.

### `bleak/backends/bluezdbus/adapter.py`, the `manager._properties` block

Reaching into `manager._properties` from here (with a pyright ignore) isn't great. It also means the explicit-adapter path skips the `Roles` check, so `NO_BLE_CENTRAL_ROLE` can only ever be raised for the default adapter.

I'd rather add a public method on `BlueZManager` next to `get_default_adapter()` and keep `get()` trivial:

```python
    @classmethod
    @override
    async def get(cls, *, bluez: BlueZAdapterArgs = {}, **kwargs: Any) -> Self:
        adapter = bluez.get("adapter")
        # Both of these raise BleakBluetoothNotAvailableError if the adapter
        # is not usable.
        adapter_path = (
            manager.get_adapter(adapter) if adapter else manager.get_default_adapter()
        )
        return cls(adapter_path)
```

(The `defs` and `bleak.exc` imports added to this file can then go away again.)

And in `manager.py`, right after `get_default_adapter()`. This reuses the existing `_check_adapter()` for the "no such adapter" case, so a typo in the adapter name stays a plain `BleakError` like it already is for the scanner and client, rather than becoming `NO_BLUETOOTH`:

```python
    def get_adapter(self, adapter: str) -> str:
        """
        Gets the D-Bus object path of a specific Bluetooth adapter and checks
        that it is usable as a BLE central.

        Args:
            adapter: The name of the adapter, e.g. "hci0".

        Returns:
            The D-Bus object path of the adapter, i.e. "/org/bluez/hciX".

        Raises:
            BleakError:
                if there is no adapter with the given name
            BleakBluetoothNotAvailableError:
                if the adapter does not support the BLE 'central' role or is
                not powered on
        """
        adapter_path = f"/org/bluez/{adapter}"

        self._check_adapter(adapter_path)

        props = cast(
            defs.Adapter1, self._properties[adapter_path][defs.ADAPTER_INTERFACE]
        )

        if "central" not in props["Roles"]:
            raise BleakBluetoothNotAvailableError(
                f"Bluetooth adapter '{adapter}' does not support the BLE 'central' role.",
                BleakBluetoothNotAvailableReason.NO_BLE_CENTRAL_ROLE,
            )

        if not props["Powered"]:
            raise BleakBluetoothNotAvailableError(
                f"Bluetooth adapter '{adapter}' is not powered on. Turn on Bluetooth and try again.",
                BleakBluetoothNotAvailableReason.POWERED_OFF,
            )

        return adapter_path
```

### `tests/integration/test_adapter.py`, `test_get_raises_when_powered_off`

I'd like to avoid shelling out to `bluetoothctl` and the fixed sleeps here. We already talk to BlueZ over D-Bus in `tests/integration/bluez_controller.py` (`power_on_controller()`), so the same call with `Powered=False` does the job without depending on an external tool.

Two other things this fixes:

- `bluetoothctl power off` acts on whatever bluetoothctl thinks the default controller is. It's safer to ask Bleak's manager for `get_default_adapter()` first and power off that path, so we know we're toggling the adapter that `get()` is going to look at.
- The "must be the last test in the module" constraint goes away if the test waits until `get()` raises (and, in `finally`, until it succeeds again) instead of sleeping a fixed second.

In `bluez_controller.py`, generalize the existing helper:

```python
async def set_controller_powered(
    bus: MessageBus, adapter_path: str, powered: bool, timeout: float = 2.0
) -> None:
    """
    Set the ``Powered`` property of a Bluetooth controller via D-Bus.

    It may take some time for the adapter to fully configure itself after
    powering on. So we make multiple attempts until timeout is reached.
    """
    async with async_timeout(timeout):
        while True:
            try:
                reply = await bus.call(
                    Message(
                        path=adapter_path,
                        member="Set",
                        signature="ssv",
                        body=[defs.ADAPTER_INTERFACE, "Powered", Variant("b", powered)],
                    )
                )
                assert_reply(reply)
                return
            except Exception as e:
                logger.debug(
                    "Failed to set Powered=%s on adapter at %s: %s",
                    powered,
                    adapter_path,
                    e,
                )
                await asyncio.sleep(0.1)


async def power_on_controller(
    bus: MessageBus, adapter_path: str, timeout: float = 2.0
) -> None:
    """
    Power on a Bluetooth controller via D-Bus.
    """
    await set_controller_powered(bus, adapter_path, True, timeout)
```

Then the test becomes:

```python
from bleak._compat import timeout as async_timeout
from bleak.backends import BleakBackend, get_default_backend

...

@pytest.mark.asyncio(loop_scope="module")
@pytest.mark.usefixtures("hci_transport")
@pytest.mark.skipif(
    get_default_backend() != BleakBackend.BLUEZ_DBUS,
    reason="powering the adapter off is only automated on BlueZ",
)
async def test_get_raises_when_powered_off() -> None:
    """``BleakAdapter.get()`` raises ``BleakBluetoothNotAvailableError`` with
    reason ``POWERED_OFF`` when the local Bluetooth adapter is not powered on.
    """
    from bleak.backends.bluezdbus.manager import get_global_bluez_manager
    from tests.integration.bluez_controller import (
        open_message_bus,
        set_controller_powered,
    )

    # Find out which adapter BleakAdapter.get() is going to pick so that we
    # power off the right one.
    adapter_path = manager.get_default_adapter()

    async with open_message_bus() as bus:
        try:
            await set_controller_powered(bus, adapter_path, False)

            # The PropertiesChanged signal takes a moment to reach Bleak's
            # manager, so retry until get() sees the new state.
            async with async_timeout(2.0):
                while True:
                    try:
                        await BleakAdapter.get()
                    except BleakBluetoothNotAvailableError as ex:
                        assert ex.reason == BleakBluetoothNotAvailableReason.POWERED_OFF
                        break
                    await asyncio.sleep(0.05)
        finally:
            await set_controller_powered(bus, adapter_path, True)

            # Don't leave the module until later tests can use the adapter again.
            async with async_timeout(2.0):
                while True:
                    try:
                        await BleakAdapter.get()
                        break
                    except BleakBluetoothNotAvailableError:
                        await asyncio.sleep(0.05)
```

The `skipif` is needed because the D-Bus helper is Linux-only; the current version would fail with `FileNotFoundError` for `bluetoothctl` if someone runs the integration tests on macOS with a real transport.

### `bleak/backends/winrt/scanner.py`, `start()`

Nit: `assert_mta()` is now called here and again inside `BleakAdapterWinRT.get()`. The one in `start()` can go.

PR #1990, opened 2026-05-29
Add BLE pairing agent (Windows + Linux)

JPHutchins:
Continues #1864 (cc @tgagneret-embedded) — this branch keeps the original pairing-callbacks commit and adds the WinRT implementation, a sum-type rework of the API, full test coverage, and review fixes. Builds on #1133 and #1100.

## What

`bleak.agent` exposes `PairingCallbacks` — optional async handlers for `ConfirmPasskey` (Numeric Comparison) and `RequestPasskey` / `DisplayPasskey` (Passkey Entry, both directions). Which callbacks are provided determines the advertised `IOCapability`, which drives the negotiated Security Manager ceremony. With no callbacks, Just Works is accepted automatically (unchanged behavior).

- **WinRT** — a deferral-bridged `PairingRequested` handler dispatches the ceremony with an exhaustive `match`.
- **BlueZ** — `org.bluez.Agent1` driven by the callbacks; the advertised capability is derived from them; the device is trusted after a successful pair.
- `timeout` / `disconnected_callback` / `pairing_callbacks` are now explicit, typed parameters on every backend (no more `kwargs.get`).
- Pairing failures raise `BleakPairingCancelledError` / `BleakPairingFailedError`.
- `examples/pairing_agent.py` demonstrates the API.

## Testing

- Unit tests: capability derivation and pairing-request dispatch (WinRT and BlueZ).
- `tests/integration/test_client_pairing.py`: bumble integration tests for all four LE Secure Connections ceremonies + unpair, runnable on Linux VHCI (no hardware) or over a USB HCI controller.
- mypy + pyright strict, flake8 / black / isort clean.
- The BlueZ `Trusted`-after-pair change addresses the review comment on #1864.

Validated on real hardware (Windows central + nRF52840 USB HCI bumble peripheral):

```
$ uv run --group test pytest tests/integration/test_client_pairing.py --bleak-hci-transport=usb:0 -v --no-cov

tests/integration/test_client_pairing.py::test_pair_just_works PASSED                        [ 16%]
tests/integration/test_client_pairing.py::test_pair_numeric_comparison PASSED                [ 33%]
tests/integration/test_client_pairing.py::test_pair_passkey_entry_central_inputs PASSED      [ 50%]
tests/integration/test_client_pairing.py::test_pair_passkey_entry_central_displays PASSED    [ 66%]
tests/integration/test_client_pairing.py::test_unpair PASSED                                 [ 83%]
[...]

PR #1991, opened 2026-06-01
backends/bluezdbus: use dbus-fast typed annotations

JPHutchins:
Split out of #1990, addressing the review request to use dbus-fast's typed annotations instead of the string-signature shims.

## What

Migrates `AdvertisementMonitor`'s `org.bluez.AdvertisementMonitor1` interface from `@no_type_check` + string signatures to dbus-fast's typed annotations:

- Method/property signatures use the `dbus_fast.annotations` aliases — `DBusObjectPath`, `DBusStr`, `DBusInt16`, `DBusUInt16`, and `Annotated[..., DBusSignature("a(yyay)")]` — instead of the `"o"` / `"s"` / `"a(yyay)"` string shims.
- Uses the typed `@dbus_method` decorator (not `@method`), with explicit `-> None` return types on the void methods.
- Drops the `@no_type_check` workaround for [python/mypy#6583](https://github.com/python/mypy/issues/6583) and the `# noqa: F821` forward-ref shims.
- Disabled property stubs `raise NotImplementedError` instead of `...` — they are type-checked now, so an empty body would fail mypy.
- `Patterns` is typed `list[_OrPatternLike]` (the `(yyay)` struct is `(int, int, bytes)`), and the or-patterns are stored as-is via `list(or_patterns)`. The old `# dbus_fast marshaling requires list instead of tuple` conversion is removed — dbus-fast marshals tuples/`NamedTuple`s for STRUCTs just as well as lists.
- Bumps `dbus-fast>=4.0.0` on Linux (no upper cap) — 4.0 is the first release that ships the annotation aliases and `dbus_method`.

## Why

This is the prerequisite that lets the BlueZ backend (and the pairing agent built on it) be fully type-checked without per-method opt-outs.

## Testing

- mypy + pyright (strict) clean, native and Linux-simulated.
- No change to the marshaled D-Bus output; the BlueZ VHCI integration suite exercises the advertisement monitor.

### DIFF
--- a/.github/alpine-vm/alpine-vm-setup.sh
+++ b/.github/alpine-vm/alpine-vm-setup.sh
@@ -26,6 +26,19 @@ step 'Enable bluetooth and dbus services'
 rc-update add bluetooth default
 rc-update add dbus default
 
+# Enable BlueZ experimental D-Bus interfaces. The AdvertisementMonitorManager1
+# interface (used by passive scanning with or_patterns) is gated behind this
+# flag, so RegisterMonitor is unavailable without it.
+step 'Enable BlueZ experimental interfaces'
+mkdir -p /etc/bluetooth
+if [ -f /etc/bluetooth/main.conf ] && grep -qiE '^[[:space:]]*#?[[:space:]]*Experimental' /etc/bluetooth/main.conf; then
+	sed -i -E 's/^[[:space:]]*#?[[:space:]]*Experimental.*/Experimental = true/I' /etc/bluetooth/main.conf
+elif [ -f /etc/bluetooth/main.conf ] && grep -qE '^\[General\]' /etc/bluetooth/main.conf; then
+	sed -i '/^\[General\]/a Experimental = true' /etc/bluetooth/main.conf
+else
+	printf '[General]\nExperimental = true\n' >> /etc/bluetooth/main.conf
+fi
+
 # Configure vhci kernel module to load at boot
 step 'Configure VHCI kernel module for boot'
 echo "hci_vhci" >> /etc/modules
--- a/bleak/backends/bluezdbus/advertisement_monitor.py
+++ b/bleak/backends/bluezdbus/advertisement_monitor.py
@@ -15,11 +15,18 @@ if TYPE_CHECKING:
 
 import logging
 from collections.abc import Iterable
-from typing import Any, no_type_check
+from typing import Annotated, Any
 from warnings import warn
 
 from dbus_fast import PropertyAccess
-from dbus_fast.service import ServiceInterface, dbus_property, method
+from dbus_fast.annotations import (
+    DBusInt16,
+    DBusObjectPath,
+    DBusSignature,
+    DBusStr,
+    DBusUInt16,
+)
+from dbus_fast.service import ServiceInterface, dbus_method, dbus_property
 
 from bleak.args.bluez import OrPattern as _OrPattern
 from bleak.args.bluez import OrPatternLike as _OrPatternLike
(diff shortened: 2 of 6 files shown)


bleak/backends/bluezdbus/advertisement_monitor.py
```diff
@@ -69,7 +76,7 @@ def __init__(
         """
         super().__init__(defs.ADVERTISEMENT_MONITOR_INTERFACE)
         # dbus_fast marshaling requires list instead of tuple
-        self._or_patterns = [list(p) for p in or_patterns]
+        self._or_patterns: list[Any] = [list(p) for p in or_patterns]
```
dlech: Why do we need to change from the inferred type here? (I assume it would be `list[_OrPatternLike`)

bleak/backends/bluezdbus/advertisement_monitor.py
```diff
@@ -79,53 +86,41 @@ def Release(self):
...
-    @no_type_check
-    def RSSISamplingPeriod(self) -> "q":  # noqa: F821
-        ...
+    def RSSISamplingPeriod(self) -> DBusUInt16:
+        raise NotImplementedError
 
     @dbus_property(PropertyAccess.READ)
-    @no_type_check
-    def Patterns(self) -> "a(yyay)":  # noqa: F821
+    def Patterns(self) -> Annotated[list[Any], DBusSignature("a(yyay)")]:
```
dlech:
```suggestion
    def Patterns(self) -> Annotated[list[tuple[bytes, bytes, list[bytes]], DBusSignature("a(yyay)")]:
```

I think? this is the actual type, but I didn't spend too much time on it.

pyproject.toml
```diff
@@ -29,7 +29,7 @@ dependencies = [
     "winrt-Windows.Foundation>=3.1; platform_system=='Windows'",
     "winrt-Windows.Foundation.Collections>=3.1; platform_system=='Windows'",
     "winrt-Windows.Storage.Streams>=3.1; platform_system=='Windows'",
-    "dbus-fast>=1.83.0; platform_system=='Linux'",
+    "dbus-fast>=4.0.0,<5.0.0; platform_system=='Linux'",
```
dlech:
```suggestion
    "dbus-fast>=4.0.0; platform_system=='Linux'",
```

Upper limits on versions cause problems. Major version changes don't always break Bleak.

bleak/backends/bluezdbus/advertisement_monitor.py
JPHutchins: I will investigate

pyproject.toml
JPHutchins: Yeah, I need to investigate 5s breaking changes, whether there's a 6, etc! Hopefully can cap it at a future breaking change.

bleak/backends/bluezdbus/advertisement_monitor.py
JPHutchins: We don't. I'm very strict with Opus, but it snuck this slop in, my mistake.

pyproject.toml
JPHutchins: I have moved to 6, which is unreleased, so that we don't have builds failing in the wild before addressing 6's breaking changes when it's released.

dlech: I would prefer to remove the upper limit completely. It is much more likely to be OK for many major versions than not.

JPHutchins: fixed

bleak/backends/bluezdbus/advertisement_monitor.py
```diff
@@ -68,8 +75,7 @@ def __init__(
                 List of or patterns that will be returned by the ``Patterns`` property.
         """
         super().__init__(defs.ADVERTISEMENT_MONITOR_INTERFACE)
-        # dbus_fast marshaling requires list instead of tuple
-        self._or_patterns = [list(p) for p in or_patterns]
+        self._or_patterns = list(or_patterns)
 
     @method()
```
dlech: We could save this for later, but it would be nice to change all `@method` to `@dbus_method` now that we have bumped the minimum dbus-fast version.

bleak/backends/bluezdbus/advertisement_monitor.py
```diff
@@ -79,53 +85,41 @@ def Release(self):
     def Activate(self):
         logger.debug("Activate")
 
-    # REVISIT: mypy is broke, so we have to add redundant @no_type_check
-    # https://github.com/python/mypy/issues/6583
-
     @method()
-    @no_type_check
-    def DeviceFound(self, device: "o"):  # noqa: F821
+    def DeviceFound(self, device: DBusObjectPath):
```
dlech:
```suggestion
    def DeviceFound(self, device: DBusObjectPath) -> None:
```

Probably best to be explicit on the return types.

bleak/backends/bluezdbus/advertisement_monitor.py
JPHutchins: fixed

bleak/backends/bluezdbus/advertisement_monitor.py
JPHutchins: fixed

bleak/backends/bluezdbus/advertisement_monitor.py
```diff
@@ -68,64 +75,51 @@ def __init__(
                 List of or patterns that will be returned by the ``Patterns`` property.
         """
         super().__init__(defs.ADVERTISEMENT_MONITOR_INTERFACE)
-        # dbus_fast marshaling requires list instead of tuple
-        self._or_patterns = [list(p) for p in or_patterns]
+        self._or_patterns = list(or_patterns)
```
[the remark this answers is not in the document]
dlech: We should probably add a test for this. I'm pretty sure dbus-fast allows tuples now for D-Bus tuples (the `()` in the signature), so we don't have to convert to list.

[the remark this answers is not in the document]
JPHutchins: Tests added.

tests/backends/bluezdbus/test_advertisement_monitor.py
```diff
@@ -0,0 +1,103 @@
...
+
+import pytest
+
+if sys.platform != "linux":
+    pytest.skip("backend only available on Linux", allow_module_level=True)
+    assert False  # HACK: work around pyright bug
+
+from collections.abc import Iterable
+
+from dbus_fast._private.marshaller import Marshaller
```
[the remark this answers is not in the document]
dlech: Actually, I agree that we should avoid using `dbus_fast._private` directly. We just need to test that Bleak works, not the implementation details of dbus-fast.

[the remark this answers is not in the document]
dlech: The test should actually be doing a passive scan with Bleak with the `or_patterns` as an integration test.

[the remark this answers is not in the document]
JPHutchins: Good idea, I've replaced the unit tests with integration tests. They are passing in all python versions in the alpine VM:

```
tests/integration/test_scanner.py::test_passive_scan_or_patterns[named_tuple] PASSED [ 91%]
tests/integration/test_scanner.py::test_passive_scan_or_patterns[tuple] PASSED [ 92%]
```

tests/integration/test_scanner.py
```diff
@@ -8,13 +8,23 @@
...
 from bleak.uuids import normalize_uuid_str
 from tests.integration.conftest import configure_and_power_on_bumble_peripheral
 
 DEFAULT_TIMEOUT = 5.0
 
+# The default bumble peripheral advertises Flags == 0x06
+# (LE General Discoverable | BR/EDR not supported), so this pattern matches it.
+_FLAGS_NAMED_TUPLE: list[OrPatternLike] = [
+    OrPattern(0, AdvertisementDataType.FLAGS, b"\x06")
+]
```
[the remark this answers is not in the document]
dlech: CI passed, so seems OK.

.github/alpine-vm/alpine-vm-setup.sh
```diff
@@ -26,6 +26,19 @@ step 'Enable bluetooth and dbus services'
...
 
+# Enable BlueZ experimental D-Bus interfaces. The AdvertisementMonitorManager1
+# interface (used by passive scanning with or_patterns) is gated behind this
+# flag, so RegisterMonitor is unavailable without it.
+step 'Enable BlueZ experimental interfaces'
+mkdir -p /etc/bluetooth
+if [ -f /etc/bluetooth/main.conf ] && grep -qiE '^[[:space:]]*#?[[:space:]]*Experimental' /etc/bluetooth/main.conf; then
+	sed -i -E 's/^[[:space:]]*#?[[:space:]]*Experimental.*/Experimental = true/I' /etc/bluetooth/main.conf
+elif [ -f /etc/bluetooth/main.conf ] && grep -qE '^\[General\]' /etc/bluetooth/main.conf; then
+	sed -i '/^\[General\]/a Experimental = true' /etc/bluetooth/main.conf
```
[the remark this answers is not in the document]
dlech: CI passed, so seems OK.

dlech:
Thanks!

PR #1992, opened 2026-06-01
bleak/pairing: add the pairing callback API

JPHutchins:
Split out of #1990 — the backend-agnostic foundation, with no wiring yet so it can be reviewed on its own.

## What

Adds a new `bleak.pairing` module:

- `PairingCallbacks` — a `NamedTuple` of optional async handlers: `confirm(device, passkey)` (Numeric Comparison) and `request_passkey(device)` / `display_passkey(device, passkey)` (Passkey Entry, both directions).
- `IOCapability` — the advertised Security Manager I/O capability.
- `io_capability(callbacks)` — derives the capability from which callbacks are set (e.g. `confirm` implies a display).
- `MAX_PASSKEY` — the 6-digit passkey ceiling.

Which callbacks are provided drives the negotiated pairing ceremony; with none, Just Works is accepted automatically (unchanged behavior). The backends consume this in follow-up PRs.

## Testing

- Unit tests covering `io_capability()` across every callback combination.
- mypy + pyright (strict) clean.

### DIFF
--- a/bleak/pairing.py
+++ b/bleak/pairing.py
@@ -1,0 +1,153 @@
+"""
+Pairing
+-------
+
+Types for participating in the BLE pairing ceremony.
+
+An application takes part in pairing by supplying an object that implements one or
+more of the capability protocols in this module -- :class:`SupportsConfirm`,
+:class:`SupportsRequestPasskey`, and :class:`SupportsDisplayPasskey`. Only the
+methods for the ceremonies the application can perform are implemented; the
+*absence* of a method means the application cannot take part in that ceremony.
+Which methods are present determines the input/output capability that Bleak
+advertises to the peer, and therefore which `Security Manager`_ pairing method is
+negotiated:
+
+============================================  =====================  =======================
+Implemented methods                           Capability             Ceremony
+============================================  =====================  =======================
+(none)                                        ``NoInputNoOutput``    Just Works
+``confirm``                                   ``DisplayYesNo``       Numeric Comparison
+``request_passkey``                           ``KeyboardOnly``       Passkey Entry (input)
+``display_passkey``                           ``DisplayOnly``        Passkey Entry (output)
+``request_passkey`` + ``display_passkey``     ``KeyboardDisplay``    Passkey Entry (either)
+============================================  =====================  =======================
+
+A ``confirm`` method also implies a display, since Numeric Comparison shows the
+value being compared, so combining it with another method upgrades the advertised
+capability (e.g. ``confirm`` + ``display_passkey`` is ``DisplayYesNo``, while
+``confirm`` + ``request_passkey`` is ``KeyboardDisplay``).
+
+There are two equivalent ways to supply an implementation:
+
+* any object -- a :class:`~typing.NamedTuple` is a natural fit -- that structurally
+  implements the chosen methods, or
+* a subclass of the matching protocols -- :class:`SupportsConfirm`,
+  :class:`SupportsRequestPasskey`, :class:`SupportsDisplayPasskey`. Their methods are
+  abstract, so the type checker *and* the interpreter require you to implement every
+  capability you inherit (merely inheriting is not enough); state may be carried on
+  ``self``, and the protocols may be combined freely by multiple inheritance.
+
+Because detection is structural (see :func:`io_capability`), implement *only* the
+methods for the ceremonies you support and omit the rest; a method that is present
+but does not work (for example, one that always raises) still advertises the
+capability, which is almost never what you want.
+
+The IO capabilities and their mapping to a pairing method follow the `Security
+Manager`_ specification: Bluetooth Core Specification v5.4, Vol 3, Part H, Section
+2.3.2 (IO capabilities) and Section 2.3.5.1 (mapping of IO capabilities to the key
+generation method).
+
+.. _Security Manager: https://www.bluetooth.com/specifications/specs/core-specification/
+"""
+
+from abc import abstractmethod
+from enum import Enum, auto
+from typing import Final, Protocol, TypeAlias, runtime_checkable
+
+from bleak.backends.device import BLEDevice
+
+MAX_PASSKEY: Final = 999999
+"""The largest valid BLE passkey; passkeys are six decimal digits (000000-999999)."""
+
+
+@runtime_checkable
+class SupportsConfirm(Protocol):
+    """Capability to take part in Numeric Comparison (``DisplayYesNo``)."""
+
+    @abstractmethod
+    async def confirm(self, device: BLEDevice, passkey: int) -> bool:
+        """Numeric Comparison: receive *device* and the *passkey* shown on both
+        devices; return ``True`` to accept the pairing or ``False`` to reject it."""
+
+
+@runtime_checkable
+class SupportsRequestPasskey(Protocol):
+    """Capability to enter a passkey for Passkey Entry (``KeyboardOnly``)."""
+
+    @abstractmethod
+    async def request_passkey(self, device: BLEDevice) -> int | None:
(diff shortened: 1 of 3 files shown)


bleak/pairing.py
```diff
@@ -0,0 +1,102 @@
...
+``request_passkey``                           ``KeyboardOnly``       Passkey Entry (input)
+``display_passkey``                           ``DisplayOnly``        Passkey Entry (output)
+``request_passkey`` + ``display_passkey``     ``KeyboardDisplay``    Passkey Entry (either)
+============================================  =====================  =======================
+
+A ``confirm`` handler also implies a display, since Numeric Comparison shows the
+value being compared, so combining it with another handler upgrades the
+advertised capability (e.g. ``confirm`` + ``display_passkey`` is ``DisplayYesNo``).
+
+.. _Security Manager: https://www.bluetooth.com/specifications/specs/core-specification/
```
dlech, line 51: Can we get the specification version and section number that was used as a reference?

bleak/pairing.py
```diff
@@ -0,0 +1,102 @@
...
+from enum import Enum, auto
+from typing import Final, NamedTuple
+
+from bleak.backends.device import BLEDevice
+
+MAX_PASSKEY: Final = 999999
+"""The largest valid BLE passkey; passkeys are six decimal digits (000000-999999)."""
+
+
+class PairingCallbacks(NamedTuple):
```
dlech: Maybe this was explained in the other PR, but why is this a `NamedTuple` instead of a regular class?

I think a regular class would be nice since it would have named parameters.

JPHutchins: NamedTuple has named fields, it's a struct. This is better than a class because it's fit for the specific purpose: a collection of optional functions.

The other option would be a frozen dataclass, or both + a protocol to allow users to create their own object. I believe that would allow users to define a class with static methods and properties to fulfill the protocol, but that's very clunky compared to a struct.

JPHutchins: Re: the specific question, NamedTuples do support named parameters:

```python
# Passkey Entry (input) — type in the value the peer is displaying
async def request_passkey(device: BLEDevice) -> int | None:
    raw = input(f"Enter the passkey shown on {device.name}: ")
    return int(raw) if raw else None

callbacks = PairingCallbacks(request_passkey=request_passkey)
```

```python
# every field passed by name, order irrelevant
callbacks = PairingCallbacks(
    display_passkey=display_passkey,
    request_passkey=request_passkey,
)
```

Usage
```python
>>> callbacks = PairingCallbacks(confirm=confirm)
>>> callbacks.confirm is confirm
True
>>> callbacks.request_passkey is None      # unset fields default to None
True
>>> io_capability(callbacks)
<IOCapability.DISPLAY_YES_NO: 2>
```

bleak/pairing.py
```diff
@@ -0,0 +1,102 @@
...
+    can_confirm = callbacks.confirm is not None
+    can_display = callbacks.display_passkey is not None or can_confirm
+    if can_input and can_display:
+        return IOCapability.KEYBOARD_DISPLAY
+    if can_input:
+        return IOCapability.KEYBOARD_ONLY
+    if can_display and can_confirm:
+        return IOCapability.DISPLAY_YES_NO
+    if can_display:
+        return IOCapability.DISPLAY_ONLY
```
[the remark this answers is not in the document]
JPHutchins: Addressed in the latest push: simplified to `if can_confirm:` — since `can_display` already includes `can_confirm`, the extra conjunction was redundant.

bleak/pairing.py
```diff
@@ -0,0 +1,102 @@
...
+    DISPLAY_YES_NO = auto()
+    """Can display a six-digit value and has two buttons the user can map to yes and no."""
+
+    KEYBOARD_ONLY = auto()
+    """Can enter the digits 0-9 and answer yes/no, but cannot display a value."""
+
+    DISPLAY_ONLY = auto()
+    """Can display a six-digit value but has no input to enter one or answer yes/no."""
+
+    KEYBOARD_DISPLAY = auto()
```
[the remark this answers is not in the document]
JPHutchins: Addressed in the latest push: the `IOCapability` docstring now states these are the five IO capabilities from the Bluetooth Core Specification (Vol 3, Part H, §2.3.2), named in Pythonic form, and notes that BlueZ spells them e.g. `DisplayYesNo`.

bleak/pairing.py
```diff
@@ -0,0 +1,102 @@
...
+
+class PairingCallbacks(NamedTuple):
+    """Application-provided handlers for the BLE pairing ceremony.
+
+    Each field is an optional asynchronous callback invoked when the peer
+    initiates the corresponding ceremony. A field left as ``None`` means the
+    application cannot take part in that ceremony; the set of provided callbacks
+    determines the advertised :class:`IOCapability` (see :func:`io_capability`).
+    With no callbacks the Just Works ceremony is used and accepted automatically.
+    """
```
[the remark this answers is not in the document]
JPHutchins: Addressed in the latest push: reworded to avoid guaranteeing acceptance here — "With no callbacks the advertised capability is `NoInputNoOutput`, which selects the Just Works ceremony on backends that support pairing."

bleak/pairing.py
dlech: I do like the suggestion of a protocol, but I don't think that works technically with optional methods.

`dataclass` doesn't seem quite right since we don't care about equality and things like that (we are just dealing with functions).

Re: the parameters, I meant the parameters of the `Callable` types.

Anyway, in general I do tend to prefer functional style programming over object oriented, so in that regard the NamedTuple is nice. I see pros and cons each way and not a clear "best" option.

JPHutchins: > Re: the parameters, I meant the parameters of the Callable types.

Aw, good point! I'm going to experiment with a protocol union that allows users to impl ABCs that satisfy some subset of the callbacks. The lib code will be slightly more complex, but the user code will be more self-documenting.

bleak/pairing.py
```diff
@@ -0,0 +1,170 @@
...
+from typing import Final, Protocol, TypeAlias, runtime_checkable
+
+from bleak.backends.device import BLEDevice
+
+MAX_PASSKEY: Final = 999999
+"""The largest valid BLE passkey; passkeys are six decimal digits (000000-999999)."""
+
+
+@runtime_checkable
+class CanConfirm(Protocol):
```
dlech:
```suggestion
class SupportsConfirm(Protocol):
```

There seems to be a convention already in Python to use `Supports` for protocols.

bleak/pairing.py
```diff
@@ -0,0 +1,170 @@
...
+
+This is the type backends accept (as ``PairingCallbacks | None``, where ``None``
+means no callbacks and selects Just Works). Satisfy it with any object -- a
+:class:`~typing.NamedTuple` is a natural fit -- that defines the chosen methods, or
+by subclassing :class:`ConfirmBase`, :class:`RequestPasskeyBase`, and/or
+:class:`DisplayPasskeyBase`.
+"""
+
+
+class ConfirmBase(ABC, CanConfirm):
```
dlech: Do we really need these ABCs? It seems like we could just make the protocol `@runtime_checkable` and inherit from that directly.

JPHutchins: Investigating!

JPHutchins: Seem to be narrowing on something elegant: combining abstractmethod with the runtime_checkable. Appears to be PEP 544 compliant.

JPHutchins: Pushed that version, I like it a lot.

bleak/pairing.py
JPHutchins: Agreed, fixed.

PR #1993, opened 2026-06-01
tests: cover existing pair() and unpair() behavior

JPHutchins:
Split out of #1990 — a regression net for the *current* `pair()` / `unpair()` behavior, landed first so the pairing-agent work that follows has coverage to change against.

## What

Adds unit and integration coverage for existing behavior, with no production changes:

- `tests/backends/winrt/test_pairing.py`, `tests/backends/bluezdbus/test_pairing.py` — unit coverage of the existing `pair()` / `unpair()` paths.
- `tests/integration/test_client_pairing.py` — bumble integration tests: `unpair()` of a connected, never-paired device; CoreBluetooth raises `NotImplementedError`.

## Testing

- Integration tests run on the BlueZ VHCI leg (bumble) in CI — no hardware required.
- CoreBluetooth cases are `skipif`-guarded.

### DIFF
--- a/tests/integration/test_client_pairing.py
+++ b/tests/integration/test_client_pairing.py
@@ -10,10 +10,26 @@ from tests.integration.conftest import (
 
 
 @pytest.mark.skipif(
+    get_default_backend() == BleakBackend.CORE_BLUETOOTH,
+    reason="unpair is not available on CoreBluetooth",
+)
+async def test_unpair_unpaired_device(bumble_peripheral: Device) -> None:
+    """unpair() after a connect/disconnect of a never-paired device does not raise."""
+    await configure_and_power_on_bumble_peripheral(bumble_peripheral)
+
+    device = await find_ble_device(bumble_peripheral)
+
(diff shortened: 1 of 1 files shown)


dlech:
I don't consider the tests with mocks very helpful since it is quite difficult mock how OS-specific Bluetooth works and it tends to change over time.

I would rather have just have integration tests for everything.

dlech:
For example, we should have an integration test where the peripheral returns the error code “Pairing Not Supported” to make sure we have a common exception raised on all platforms.

We might also want to have integrations tests for both peripherals that support "LE legacy pairing" (temporary key + short term key) and ones that support "LE Secure Connections" (long term key) to make sure the OS handles both (parameterized tests). Hopefully the OS handles the difference between these transparently, but one never knows.

I probably need to do some more reading to see if there are any other interesting cases we would want to test.

JPHutchins:
I will remove the mocked tests and leave in the integration test.

PR #1994, opened 2026-06-01
backends: make timeout and disconnected_callback explicit kwargs

JPHutchins:
Split out of #1990.

## What

Makes the previously-implicit `timeout` / `disconnected_callback` contract explicit on the backend clients instead of threading them through `**kwargs`:

- `BaseBleakClient.__init__` takes `timeout` and `disconnected_callback` as explicit keyword-only parameters.
- The WinRT, BlueZ, CoreBluetooth, and p4android backends accept and forward them explicitly.
- Corrects the `disconnected_callback` docstring: the backend-level callable takes **no** arguments (signature is `Callable[[], None]`), matching how backends invoke it — the previous "must take one argument (the client)" text described the public facade, not this interface.

This makes the constructor signatures self-documenting and is the groundwork for adding `pairing_callbacks` as a sibling explicit parameter in a follow-up.

## Note for review

This also drops the `branches: [master, develop]` filter from the `pull_request:` trigger in `build_and_test.yml` and `format_and_lint.yml` so CI runs on every PR (it was needed to run CI on the stacked split PRs). If you'd rather keep that filter on upstream, I'm happy to pull those two lines back out of this PR.

## Testing

- mypy + pyright (strict) clean, native and Linux-simulated.
- No behavior change; existing connect/disconnect coverage exercises the forwarded parameters.

### DIFF
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -66,10 +66,16 @@ class BleakClientBlueZDBus(BaseBleakClient):
         address_or_ble_device: Union[BLEDevice, str],
         services: Optional[set[str]] = None,
         *,
+        disconnected_callback: Callable[[], None] | None,
+        timeout: float,
         bluez: BlueZClientArgs,
         **kwargs: Any,
     ):
-        super().__init__(address_or_ble_device, **kwargs)
+        super().__init__(
+            address_or_ble_device,
+            disconnected_callback=disconnected_callback,
+            timeout=timeout,
+        )
 
         self._adapter = bluez.get("adapter")
         self._device_path: Optional[str]
--- a/bleak/backends/client.py
+++ b/bleak/backends/client.py
@@ -4,7 +4,7 @@ Base class for backend clients.
 """
 import abc
 from collections.abc import Callable
-from typing import Any, Optional, Union
+from typing import Any, Optional
 
 from bleak.args import SizedBuffer
 from bleak.backends import BleakBackend, get_default_backend
@@ -28,11 +28,18 @@ class BaseBleakClient(abc.ABC):
     Keyword Args:
         timeout (float): Timeout for required ``discover`` call.
         disconnected_callback (callable): Callback that will be scheduled in the
-            event loop when the client is disconnected. The callable must take one
-            argument, which will be this client object.
+            event loop when the client is disconnected. The callable takes no
+            arguments.
     """
 
-    def __init__(self, address_or_ble_device: Union[BLEDevice, str], **kwargs: Any):
+    def __init__(
+        self,
+        address_or_ble_device: BLEDevice | str,
+        *,
+        disconnected_callback: Callable[[], None] | None,
+        timeout: float,
+        **kwargs: Any,
+    ) -> None:
         if isinstance(address_or_ble_device, BLEDevice):
             self.address = address_or_ble_device.address
         else:
@@ -40,10 +47,8 @@ class BaseBleakClient(abc.ABC):
 
         self.services: Optional[BleakGATTServiceCollection] = None
 
-        self._timeout = kwargs["timeout"]
-        self._disconnected_callback: Optional[Callable[[], None]] = kwargs.get(
-            "disconnected_callback"
-        )
+        self._timeout = timeout
+        self._disconnected_callback = disconnected_callback
 
     # NB: this is not marked as @abc.abstractmethod because that would break
     # 3rd-party backends. We might change this in the future to make it required.
--- a/bleak/backends/corebluetooth/client.py
+++ b/bleak/backends/corebluetooth/client.py
@@ -13,6 +13,7 @@ if TYPE_CHECKING:
 
 import asyncio
 import logging
+from collections.abc import Callable
 from typing import Any, Optional, Union
 
 from CoreBluetooth import (
@@ -59,9 +60,16 @@ class BleakClientCoreBluetooth(BaseBleakClient):
         self,
         address_or_ble_device: Union[BLEDevice, str],
         services: Optional[set[str]] = None,
+        *,
+        disconnected_callback: Callable[[], None] | None,
+        timeout: float,
         **kwargs: Any,
     ):
-        super().__init__(address_or_ble_device, **kwargs)
+        super().__init__(
+            address_or_ble_device,
+            disconnected_callback=disconnected_callback,
+            timeout=timeout,
+        )
 
         self._peripheral: Optional[CBPeripheral] = None
         self._delegate: Optional[PeripheralDelegate] = None
--- a/bleak/backends/p4android/client.py
+++ b/bleak/backends/p4android/client.py
@@ -13,6 +13,7 @@ import asyncio
 import logging
 import uuid
 import warnings
+from collections.abc import Callable
 from typing import Any, Optional, Union
 
 from android.broadcast import BroadcastReceiver
@@ -46,9 +47,16 @@ class BleakClientP4Android(BaseBleakClient):
         self,
         address_or_ble_device: Union[BLEDevice, str],
         services: Optional[set[uuid.UUID]],
-        **kwargs,
+        *,
+        disconnected_callback: Callable[[], None] | None,
+        timeout: float,
+        **kwargs: Any,
     ):
-        super().__init__(address_or_ble_device, **kwargs)
+        super().__init__(
+            address_or_ble_device,
+            disconnected_callback=disconnected_callback,
+            timeout=timeout,
+        )
         self._requested_services = (
             set(map(defs.UUID.fromString, services)) if services else None
         )
--- a/bleak/backends/winrt/client.py
+++ b/bleak/backends/winrt/client.py
@@ -149,10 +149,16 @@ class BleakClientWinRT(BaseBleakClient):
         address_or_ble_device: Union[BLEDevice, str],
         services: Optional[set[str]] = None,
         *,
+        disconnected_callback: Callable[[], None] | None,
+        timeout: float,
         winrt: _WinRTClientArgs,
         **kwargs: Any,
     ):
-        super().__init__(address_or_ble_device, **kwargs)
+        super().__init__(
+            address_or_ble_device,
+            disconnected_callback=disconnected_callback,
+            timeout=timeout,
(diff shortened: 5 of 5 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -5,7 +5,6 @@ on:
     branches: [master, develop]
     paths: ["bleak/**", "tests/**", ".github/workflows/build_and_test.yml", "pyproject.toml"]
   pull_request:
-    branches: [master, develop]
     paths: ["bleak/**", "tests/**", ".github/workflows/build_and_test.yml", "pyproject.toml"]
```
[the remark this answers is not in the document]
JPHutchins: Generally there's no reason to gate on open source, depending on volume of CI runs.

bleak/backends/client.py
```diff
@@ -28,22 +28,27 @@ class BaseBleakClient(abc.ABC):
...
 
-    def __init__(self, address_or_ble_device: Union[BLEDevice, str], **kwargs: Any):
+    def __init__(
+        self,
+        address_or_ble_device: BLEDevice | str,
+        *,
+        timeout: float,
+        disconnected_callback: Callable[[], None] | None = None,
+        **kwargs: Any,
+    ) -> None:
```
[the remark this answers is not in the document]
JPHutchins: Appears that timeout was always required but not documented at the type level.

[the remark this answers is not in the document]
JPHutchins: `BaseBleakClient` is the backend implementation interface, not the public API, and its `disconnected_callback` was already `Callable[[], None]` before this PR — see the prior body: `self._disconnected_callback: Optional[Callable[[], None]] = kwargs.get("disconnected_callback")`. This change only promotes that existing parameter out of `**kwargs` and corrects the docstring, which incorrectly described a one-argument callable.

The public `bleak.BleakClient(disconnected_callback=...)` contract is unchanged: it still takes `Callable[[BleakClient], None]` and wraps it with `functools.partial(callback, self)` before handing the zero-argument closure to the backend. So there is no public API change and nothing to deprecate.

bleak/backends/corebluetooth/client.py
```diff
@@ -59,9 +60,16 @@ def __init__(
...
+        timeout: float,
+        disconnected_callback: Callable[[], None] | None = None,
         **kwargs: Any,
     ):
-        super().__init__(address_or_ble_device, **kwargs)
+        super().__init__(
+            address_or_ble_device,
+            timeout=timeout,
+            disconnected_callback=disconnected_callback,
+        )
```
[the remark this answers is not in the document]
JPHutchins: Extra keyword arguments were already discarded before this PR: `BaseBleakClient.__init__` itself takes `**kwargs: Any` and uses nothing beyond its named parameters, so forwarding them from the backend would not change the outcome. This PR's scope is just promoting `timeout`/`disconnected_callback` out of `**kwargs`; tightening the remaining catch-all is a separate cleanup.

bleak/backends/client.py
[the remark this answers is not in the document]
dlech: > `BaseBleakClient` is the backend implementation interface, not the public API,

We do have 3rd-party backends though, so it is technically a public API, so still need to be careful about making breaking changes to it. We do have a bit more flexibility though since we know how Bleak calls into the backends, so if we can change it in a way that still works with the current code, it should be OK rather than having to consider all possibilities.

.github/workflows/build_and_test.yml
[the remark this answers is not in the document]
dlech: This is an unrelated change so doesn't belong in this PR.

.github/workflows/format_and_lint.yml
```diff
@@ -4,7 +4,6 @@ on:
   push:
     branches: [master, develop]
   pull_request:
-    branches: [master, develop]
```
dlech: ditto

bleak/backends/client.py
```diff
@@ -28,22 +28,27 @@ class BaseBleakClient(abc.ABC):
...
+            arguments.
     """
 
-    def __init__(self, address_or_ble_device: Union[BLEDevice, str], **kwargs: Any):
+    def __init__(
+        self,
+        address_or_ble_device: BLEDevice | str,
+        *,
+        timeout: float,
+        disconnected_callback: Callable[[], None] | None = None,
```
dlech:
```suggestion
        disconnected_callback: Callable[[], None] | None,
```

Not much use in giving a default since we should always be passing it from the subclass's `__init__` (in fact it could hide a bug if a new backend forgot to pass it along).

bleak/backends/client.py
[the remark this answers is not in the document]
dlech: It looks to me like these changes should be OK in that regard.

bleak/backends/bluezdbus/client.py
```diff
@@ -67,9 +67,15 @@ def __init__(
         services: Optional[set[str]] = None,
         *,
         bluez: BlueZClientArgs,
+        timeout: float,
+        disconnected_callback: Optional[Callable[[], None]] = None,
```
dlech: Let's put these in the same order as `bleak.BleakClient` to keep things consistent.

bleak/backends/client.py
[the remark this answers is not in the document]
JPHutchins: Confirmed.

.github/workflows/build_and_test.yml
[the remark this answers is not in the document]
JPHutchins: fixed

.github/workflows/format_and_lint.yml
JPHutchins: fixed

bleak/backends/client.py
JPHutchins: fixed

bleak/backends/bluezdbus/client.py
JPHutchins: Yes, it is always timeout, disconnected_callback, unless I'm missing one someplace!

dlech: Here these are after `bluez` when they should be before.

dlech: Ah, looks like it is fixed.

bleak/backends/bluezdbus/client.py
```diff
@@ -66,10 +66,16 @@ def __init__(
         address_or_ble_device: Union[BLEDevice, str],
         services: Optional[set[str]] = None,
         *,
+        disconnected_callback: Callable[[], None] | None = None,
```
```suggestion
        disconnected_callback: Callable[[], None] | None,
```

bleak/backends/corebluetooth/client.py
```diff
@@ -59,9 +60,16 @@ def __init__(
         self,
         address_or_ble_device: Union[BLEDevice, str],
         services: Optional[set[str]] = None,
+        *,
+        disconnected_callback: Callable[[], None] | None = None,
```
dlech:
```suggestion
        disconnected_callback: Callable[[], None] | None,
```

bleak/backends/p4android/client.py
```diff
@@ -46,9 +47,16 @@ def __init__(
         self,
         address_or_ble_device: Union[BLEDevice, str],
         services: Optional[set[uuid.UUID]],
-        **kwargs,
+        *,
+        disconnected_callback: Callable[[], None] | None = None,
```
dlech:
```suggestion
        disconnected_callback: Callable[[], None] | None,
```

bleak/backends/winrt/client.py
```diff
@@ -149,10 +149,16 @@ def __init__(
         address_or_ble_device: Union[BLEDevice, str],
         services: Optional[set[str]] = None,
         *,
+        disconnected_callback: Callable[[], None] | None = None,
+        timeout: float,
```
dlech:
```suggestion
        timeout: float,
        disconnected_callback: Callable[[], None] | None,
```

bleak/backends/bluezdbus/client.py
JPHutchins: fixed

bleak/backends/corebluetooth/client.py
JPHutchins: fixed

bleak/backends/p4android/client.py
JPHutchins: fixed

bleak/backends/winrt/client.py
JPHutchins: All now match the ordering of the PlatformBleakClient facade class.

bleak/backends/client.py
```diff
@@ -28,22 +28,27 @@ class BaseBleakClient(abc.ABC):
...
 
-    def __init__(self, address_or_ble_device: Union[BLEDevice, str], **kwargs: Any):
+    def __init__(
+        self,
+        address_or_ble_device: BLEDevice | str,
+        *,
+        disconnected_callback: Callable[[], None] | None,
+        timeout: float,
+        **kwargs: Any,
+    ) -> None:
```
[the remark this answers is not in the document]
dlech: I think this is a hallucination. The backends all use `Callable[[], None]`. The frontend uses `functools.partial` to turn `Callable[[BaseBleakClient], None]` into `Callable[[], None]` before passing it to the backend.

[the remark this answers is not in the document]
JPHutchins: So the user is not given access to the client via this callback? I'll review the design goals.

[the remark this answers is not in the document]
dlech: The user has to have the `BaseBleakClient` parameter in their callback, but the backend does not have to pass it as an arg because we do `functool.partial` to bind it before passing the function to the backend.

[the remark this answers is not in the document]
JPHutchins: Great, I think I'll update the docstring to explain that it's a closure then.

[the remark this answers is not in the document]
JPHutchins: I have reconsidered. I think that this is best left as is instead of leaking usage details of the BleakClient here. It is true that the callback takes no arguments, no reason to lie about it.

dlech:
Thanks!

PR #1997, opened 2026-06-12
backends/corebluetooth: clear notify callbacks on disconnect

### LINKED ISSUE #1969: BleakClient.stop_notify() documentation seems not relevant for MAC OS
The [BleakClient.stop_notify()](https://bleak.readthedocs.io/en/latest/api/client.html#bleak.BleakClient.stop_notify) documentation says:

> Notifications are stopped automatically on disconnect, so this method does not need to be called unless notifications need to be stopped some time before the device disconnects.

But on MAC (Apple M2 Pro with macOS Tahoe v26.4.1), when bleak connects to a BLE peripheral, enables notifications, disconnects then reconnects and try to enable notifications, the following errors is raised:

```
File "/Users/user/Library/Application Support/hatch/env/virtual/project/hoVqGm6Y/project/lib/python3.14/site-packages/bleak/backends/corebluetooth/client.py", line 382, in start_notify
    await self._delegate.start_notifications(
    ...<3 lines>...
    )
  File "/Users/user/Library/Application Support/hatch/env/virtual/project/hoVqGm6Y/project/lib/python3.14/site-packages/bleak/backends/corebluetooth/PeripheralDelegate.py", line 463, in start_notifications
    raise ValueError("Characteristic notifications already started")
ValueError: Characteristic notifications already started
```

The expected behavior would be no Exception raised. By the way, under Linux, everything is fine.

miguel-neto-andrade:
Fixes #1969.

On macOS, calling `start_notify()` after disconnecting and reconnecting with the same `BleakClient` raises `ValueError("Characteristic notifications already started")`.

CoreBluetooth stops notifications at the OS level on disconnect, but the `PeripheralDelegate` is reused on reconnect (`BleakClientCoreBluetooth.connect()` only creates a new one when `self._delegate is None`), so the stale entries in `_characteristic_notify_callbacks` / `_characteristic_notification_discriminators` survive the disconnect and trip the guard in `start_notifications()`. This matches the root cause analysis in the issue comments.

Changes:
- Add `PeripheralDelegate.clear_notify_callbacks()` and call it from the client's `disconnect_callback`, alongside the existing future cleanup.
- Add `tests/integration/test_issue_1969.py` (connect → `start_notify` → disconnect → reconnect with the same client → `start_notify`), following the existing bumble-based `test_issue_*` pattern. It is CoreBluetooth-only via `skipif`, since BlueZ tracks notification state at the D-Bus level and is unaffected.
- Changelog entry under Unreleased.

`poe lint`, `poe typecheck` and the unit tests pass locally on macOS. 

Verified on real hardware (macOS, Apple Silicon, BlueNRG-based peripheral): on develop the second start_notify() after reconnecting with the same client instance raises ValueError("Characteristic notifications already started"); with this branch the same connect → notify → disconnect → reconnect → notify sequence succeeds.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,7 @@ Added
 Fixed
 -----
 * Fixed handling empty notification payloads in BlueZ backend when using "AcquireNotify". Fixes #1982.
+* Fixed ``ValueError`` when calling ``start_notify()`` after reconnecting in the CoreBluetooth backend. Fixes #1969.
 
 `3.0.2`_ (2026-05-02)
 =====================
--- a/bleak/backends/corebluetooth/PeripheralDelegate.py
+++ b/bleak/backends/corebluetooth/PeripheralDelegate.py
@@ -339,6 +339,18 @@ class PeripheralDelegate:
             self._read_rssi_futures.values(),
         )
 
+    def clear_notify_callbacks(self) -> None:
+        """
+        Removes all registered notification callbacks.
+
+        CoreBluetooth automatically stops notifications when a peripheral
+        disconnects, so this should be called on disconnect to keep the
+        delegate state in sync. Otherwise ``start_notifications`` will fail
+        with "Characteristic notifications already started" after reconnecting.
+        """
+        self._characteristic_notify_callbacks.clear()
+        self._characteristic_notification_discriminators.clear()
+
     async def discover_services(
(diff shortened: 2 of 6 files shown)


dlech:
I'm not a fan of the non-integration tests that are testing the implementation details. They don't check if it actually works with the OS Bluetooth stack and real hardware, so just give a false sense of things working correctly.

PR #1998, opened 2026-06-12
github/workflows: remove pull_request branches selection

JPHutchins:
GitHub imposes no limit on on actions usage by public repositories, so filtering of PRs is removed to aid developers who branch and PR to/from arbitrary branches.


dlech:
Seems reasonable. Thanks.

PR #1999, opened 2026-06-12
bleak: add return_adv parameter to BleakScanner.find_device_by_* methods

### LINKED ISSUE #1277: add return_adv argument to BleakScanner.find_device_by... methods
Similar to `BleakScanner.discover()` we also need a `return_adv=` argument added to the `BleakScanner.find_device_by...()` methods so we can get advertising data without using the deprecated `BLEDevice.metadata`.

https://bleak.readthedocs.io/en/latest/api/scanner.html#bleak.BleakScanner.discover

miguel-neto-andrade:
Fixes #1277.

Adds a `return_adv=` keyword argument to `BleakScanner.find_device_by_address()`, `find_device_by_name()` and `find_device_by_filter()`, mirroring the existing parameter of `BleakScanner.discover()`, so callers can get the matching advertisement data without using the deprecated `BLEDevice.metadata`.

Implementation notes:
- Same `@overload` pattern as `discover()`: `Literal[False]`/`Literal[True]` overloads narrow the return type to `Optional[BLEDevice]` or `Optional[tuple[BLEDevice, AdvertisementData]]`.
- The three public methods delegate to a new private `_find_device_by_filter()` that takes a plain `bool`, so the `by_address`/`by_name` implementations don't have to branch to satisfy overload resolution.
- For `find_device_by_filter()` the returned advertisement is the one that matched the filter; for the other two it is the most recently received one at match time.
- Added an integration test (`test_find_by_name_return_adv`) to `tests/integration/test_scanner.py` and a changelog entry.

`poe lint`, `poe typecheck` (mypy + pyright) and the unit tests pass locally. Also verified on real hardware on macOS (BlueNRG-based peripheral): `return_adv=True` returns the `(BLEDevice, AdvertisementData)` tuple with plausible RSSI, and the default call still returns a bare `BLEDevice`, for both `find_device_by_filter()` and `find_device_by_address()`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,7 @@ and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0
 Added
 -----
 * Added ``BleakAdapter`` class with ``get_connected_devices()`` to retrieve BLE devices that are already connected to the system without scanning.
+* Added ``return_adv`` parameter to ``BleakScanner.find_device_by_address()``, ``BleakScanner.find_device_by_name()`` and ``BleakScanner.find_device_by_filter()``. Fixes #1277.
 
 Fixed
 -----
--- a/bleak/__init__.py
+++ b/bleak/__init__.py
@@ -341,57 +341,148 @@ class BleakScanner:
         """
         return {d[0].address: d for d in self._backend.seen_devices.values()}
 
+    @overload
     @classmethod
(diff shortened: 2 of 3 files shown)


dlech:
I'm not a fan of the way we implemented this before. Having an argument that changes the return type just makes for bad API to work with.

These should be new functions with a different name instead.

PR #2000, opened 2026-06-12
bleak: raise ValueError for oversized characteristic writes

### LINKED ISSUE #737: Better error message for write without response with too large of data
I'm not super familiar with the inner workings of Bleak, but I am getting a `OSError: The parameter is incorrect` after calling `await client.write_gatt_char(MY_UUID, msg)`. Is this related to this issue, or am I doing something wrong?

I have tried it on v0.13, v0.14, and the `develop` branch (v0.15.1 I believe)

Context:
```python
async def run(address, name):
    print("Connecting to {} ({})...".format(name, address))
        print("Connected")
        msg = initMsg()
        await client.write_gatt_char(MY_UUID, msg)
        print("Message sent")

...

loop.run_until_complete(
    run(address, name))
```

_Originally posted by @kaedenbrinkman in https://github.com/hbldh/bleak/issues/705#issuecomment-1012664228_

miguel-neto-andrade:
Fixes #737.

Writing more than `max_write_without_response_size` bytes with `response=False` currently fails with a cryptic OS-specific error (e.g. `OSError: [WinError -2147024809] The parameter is incorrect` on Windows). This adds up-front size checks in the shared `BleakClient.write_gatt_char()` that raise a descriptive `ValueError` instead, before anything is sent to the backend:

- `response=False`: data larger than `characteristic.max_write_without_response_size` raises. Note: BlueZ < 5.62 (Oct 2021) always reports 20, so on those versions the check can raise even when the actual limit is higher.
- `response=True`: data larger than the 512-byte maximum characteristic value length raises.

One place in shared code instead of four backend implementations. Includes unit tests covering all branches, an integration test next to `test_write_gatt_char_no_response`, and a changelog entry under Changed.

`poe lint`, `poe typecheck` (mypy + pyright) and the tests pass locally. Verified on real hardware on macOS with the initial revision of this PR (BlueNRG-based peripheral, negotiated MTU 185): a 183-byte write to a 182-byte-limit characteristic with `response=False` raised the `ValueError` before transmission; the check itself is unchanged since, apart from removing the >20 guard and rewording the message.


bleak/__init__.py
```diff
@@ -922,6 +930,22 @@ async def write_gatt_char(
             # characteristic properties, so doesn't work in some cases.
             response = "write" in characteristic.properties
 
+        if not response:
+            max_size = characteristic.max_write_without_response_size
+            # A value of 20 may just be a fallback for when the actual limit
+            # is not (yet) known, e.g. on BlueZ < 5.62 or before the MTU
```
dlech: BlueZ 5.62 is quite old at this point (Oct 2021), so I would be tempted to leave this workaround out and see if anyone complains.

At a minimum, this workaround should only be enabled for BlueZ. There are real-world peripherals max MTU of 23, so we should expect `max_size == 20` for these. 

And for that matter, we could do a runtime check for older BlueZ by checking if the characteristic has the D-Bus property or not if we want to continue supporting older versions.

bleak/__init__.py
```diff
@@ -922,6 +930,22 @@ async def write_gatt_char(
             # characteristic properties, so doesn't work in some cases.
             response = "write" in characteristic.properties
 
+        if not response:
```
dlech: Why not also checking `response` true case for max size of 512?

bleak/__init__.py
```diff
@@ -922,6 +930,22 @@ async def write_gatt_char(
...
+            # A value of 20 may just be a fallback for when the actual limit
+            # is not (yet) known, e.g. on BlueZ < 5.62 or before the MTU
+            # exchange has completed, so the limit is only enforced when it
+            # is known to be higher to avoid false positives. When the check
+            # is skipped, oversized writes fail with a cryptic OS-specific
+            # error instead, as before.
+            if max_size > 20 and len(data) > max_size:
+                raise ValueError(
+                    f"data is {len(data)} bytes, but characteristic "
+                    f"{characteristic.uuid} only supports {max_size} bytes "
```
dlech: I think the message should mention to use `characteristic.max_write_without_response_size` and not mention the `{max_size}`. Otherwise users may be tempted to hard-code the value, but the value depends on both the host adapter, the OS and the peripheral, so the number won't be the same for everyone even for the same peripheral.

tests/test_write_gatt_char.py
```diff
@@ -0,0 +1,85 @@
...
+from bleak import BleakClient
+from bleak.backends.characteristic import BleakGATTCharacteristic
+from bleak.backends.client import BaseBleakClient
+from bleak.backends.service import BleakGATTService, BleakGATTServiceCollection
+
+TEST_CHARACTERISTIC_UUID = "797f54d7-83d1-49c7-9992-0d0b136117dd"
+
+
+def create_client(backend: BaseBleakClient) -> BleakClient:
+    """Create a BleakClient that uses the given (mock) backend instance."""
```
dlech: Please, no tests with mock backends. The integration tests should be good enough.

bleak/__init__.py
miguel-neto-andrade: Hi! Long-time bleak user here, first time contributing. Thanks for the quick review. Agreed on this one: done in, the limit is now enforced unconditionally with a short comment about BlueZ < 5.62 always reporting 20. Can switch to checking for the D-Bus MTU property instead if you'd rather keep old BlueZ working.

bleak/__init__.py
miguel-neto-andrade: Done. Plus updated the PR title since it's no longer only about write without response.

bleak/__init__.py
miguel-neto-andrade: Done.

tests/integration/test_client_characteristics.py
```diff
@@ -278,6 +278,24 @@ def peripheral_write_callback(connection: Connection, value: bytes):
     assert written_value == b"DATA"
 
 
+@pytest.mark.asyncio(loop_scope="module")
+async def test_write_gatt_char_no_response_too_large(
```
dlech: We should parameterize this (or add a 2nd test) to cover the `response=True` case.

tests/test_write_gatt_char.py
miguel-neto-andrade: Just removed them. I had only added them to increase coverage, since my first PR was failing the codecov patch check for lack of coverage.

tests/integration/test_client_characteristics.py
```diff
@@ -278,6 +278,32 @@ def peripheral_write_callback(connection: Connection, value: bytes):
     assert written_value == b"DATA"
 
 
+@pytest.mark.asyncio(loop_scope="module")
+@pytest.mark.parametrize("response", [True, False])
+async def test_write_gatt_char_too_large(
+    char_test_peripheral: CharTestPeripheral, response: bool
+):
+    """Writing more data than the write operation allows raises ValueError."""
+    if response:
```
dlech: I would either move this stuff into the parameters or just write 2 functions as there is currently less shared code than not.

miguel-neto-andrade: Done.

tests/integration/test_client_characteristics.py
miguel-neto-andrade: Done.

dlech:
Oops, did Claude delete the repo :boom: ?

PR #2010, opened 2026-07-03
feat: expose address type

wiomoc:
closes https://github.com/hbldh/bleak/discussions/1414

Motivation: on linux connecting to a ble peripheral via a [l2cap socket](https://github.com/bluez/bluez/wiki/L2CAP) requires you to specify the address type ([`l2_bdaddr_type`](https://docs.python.org/3/library/socket.html#socket.BDADDR_LE_PUBLIC)). 

Tested on Linux with bluez 5.82 and Windows 11
Untested on Android

### DIFF
--- a/bleak/backends/p4android/scanner.py
+++ b/bleak/backends/p4android/scanner.py
@@ -16,6 +16,7 @@ from jnius import cast, java_method
 
 from bleak._compat import override
 from bleak._compat import timeout as async_timeout
+from bleak.backends.device import BLEAddressType
 from bleak.backends.p4android import defs, utils
 from bleak.backends.scanner import (
     AdvertisementData,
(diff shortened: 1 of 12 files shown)


dlech:
Thanks. A bit of feedback:

* Would be nice to have some integration tests for this.
* I would prefer the enum to have an `UNKNOWN` option rather than allowing `None`. This way we don't have to add extra explanation in the docs about what `None` means in this context.
* It would be nice if we could make this a non-breaking change to the `BLEDevice` constructor as we have some 3rd-party backends now. So either move the new parameter to the end of the list and give it a default value or if we want to keep a more logical order to the args, add an overload and handle it based on the number of args.

wiomoc:
I've followed your feedback up to the integration tests, since I wasn't able to get them to run locally.

bleak/backends/p4android/scanner.py
```diff
@@ -247,6 +248,15 @@ def _handle_scan_result(self, result) -> None:
...
             tx_power = None
 
+        address_type: BLEAddressType
+        match native_device.getAddressType():
+            case defs.ADDRESS_TYPE_PUBLIC:
+                address_type = BLEAddressType.PUBLIC
+            case defs.ADDRESS_TYPE_RANDOM:
+                address_type = BLEAddressType.RANDOM
+            case _:
+                address_type = BLEAddressType.UNKNOWN
```
dlech, line 258: Let's put this in a try/except and fallback to `BLEAddressType.UNKNOWN`

`getAddressType()` appears to be a somewhat new API, so may fail on android < 15

bleak/backends/scanner.py
```diff
@@ -238,6 +238,7 @@ def create_or_update_device(
         self,
         key: str,
         address: str,
+        address_type: BLEAddressType,
```
dlech, line 241: This can also break 3rd-party backends that might be calling this.

Perhaps we should instead modify the `address` parameter to allow two types?

A `str` and a tuple of `str`, `BLEAddressType`. If we get a `str`, then convert it to a tuple with `BLEAddressType.UNKNOWN`.

PR #2012, opened 2026-07-22
backends/bluezdbus: handle D-Bus connection loss

bdraco:
The BlueZ manager never noticed its message bus dying, for example when a D-Bus daemon or proxy restarts. Cached properties kept reporting devices as connected, device watchers never fired, and clients reported connected forever with no disconnect callback, so consumers never reconnected. This is the stall reported in Jc2k/aiohomekit#534.

async_init now arms a watcher on the bus via wait_for_disconnect. When the bus dies the manager clears its cached state, fires Connected condition callbacks and device watchers with False, and reports devices removed so pending service discovery unblocks; the next async_init rebuilds the bus exactly as before. If async_init observes the dead bus before the watcher task runs, it waits for the watcher to finish before replacing the bus, so the notification is delivered exactly once whichever side sees the loss first. BleakClientBlueZDBus.disconnect also tolerates a dead bus now, running the same teardown as a disconnect signal so the disconnected callback fires once and the bus is always released, while still raising the documented timeout if the device does not disconnect.

Devices are reported disconnected even when only a proxy in front of bluetoothd restarted and the BLE link is still up, since without the bus there is no way to tell. This makes clients clean up and reconnect, and connect() already skips the Connect call when the manager sees the device is still connected, so the reconnect is cheap in that case.

Per the guidelines: CHANGELOG.rst is updated, new unit tests cover the state reset, watcher notification, discovery unblock, recovery, reinit race, closed loop cleanup and client disconnect paths, and the branch is a single squashed commit. One known limitation, scanner sessions are not restored automatically after a bus loss, matching existing behavior.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -21,6 +21,8 @@ Changed
 Fixed
 -----
 * Fixed handling empty notification payloads in BlueZ backend when using "AcquireNotify". Fixes #1982.
+* Fixed BlueZ backend not detecting loss of the D-Bus connection, which left clients reporting connected forever and disconnect callbacks never firing.
+* Fixed ``disconnect()`` raising ``EOFError`` and leaking the D-Bus connection in BlueZ backend when the message bus died.
 
 `3.0.2`_ (2026-05-02)
 =====================
--- a/bleak/backends/bluezdbus/client.py
+++ b/bleak/backends/bluezdbus/client.py
@@ -419,12 +419,45 @@ class BleakClientBlueZDBus(BaseBleakClient):
 
                     async with async_timeout(10):
                         await self._disconnecting_event.wait()
+            except (EOFError, OSError) as e:
+                # The D-Bus connection is dead, so the "PropertiesChanged"
+                # signal will never arrive; clean up here so the client does
+                # not report connected forever.
+                logger.debug(
+                    "D-Bus connection lost while disconnecting (%s): %r",
+                    self._device_path,
+                    e,
+                )
+                self._is_connected = False
+                self._cleanup_all()
+                # Wake any concurrent disconnect() call waiting on the event.
+                self._disconnecting_event.set()
             finally:
                 self._disconnecting_event = None
 
-            self._bus.disconnect()
-            await self._bus.wait_for_disconnect()
-            self._bus = None
+            try:
+                self._bus.disconnect()
+                # On a bus that died abnormally, wait_for_disconnect()
+                # re-raises the original socket error.
+                await self._bus.wait_for_disconnect()
+            except (EOFError, OSError) as e:
+                # Expected when the bus is already dead
+                logger.debug(
+                    "error disconnecting from bus (%s): %r", self._device_path, e
+                )
+            except Exception as e:
+                # Deliberately not re-raised: wait_for_disconnect() replays
+                # whatever error killed the reader when the bus died, which
+                # can be any type and was already handled at that time, and
+                # a raise here would leave callers unable to release a dead
+                # connection.
+                logger.warning(
+                    "unexpected error disconnecting from bus (%s): %r",
+                    self._device_path,
+                    e,
+                )
+            finally:
+                self._bus = None
 
         # sanity check to make sure _cleanup_all() was triggered by the
         # "PropertiesChanged" signal handler and that it completed successfully
--- a/bleak/backends/bluezdbus/manager.py
+++ b/bleak/backends/bluezdbus/manager.py
@@ -18,7 +18,7 @@ import contextlib
 import logging
 import os
 from collections import defaultdict
-from collections.abc import Callable, Coroutine
+from collections.abc import Callable, Coroutine, Iterator
 from functools import partial
 from typing import Any, NamedTuple, Optional, cast
 
@@ -159,6 +159,17 @@ _ADVERTISING_DATA_PROPERTIES = {
 }
 
 
+def _fire_callback(callback: Callable[..., None], *args: Any) -> None:
+    """
+    Calls a callback, isolating exceptions so they do not stop the
+    remaining callbacks or propagate into the D-Bus message reader.
+    """
+    try:
+        callback(*args)
+    except Exception:
+        logger.exception("error in callback")
+
+
 def get_max_write_without_response_size(char_props: GattCharacteristic1) -> int:
     # "MTU" property was added in BlueZ 5.62, otherwise fall
     # back to minimum MTU according to Bluetooth spec.
@@ -175,6 +186,7 @@ class BlueZManager:
     def __init__(self) -> None:
         self._bus: Optional[MessageBus] = None
         self._bus_lock = asyncio.Lock()
+        self._bus_watcher_task: Optional[asyncio.Task[None]] = None
 
         # dict of object path: dict of interface name: dict of property name: property value
         self._properties: dict[str, dict[str, dict[str, Any]]] = {}
@@ -199,6 +211,25 @@ class BlueZManager:
         self._device_watchers: dict[str, set[DeviceWatcher]] = {}
         self._condition_callbacks: dict[str, set[DeviceConditionCallback]] = {}
         self._services_cache: dict[str, BleakGATTServiceCollection] = {}
+
+    def _reset_cached_state(self) -> None:
+        """Clears all state cached from the message bus."""
+        self._properties.clear()
+        self._service_map.clear()
+        self._characteristic_map.clear()
+        self._descriptor_map.clear()
+        self._services_cache.clear()
+        self._adapters.clear()
+
+    def _iter_device_props(self) -> Iterator[tuple[str, Device1]]:
+        """
+        Yields the D-Bus object path and device properties for each
+        device in BlueZ.
+        """
+        for path, interfaces in self._properties.items():
+            props = cast(Device1 | None, interfaces.get(defs.DEVICE_INTERFACE))
+            if props is not None:
+                yield path, props
 
     def _check_adapter(self, adapter_path: str) -> None:
         """
@@ -309,12 +340,9 @@ class BlueZManager:
                 )
                 assert_reply(reply)
 
-                # dictionaries are cleared in case AddInterfaces was received first
+                # state is cleared in case AddInterfaces was received first
                 # or there was a bus reset and we are reconnecting
-                self._properties.clear()
-                self._service_map.clear()
-                self._characteristic_map.clear()
-                self._descriptor_map.clear()
+                self._reset_cached_state()
 
                 for path, interfaces in reply.body[0].items():
                     props = unpack_variants(interfaces)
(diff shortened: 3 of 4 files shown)


bdraco:
@bluetoothbot review

bdraco:
@bluetoothbot review

bdraco:
@bluetoothbot review

bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]
> No blocking issues found — ready to merge.

bleak/backends/bluezdbus/manager.py
```diff
@@ -367,6 +383,86 @@ async def async_init(self) -> None:
...
+
+    async def _monitor_bus_disconnect(self, bus: MessageBus) -> None:
+        """
+        Waits for the message bus to disconnect and resets the manager state.
+        """
+        exc: Optional[Exception] = None
+        try:
+            await bus.wait_for_disconnect()
+        except Exception as e:
+            exc = e
```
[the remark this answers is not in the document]
bdraco: This is wrong, CancelledError is BaseException and we don't support older python where it wasn't

bleak/backends/bluezdbus/client.py
```diff
@@ -419,11 +419,31 @@ async def disconnect(self) -> None:
...
-            await self._bus.wait_for_disconnect()
+            try:
+                self._bus.disconnect()
+                # On a bus that died abnormally, wait_for_disconnect()
+                # re-raises the original socket error.
+                await self._bus.wait_for_disconnect()
+            except Exception as e:
+                logger.debug(
+                    "error disconnecting from bus (%s): %r", self._device_path, e
+                )
```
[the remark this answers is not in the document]
bdraco: This is wrong, CancelledError is BaseException and we don't support older python where it wasn't

bleak/backends/bluezdbus/manager.py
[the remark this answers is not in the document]
bluetoothbot: Correct. `CancelledError` extends `BaseException`. `except Exception` never catches it. Cancellation propagates cleanly.

bleak/backends/bluezdbus/client.py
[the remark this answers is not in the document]
bluetoothbot: Correct. `CancelledError` extends `BaseException`. Broad `except Exception` skips it. Cooperative cancellation preserved.

bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]
> No blocking issues found — ready to merge.

bdraco:
The services cache clear at the top of async_init was already restored in the newer head, and the teardown catch now logs expected socket errors at debug and anything unexpected at warning while still always releasing the bus.

@bluetoothbot review

bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]
> No blocking issues found — ready to merge.

bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]
> No blocking issues found — ready to merge.

bdraco:
Just needs another round of functional testing after the last round of cleanups before can be marked ready

bdraco:
bot's find in https://github.com/hbldh/bleak/pull/2013#pullrequestreview-4759467676 really belongs here so one more push coming

bdraco:
Good catch on the closed loop cancel; task.cancel is now wrapped in suppress since the wakeup would be scheduled on the closed loop, and there is a new regression test that builds a real manager on a genuinely closed loop through an executor thread so the real cancel path is exercised instead of a mock.

@bluetoothbot review

bdraco:
The teardown catch all is deliberate; wait_for_disconnect replays whatever error killed the reader when the bus died, which can be any exception type and was already handled at that time, and raising from disconnect would leave callers unable to release a dead connection, which is the exact stall this fixes. The release of the bus is now in a finally and the comment documents the intent.

@bluetoothbot review

bluetoothbot (approved):
> [!TIP]
> No blocking issues found — ready to merge.

dlech:
Fable's review:


Verdict: the approach is right and I'd merge it after two fixes. One is a real gap in the fix itself, the other is a misleading piece of code the tests enshrine. Everything else checks out: the full suite, ruff/black/isort/flake8, mypy and pyright all pass locally on the branch, and the new tests are well targeted.

What it does. async_init starts a task that awaits wait_for_disconnect() on the manager's bus. When the bus dies, the task takes the bus lock, clears cached state, fires Connected=False to condition callbacks and device watchers, and reports every device removed so pending service discovery and scanners unblock. The client's disconnect() now tolerates a dead bus instead of raising EOFError. I traced the client side: the watcher callback sets _is_connected=False, cleans up, and fires the user's disconnected callback, which is exactly what aiohomekit needs.

Finding 1, should fix: reinit can race the watcher and silently drop the notifications. dbus_fast's _finalize fails pending method calls first, then resolves the disconnect future, so a caller whose bus.call just failed wakes before the watcher task. If that caller retries connect(), it reaches async_init, which sees connected == False, swaps in a new bus, and cancels the old watcher while it is still waiting on the lock. Nobody is ever told the old bus died. I reproduced this against the PR's own fake bus with a test that calls bus.die() followed immediately by manager.async_init(). The device watcher list stays empty. This is the original stall reappearing in a narrow window, and it matters most when bluetoothd also restarted, since then the new GetManagedObjects won't even list the device. Suggested fix: in async_init, when the existing bus is not connected, call _handle_bus_disconnect() before creating the replacement bus. The watcher's bus is not self._bus guard then makes the late task a no-op. That also means test_reinit_does_not_notify_watchers asserts the wrong thing. Its comment says "the socket has not errored yet", but in dbus_fast connected only turns False in _finalize, which resolves the future in the same call, so that state is not reachable for the manager's bus.

Finding 2, should fix: the closed-loop task.cancel() does not do what its comment says. The comment claims cancelling avoids the "Task was destroyed but it is pending" message. I tested this in isolation: the message prints with or without the cancel, and cancel() raises RuntimeError after already cancelling the inner future. Running the PR's own tests also prints that message from the closed-loop tests. Either drop the cancel block and the assert_called_once_with() in test_global_manager_cleans_up_closed_loops, or reword the comment to say it's best effort and does not suppress the message. Note this is only a concern for users who call run_until_complete and close() by hand. asyncio.run cancels all tasks on shutdown, so the common path is clean.

Smaller notes, non-blocking:

The dead-bus branch in disconnect() logs at DEBUG. The bot's suggestion to raise it to INFO or WARNING is reasonable, since the manager logs the same event at WARNING.
_fire_callback changes behaviour for PropertiesChanged too. An exception in a user's disconnected callback used to propagate into the dbus_fast reader. Now it's logged. That's an improvement, but it's a user-visible change and could use a line in the changelog.
Semantics when only a proxy restarts. The PR reports devices disconnected even though the BLE link may still be up in bluetoothd. The inline comment acknowledges this, and connect() already skips Connect when the manager sees the device connected, so reconnects work. Worth a sentence in the PR body or docs.
Copilot's duplicate-clearing comment is stale. _handle_bus_disconnect already uses _reset_cached_state(). bdraco's replies to the CancelledError comments are correct.
The .readthedocs.yml hunk in the PR diff is not the PR's. The branch is based on 496d16e and develop has moved one commit. A rebase will make it disappear.
One thing for the #2013 follow-up: the real-daemon test would be a natural place to also cover finding 1, since a SIGKILLed daemon plus an immediate async_init is exactly the sequence that races.

dlech:
I'm not especially a fan of `_fire_callback` swallowing exceptions. If there is an unhandled exception in a callback, I think it should fail loudly rather than just being logged. If there is an unhandled exception, I consider it a bug that needs to be fixed rather than getting away with ignoring it.

bluetoothbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - async_init() races the watcher task and silently drops the bus-loss notification
> - `except (EOFError, OSError)` swallows the documented 10-second disconnect timeout on Python 3.11+

bdraco:
I forgot about this PR. I'll do some cleanup

bdraco:
> I'm not especially a fan of `_fire_callback` swallowing exceptions. If there is an unhandled exception in a callback, I think it should fail loudly rather than just being logged. If there is an unhandled exception, I consider it a bug that needs to be fixed rather than getting away with ignoring it.

Its been a minute, but will check this. If its third party code, catching Exception seems right. If its first party, agree that its a bug and can just blow up

bdraco:
ok first party here. dropping the Exception catch

bluetoothbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Dead-bus disconnect path leaks the _disconnect_monitor task and never fires disconnected_callback

bdraco:
clean it up a bit. still need to think about disconnect loop race

bluetoothbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - disconnected_callback fires twice when both buses die during disconnect()

bdraco:
> 3. One raising callback aborts notification for every remaining device
> [bleak/backends/bluezdbus/manager.py:436-443](https://github.com/hbldh/bleak/blob/376dd402d666bc8d84b9dbd564826460aac61987/bleak/backends/bluezdbus/manager.py#L436-L443)

accepted per maintainer request above

bdraco:
> 4. Teardown catch-all is wider than its own comment, hiding real bugs at debug level

intended disconnect() can raise OSError on a dying socket and the catch is what guarantees the handle is released

bluetoothbot (approved):
> [!TIP]
> No blocking issues found — ready to merge.

bluetoothbot (approved):
> [!TIP]
> No blocking issues found — ready to merge.

bluetoothbot (approved):
> [!TIP]
> No blocking issues found — ready to merge.

bdraco:
> 1. [Pre-Existing Issue] `startswith(adapter_path)` matches sibling adapters with a numeric-prefix name
> [bleak/backends/bluezdbus/manager.py:477-479](https://github.com/hbldh/bleak/blob/726562e9468ecf73080eb9f18a990607f521e911/bleak/backends/bluezdbus/manager.py#L477-L479)
> `"/org/bluez/hci10/dev_X".startswith("/org/bluez/hci1")` is `True`

real bug I probably actually hit when testing with 12 hcis ...but probably has audience of 1, still guard is cheap

bluetoothbot (approved):
> [!TIP]
> No blocking issues found — ready to merge.

bleak/backends/bluezdbus/manager.py
```diff
@@ -359,14 +385,100 @@ async def async_init(self) -> None:
...
+                # callbacks may remove the watcher, hence the copy
+                for watcher in watchers.copy():
+                    watcher.on_connected_changed(connected)
+
+    def _notify_device_removed(self, device_path: str) -> None:
+        """
+        Runs the device removed callbacks for the adapter of the device.
+        """
+        for callback, adapter_path in self._device_removed_callbacks:
+            # hci1 must not match hci10
```
dlech: Is this just because of the test code?

Maybe we should make that hci99 and not have a mysterious comment there.

CHANGELOG.rst
```diff
@@ -23,6 +23,9 @@ Fixed
 * Fixed deprecated ``adapter`` keyword argument mutating shared defaults and caller-provided ``bluez`` arguments in ``BleakScanner`` and ``BleakClient``. Fixes #2028.
 * Fixed handling empty notification payloads in BlueZ backend when using "AcquireNotify". Fixes #1982.
 * Fixed crash in WinRT backend if a WinRT event handler is called after the asyncio event loop is closed.
+* Fixed BlueZ backend not detecting loss of the D-Bus connection, which left clients reporting connected forever and disconnect callbacks never firing.
+* Fixed ``disconnect()`` raising ``EOFError`` and leaking the D-Bus connection in BlueZ backend when the message bus died.
+* Fixed device removed callbacks in BlueZ backend matching devices on adapters whose name starts with the watched adapter name, e.g. ``hci10`` for ``hci1``.
```
dlech: Was this a pre-existing issue? Can we split that out into a separate PR?

bdraco: Yes, pre-existing

bdraco: It will conflict here though

dlech: Can we redo the commits here to a small number of logical commits then?

bdraco: https://github.com/hbldh/bleak/pull/2031

bleak/backends/bluezdbus/manager.py
dlech: disregard, my brain read `hci1` as `this` for some reason.

CHANGELOG.rst
bdraco: I'll resolve conflicts and squash this once https://github.com/hbldh/bleak/pull/2031 goes in

dlech: done

bdraco: double checking the conflict res now

bdraco: all good

dlech:
This was a tricky one, thanks!

bdraco:
Thanks. It was for sure, and the bot found seemingly endless edge cases

PR #2013, opened 2026-07-22
tests/bluezdbus: exercise bus loss recovery against a real dbus-daemon

bdraco:
The unit tests added in #2012 simulate the message bus with a fake, so the real failure physics were untested. This adds integration tests that spawn a private dbus-daemon from a minimal config, serve a fake org.bluez adapter and connected device over it with dbus_fast exported interfaces, then kill the daemon with SIGKILL. That drives a genuine EOF through the dbus_fast reader, real authentication, real GetManagedObjects serialization and real PropertiesChanged delivery, and asserts the watcher fires, state resets, async_init recovers against a replacement daemon at a new address, and a watcher registered before the loss still receives signals on the new bus. Further cases cover async_init observing the dead bus before the watcher task runs, and finalizing the stale bus of a closed event loop.

The shared test helpers move into a conftest with a registry cleanup fixture and a closed loop helper used by both test modules. Skips cleanly when dbus-daemon is unavailable; no production code changes, so no changelog entry.

PR #2018, opened 2026-08-29
Fix EOFError from BleakClient.disconnect() on BlueZ backend

monperrus:
## Bug

Bleak's BlueZ client opens a dedicated D-Bus connection per BLE session (self._bus) and sends its own "Disconnect" method call on it from disconnect(), then awaits the reply. Separately, the shared global manager holds its own D-Bus connection that watches BlueZ's "Disconnected" PropertiesChanged signal and, on it, synchronously runs on_connected_changed() -> _cleanup_all(), which closes self._bus.

These two connections race: when BlueZ's "Disconnected" signal is delivered on the manager's connection before the reply to the client's own "Disconnect" call arrives on self._bus, _cleanup_all() closes self._bus out from under the pending call.

Reproduced consistently against a real BlueZ peripheral on a dbus-fast/BlueZ combination.

## Fix

Fix disconnect() to catch EOFError from the pending "Disconnect" call and treat it as a successful disconnect when self._bus is already None (i.e. _cleanup_all() has already run) . Also guard the follow-up self._bus.disconnect()/wait_for_disconnect() so it isn't called a second time when the race has already closed and cleared self._bus.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -21,6 +21,7 @@ Changed
 Fixed
 -----
 * Fixed handling empty notification payloads in BlueZ backend when using "AcquireNotify". Fixes #1982.
+* Fixed ``BleakClient.disconnect()`` on the BlueZ backend raising ``EOFError`` when BlueZ's "Disconnected" signal (delivered on the shared manager D-Bus connection) races the reply to the client's own "Disconnect" call and closes the client's D-Bus connection first.
 
 `3.0.2`_ (2026-05-02)
 =====================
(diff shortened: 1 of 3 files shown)


dlech:
Any chance we can get a test to exercise this code?

monperrus:
@dlech done, thanks!

dlech:
The description sounds like maybe this was a bug in Bleak 2.x but should not be a problem in Bleak 3.x since https://github.com/hbldh/bleak/pull/1923. Can you confirm?

dlech:
I think this might also be similar to #2013 and/or #2012.

monperrus:
I moved on to another project, no way to reproduce the bug anymore sorry. Thanks.

dlech:
No worries, I'm pretty sure #2012 fixed this already, so we can close this one. Thanks for the reply.

PR #2020, opened 2026-09-02
Fix CoreBluetooth scan memory growth

hxseven:
## Context

Hello! I came across this while investigating what appeared to be a memory leak in a [Theengs Gateway](https://github.com/theengs/gateway) process that uses Bleak for continuous BLE-to-MQTT scanning. After a week on macOS, its memory usage had grown from roughly 50 MiB to over 2 GiB while the Python heap remained small. I traced the growing number of native objects through Theengs Gateway → Bleak → PyObjC → Foundation/CoreBluetooth → the Swift runtime and found two separate ways objects built up while processing each BLE advertisement.

The issue becomes apparent with real CoreBluetooth traffic forwarded onto a long-running asyncio thread and enough processed advertisements for the native objects to accumulate.

## Summary

Long-running CoreBluetooth scans accumulated native objects through two paths:

- discovery processing ran on an asyncio worker thread without an autorelease pool
- service and manufacturer data used PyObjC's leaking `NSData.bytes()` mapping

This change processes each discovery inside `objc.autorelease_pool()` and copies `NSData` through its managed buffer protocol. The PyObjC bug is addressed at its source in [pyobjc#689](https://github.com/ronaldoussoren/pyobjc/pull/689); the local conversion keeps Bleak compatible with its currently supported PyObjC versions.

This is separate from the delegate-cycle leak fixed in #1925 and released in 3.0.0. That bug kept delegate objects alive after a scanner stopped; this one accumulates objects while the scanner is still running and reproduces on current `develop`.

## Evidence

On macOS 14.8.4, I ran repeated 7-second scans separated by 5-second pauses and measured native allocations with `/usr/bin/heap`. After 635 advertisements, unpatched Bleak held 1,992 `Swift.StringStorage`, 277 `__NSFastEnumerationEnumerator`, and 668 `NSData` objects.

An autorelease-pool-only run stopped the growth in Swift strings and enumerators. However, the number of live `NSData` objects continued to grow by roughly one per processed advertisement. This confirmed that the buffer conversion was a second source of the leak.

With this PR and stock PyObjC 12.1, the native object counts remained bounded. After more than 1,000 advertisements, a snapshot held 10 Swift strings, no fast-enumeration objects, and 5 NSData objects with a 22 MiB footprint. Counts varied with currently visible devices instead of growing with the total number of advertisements.

The same patched Bleak build then ran in the original Theengs Gateway process for more than four hours while continuing to publish BLE data. Its memory usage was still 34 MiB with a 38 MiB peak. A native heap snapshot taken near the four-hour mark also showed that the relevant object counts remained low.

## Checks

- lint, mypy, and pyright
- Python 3.10–3.14 test matrix (`18 passed, 104 skipped` per version)
- refcount regression test for `NSData` conversion
- repeated scans and the live Theengs workload with native heap snapshots

I kept the two Bleak fixes together because either one alone leaves the reproduced scanner growth unresolved, but I am happy to split them if that is easier to review.

Disclosure: I used OpenAI Codex extensively to investigate and implement this fix. I carefully reviewed and tested the changes, reproduced the behavior described above, and take responsibility for the contribution.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -21,8 +21,7 @@ Changed
 Fixed
 -----
 * Fixed handling empty notification payloads in BlueZ backend when using "AcquireNotify". Fixes #1982.
-* Fixed native object accumulation while scanning with CoreBluetooth on a
-  long-running asyncio worker thread.
+* Fixed memory leak when scanning in CoreBluetooth backend.
 
 `3.0.2`_ (2026-05-02)
 =====================
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -21,6 +21,8 @@ Changed
 Fixed
 -----
 * Fixed handling empty notification payloads in BlueZ backend when using "AcquireNotify". Fixes #1982.
+* Fixed native object accumulation while scanning with CoreBluetooth on a
+  long-running asyncio worker thread.
 
 `3.0.2`_ (2026-05-02)
 =====================
--- a/bleak/backends/corebluetooth/CentralManagerDelegate.py
+++ b/bleak/backends/corebluetooth/CentralManagerDelegate.py
@@ -347,7 +347,7 @@ class CentralManagerDelegate:
         # This behaviour could be affected by the
         # CBCentralManagerScanOptionAllowDuplicatesKey global setting.
 
-        with objc.autorelease_pool():  # type: ignore[attr-defined]
+        with objc.autorelease_pool():
             uuid_string = peripheral.identifier().UUIDString()
 
             for callback in self.callbacks.values():
--- a/bleak/backends/corebluetooth/CentralManagerDelegate.py
+++ b/bleak/backends/corebluetooth/CentralManagerDelegate.py
@@ -347,6 +347,9 @@ class CentralManagerDelegate:
         # This behaviour could be affected by the
         # CBCentralManagerScanOptionAllowDuplicatesKey global setting.
 
+        # This method may run on a long-running asyncio thread where temporary
+        # Objective-C objects would otherwise accumulate. Use an autorelease pool per
+        # advertisement to release them promptly.
         with objc.autorelease_pool():
             uuid_string = peripheral.identifier().UUIDString()
 
--- a/bleak/backends/corebluetooth/CentralManagerDelegate.py
+++ b/bleak/backends/corebluetooth/CentralManagerDelegate.py
@@ -347,19 +347,20 @@ class CentralManagerDelegate:
         # This behaviour could be affected by the
         # CBCentralManagerScanOptionAllowDuplicatesKey global setting.
 
-        uuid_string = peripheral.identifier().UUIDString()
+        with objc.autorelease_pool():  # type: ignore[attr-defined]
+            uuid_string = peripheral.identifier().UUIDString()
 
-        for callback in self.callbacks.values():
-            callback(peripheral, cast(CBAdvertisementData, advertisementData), RSSI)
+            for callback in self.callbacks.values():
+                callback(peripheral, cast(CBAdvertisementData, advertisementData), RSSI)
 
-        logger.debug(
-            "Discovered device %s: %s @ RSSI: %d (kCBAdvData %r) and Central: %r",
-            uuid_string,
-            peripheral.name(),
-            RSSI,
-            advertisementData.keys(),
-            central,
-        )
+            logger.debug(
+                "Discovered device %s: %s @ RSSI: %d (kCBAdvData %r) and Central: %r",
+                uuid_string,
+                peripheral.name(),
+                RSSI,
+                advertisementData.keys(),
+                central,
+            )
 
     def did_connect_peripheral(
         self, central: CBCentralManager, peripheral: CBPeripheral
--- a/tests/backends/corebluetooth/test_scanner.py
+++ b/tests/backends/corebluetooth/test_scanner.py
@@ -1,4 +1,9 @@
 import sys
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    if sys.platform != "darwin":
+        assert False, "This backend is only available on macOS"
 
 import pytest
 
@@ -15,7 +20,7 @@ from bleak.backends.corebluetooth.scanner import (
 
 
 def test_nsdata_to_bytes_does_not_retain_data():
-    data = NSData(b"test")  # pyright: ignore[reportAbstractUsage, reportCallIssue]
+    data = NSData.alloc().initWithBytes_length_(b"test", 4)
     initial_refcount = sys.getrefcount(data)
 
     for _ in range(10):
--- a/tests/backends/corebluetooth/test_scanner.py
+++ b/tests/backends/corebluetooth/test_scanner.py
@@ -1,0 +1,24 @@
+import sys
+
+import pytest
+
+# isort: off
+
+if not sys.platform.startswith("darwin"):
+    pytest.skip("backend only available on macOS", allow_module_level=True)
+
+from Foundation import NSData
+
+from bleak.backends.corebluetooth.scanner import (
+    _nsdata_to_bytes,  # pyright: ignore[reportPrivateUsage]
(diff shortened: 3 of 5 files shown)


bleak/backends/corebluetooth/CentralManagerDelegate.py
```diff
@@ -347,19 +347,20 @@ def did_discover_peripheral(
...
-
-        logger.debug(
-            "Discovered device %s: %s @ RSSI: %d (kCBAdvData %r) and Central: %r",
-            uuid_string,
-            peripheral.name(),
-            RSSI,
-            advertisementData.keys(),
-            central,
-        )
+        with objc.autorelease_pool():  # type: ignore[attr-defined]
```
dlech: we should add this to our local typings

dlech:
> The PyObjC bug is addressed at its source in [pyobjc#689](https://github.com/ronaldoussoren/pyobjc/pull/689); the local conversion keeps Bleak compatible with its currently supported PyObjC versions.

let's add some comments in the code explaining this so that we can drop the workaround when we update the minimum version of PyObj, assuming that the fix is accepted upstream.

dlech:
Need to fix the test so it doesn't make other oses fail.

tests/backends/corebluetooth/test_scanner.py
```diff
@@ -0,0 +1,24 @@
...
+
+from Foundation import NSData
+
+from bleak.backends.corebluetooth.scanner import (
+    _nsdata_to_bytes,  # pyright: ignore[reportPrivateUsage]
+)
+
+
+def test_nsdata_to_bytes_does_not_retain_data():
+    data = NSData(b"test")  # pyright: ignore[reportAbstractUsage, reportCallIssue]
```
dlech: Should be able to fix it so we don't need `# pyright: ignore[reportAbstractUsage, reportCallIssue]`.

Apparently we are using `NSData.alloc().initWithBytes_length_(data, len(data))` instead. So I would do that or fix the typings.

CHANGELOG.rst
```diff
@@ -21,6 +21,8 @@ Changed
 Fixed
 -----
 * Fixed handling empty notification payloads in BlueZ backend when using "AcquireNotify". Fixes #1982.
+* Fixed native object accumulation while scanning with CoreBluetooth on a
+  long-running asyncio worker thread.
```
dlech: Let's try to make this shorter and more understandable by a human.

e.g. "Fixed memory leak when scanning in CoreBluetooth backend."

bleak/backends/corebluetooth/CentralManagerDelegate.py
hxseven: Added autorelease_pool to the local objc typings and removed the ignore in 596180b. Thanks!

tests/backends/corebluetooth/test_scanner.py
hxseven: Updated the test to use alloc().initWithBytes_length_() and removed the ignores in 596180b.

CHANGELOG.rst
hxseven: Changed to the suggested changelog wording, thanks.

dlech:
Can we get a comment explaining what the autorelease pool is doing too?

hxseven:
Thanks for the review! I addressed the cross-platform Pyright issue and added code comments explaining both the PyObjC workaround and the autorelease pool.

The autorelease pool is independent of [pyobjc#689](https://github.com/ronaldoussoren/pyobjc/pull/689) and is intended to remain. It bounds the lifetime of temporary Objective-C objects created while processing each advertisement instead of allowing them to accumulate on a long-running asyncio thread. Nested pools are supported, so this is also safe when the thread already has one. The trade-off is one pool creation and drain per advertisement.

Only `_nsdata_to_bytes()` is a temporary compatibility workaround. It can be removed once Bleak’s minimum PyObjC version includes the upstream fix.

I also verified the full type checks and the Python 3.13 test suite in a clean Linux container. In the previous CI run, the two failed BlueZ jobs timed out while downloading static `apk-tools` during Alpine VM provisioning, before any tests ran. The same suite passed on the other three Python versions, and all regular macOS, Windows, and Ubuntu build/test jobs passed.

Codecov still flags the platform-specific autorelease-pool block and the two scanner conversion call sites as uncovered, while project coverage passes. Would you prefer additional unit coverage for those paths, or is the refcount regression test together with the native-heap validation sufficient?

For future reference, I’ve included the design rationale and trade-offs below.

<details>
<summary>Autorelease pool: design rationale and approaches</summary>

### How the autorelease pool applies here

Some Objective-C APIs return temporary objects that are scheduled for later release through the current thread’s autorelease pool. Draining the pool releases objects that have not been retained elsewhere. PyObjC automatically creates a pool on the thread where it is first imported, but other threads require explicit pool management. Since this callback is dispatched to Bleak’s asyncio event loop, it can run repeatedly on a long-running thread without a regularly drained pool.

Entering the context creates a nested pool; leaving it drains that pool. This gives native temporaries created while processing each advertisement a short, predictable lifetime.

### Behavior in different situations

| Situation | Behavior |
| --- | --- |
| No regularly drained outer pool | The local pool prevents temporary Objective-C objects from accumulating during a long-running scan. |
| An outer pool already exists | The local pool is safely nested and releases advertisement-specific temporary objects sooner. |
| Short scan or short-lived process | Functional behavior remains the same, apart from the additional cost of creating and draining a pool for each advertisement. |
| Long-running scan | Autoreleased native objects receive a predictable, bounded lifetime instead of accumulating between outer-pool drains. |
| An exception leaves the callback | The context manager still drains the pool while propagating the exception. |
| An object is retained elsewhere | It remains valid; draining affects only objects that have not been retained beyond the pool. |

### Approaches compared

| Approach | Advantages | Trade-offs | Decision |
| --- | --- | --- | --- |
| Do not create a local pool | No per-advertisement pool-management overhead. | Relies on the thread already having a regularly drained pool. Background asyncio threads may not, allowing temporary objects to accumulate. | Behavior before this change |
| Use one pool for the whole scan | Creates and drains only one pool. | A scan can run indefinitely, so temporary objects could accumulate until it stops. | Not selected |
| Drain a pool periodically | Amortizes the pool-management overhead across multiple advertisements. | Requires additional batching or lifecycle logic, an arbitrary drain interval, and still permits temporary growth between drains. | Not selected |
| Use one pool per advertisement | Gives temporary objects the smallest natural lifetime and works with or without an outer pool. | Adds one pool creation and drain per advertisement. | **Selected** |

### Trade-off and lifetime

I did not benchmark the per-advertisement pool overhead. However, this is the conventional pattern for repeated Objective-C work on a long-running thread, and it avoids adding batching state or relying on the caller’s thread setup. If profiling later shows material overhead at very high advertisement rates, periodic draining could be reconsidered with measurements.

The autorelease pool should remain even after [pyobjc#689](https://github.com/ronaldoussoren/pyobjc/pull/689) is available. That fix only addresses the exporter reference retained by `NSData.bytes()`. It makes `_nsdata_to_bytes()` removable once Bleak requires a fixed PyObjC release, but it does not manage the lifetime of other temporary Objective-C objects created while processing advertisements. Removing the local pool would restore the previous reliance on the thread’s pool setup and could reintroduce native-memory growth in long-running background scanners.

</details>

dlech:
I think we have a similar leak here.

https://github.com/hbldh/bleak/blob/9a277f67e88140c6fe2007b5d2392784f1ef8266/bleak/backends/corebluetooth/PeripheralDelegate.py#L592

hxseven:
Good catch! I confirmed that this is the same `NSData.bytes()` exporter-reference leak. A local refcount check showed that `bytes(value)` retains one reference per call, while the adjacent `bytearray(value)` conversion does not. I moved the workaround into the shared CoreBluetooth utilities and used it for this conversion as well.

dlech:
Thanks!

hxseven:
Thanks for reviewing and merging this. I really appreciate the effort you put into Bleak & maintaining it! 😊

Issue #1689, opened 2024-11-18
[org.bluez.Error.InProgress] Operation already in progress


9acca9:
Im using the development Bleak version in Ubuntu 22.04

The thing is that i get this erro when try to run my code:
```
python mainBleak.py 
Connecting to device...
[org.bluez.Error.InProgress] Operation already in progress

```
It seems that Ubuntu is always scanning and well... Somebody here have experience with this issue? i try several things and nothing... 
ex.:
```
 bluetoothctl
Agent registered
[CHG] Controller 00:15:83:0C:BF:EB Pairable: yes
[bluetooth]# scan off
Failed to stop discovery: org.bluez.Error.Failed
[bluetooth]# 
```

Thanks

dlech:
I had this happen to me today. I have a USB Bluetooth dongle, so I unplugged it and plugged it back in and I stopped getting the "Operation already in progress" error when scanning. (Ubuntu 24.04, BlueZ 5.72).

So seems like a BlueZ issue.

hemangi825:
i get issue while write in BLE device, how to solve in nodejs i am using node-ble libs for that 
Please anyone suggest how to solve this issue

Issue #958, opened 2022-08-23
"BleakDBusError[org.bluez.Error.InProgress] Operation already in progress" when stopping scanner


des-gan09:
* bleak version: 0.14.3
* Python version: 3.9.2
* Operating System: Raspberry Pi 3B+ (Raspbian Bullseye) 
* BlueZ version (`bluetoothctl -v`) in case of Linux: 5.55

### Description

I am trying to send a command and read from a sensor. I used a modified uart_example which works for the Pi 4s but for the Pi 3s, I sometimes encounter this issue. 
```
Traceback (most recent call last):
  File "/home/pi/powersensor-firmware/sampling/debug.py", line 115, in <module>
    asyncio.run(uart_terminal())
  File "/usr/lib/python3.9/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/usr/lib/python3.9/asyncio/base_events.py", line 642, in run_until_complete
    return future.result()
  File "/home/pi/powersensor-firmware/sampling/debug.py", line 49, in uart_terminal
    device = await BleakScanner.find_device_by_address(MAC, timeout=60)
  File "/home/pi/.local/lib/python3.9/site-packages/bleak/backends/scanner.py", line 220, in find_device_by_address
    return await cls.find_device_by_filter(
  File "/home/pi/.local/lib/python3.9/site-packages/bleak/backends/scanner.py", line 253, in find_device_by_filter
    return None
  File "/home/pi/.local/lib/python3.9/site-packages/bleak/backends/scanner.py", line 99, in __aexit__
    await self.stop()
  File "/home/pi/.local/lib/python3.9/site-packages/bleak/backends/bluezdbus/scanner.py", line 177, in stop
    assert_reply(reply)
  File "/home/pi/.local/lib/python3.9/site-packages/bleak/backends/bluezdbus/utils.py", line 23, in assert_reply
    raise BleakDBusError(reply.error_name, reply.body)
bleak.exc.BleakDBusError: [org.bluez.Error.InProgress] Operation already in progress
```

Here's my code

```python
import asyncio
from gettext import find
from http import client
import sys
import logging
import os

from bleak import BleakScanner, BleakClient
from bleak.backends.scanner import AdvertisementData
from bleak.backends.device import BLEDevice
from datetime import datetime
import time
import queue

MAC = sys.argv[1].upper().rstrip('\n')
now = datetime.now()
dt_string = now.strftime("%Y-%m-%dT%H:%M:%S")
[...]

PR #2022, opened 2026-09-03
backends/bluezdbus: tolerate InProgress from StopDiscovery

### LINKED ISSUE #2021: BlueZ: BleakScanner.stop() raises org.bluez.Error.InProgress after the session is already gone; treat it like NotReady
### Environment

* bleak 3.0.2 (BlueZ backend)
* BlueZ 5.72 (bluetoothd), Linux 6.12.90-venus-4, Victron Venus OS v3.79 on a Cerbo GX (armhf)
* Python 3.12

### Summary

`BleakScanner.stop()` raises `BleakDBusError: [org.bluez.Error.InProgress] Operation already in progress` on the teardown that follows a scan timeout. By the time the error reaches bleak, BlueZ has already detached the discovery session and the kernel is no longer scanning. The error describes a failed acknowledgment, not a failed stop. bleak already tolerates `org.bluez.Error.NotReady` on this exact path (since 0.20.0) and retries `InProgress` on GATT reads and writes (#1282), but not here, so the exception propagates out of the `async with` in `find_device_by_filter()` and destroys the caller's result. This is the same symptom as #958, which was closed on the reporter's observation that a BlueZ upgrade made it go away. It is still present on 5.72, and the mechanism is now known from the BlueZ source.

### Reproduction

```python
device = await BleakScanner.find_device_by_address(addr, timeout=15.0)
```

with a device that is not currently advertising. The timeout fires, `find_device_by_filter()` returns `None` from its `except asyncio.TimeoutError`, the scanner's `__aexit__` calls `stop()`, and `StopDiscovery` is answered with `InProgress`:

```
  File "bleak/__init__.py", line 410, in find_device_by_filter
    async with cls(**kwargs) as scanner:
  File "bleak/__init__.py", line 171, in __aexit__
    await self.stop()
  File "bleak/backends/bluezdbus/scanner.py", line 144, in stop
    await stop()
  File "bleak/backends/bluezdbus/manager.py", line 503, in stop
    assert_reply(reply)
bleak.exc.BleakDBusError: [org.bluez.Error.InProgress] Operation already in progress
```

Field data from one production box: 22 occurrences in a day, every one on the timeout teardown, none on a scan that found its device. Scans are 15 s; the same scanner instance is never stopped twice.

### What BlueZ is actually reporting

Read against the BlueZ 5.72 tag (`src/adapter.c`, `src/error.c`).

1. The message "Operation already in progress" is `btd_error_busy`, which bluetoothd uses to relay any non-zero MGMT status from the kernel. The error name carries no information about which state produced it.

2. `StopDiscovery` is accepted by the D-Bus handler and forwarded to the kernel as `MGMT_OP_STOP_DISCOVERY`. The kernel rejects it (`MGMT_STATUS_REJECTED`) whenever `hci_discovery_active()` is false, that is, whenever the discovery state is anything other than `DISCOVERY_FINDING` or `DISCOVERY_RESOLVING`. So at the moment of rejection the kernel is, by definition, not scanning.

3. In `stop_discovery_complete()`, bluetoothd calls `discovery_remove(client)` unconditionally, before it checks the status. The client's discovery session is therefore already gone when the error reply is sent. This is consistent with the field behaviour: the same D-Bus sender can issue `SetDiscoveryFilter` plus `StartDiscovery` immediately afterwards and it succeeds, which would be refused with `InProgress` if the session were still registered.

4. The early return on the error path skips `adapter->discovering = false`, so the `Discovering` property stays true with no clients until a later successful stop or a power cycle. That is bluez/bluez#807, closed without a fix. It is worth knowing because anything that polls `Discovering` to decide whether to stop will be misled after this happens.

A plausible trigger for the timeout-teardown correlation, unconfirmed: the kernel disables an LE scan on its own after `DISCOV_LE_TIMEOUT` (10.24 s) and bluetoothd re-arms it a few seconds later, so a stop that lands in that gap meets `DISCOVERY_STOPPED` and is rejected. A scan that finds its device stops early, inside the first window.

### What bleak leaves behind

In `BlueZManager.active_scan()` the `stop` closure removes the advertisement and device-removed callbacks before calling `StopDiscovery`, then `assert_reply()` raises. Two consequences:

* The `SetDiscoveryFilter({})` reset sits in the `else` of that `assert_reply`, so it never runs on this path.
* `BleakScannerBlueZDBus.stop()` sets `self._stop = None` before awaiting the closure, so a second `stop()` from the caller is a silent no-op. There is no way for a caller to retry the stop.

The scanner itself is fine afterwards, since a new scan builds fresh callbacks. The damage is entirely that a completed scan raises.

### Proposal

In the `stop` closure in `bleak/backends/bluezdbus/manager.py`, treat `org.bluez.Error.InProgress` the same way `org.bluez.Error.NotReady` is already treated:

```python
try:
    assert_reply(reply)
except BleakDBusError as ex:
    if ex.dbus_error not in (
        defs.BLUEZ_ERROR_NOT_READY,
        "org.bluez.Error.InProgress",
    ):
        raise
    logger.debug("StopDiscovery answered %s; the session is already gone", ex.dbus_error)
# and run the SetDiscoveryFilter reset unconditionally after the try
```

Rationale: in both cases the outcome the caller asked for, that discovery ends for this client, has already happened. Raising turns a successful timeout into an exception, and the caller cannot retry because `_stop` has been cleared. Swallowing is also what bleak already does for `NotReady` on the same line, and what #1282 established as the precedent for `InProgress` on reads and writes.

I would not suggest a retry instead of a swallow: a second `StopDiscovery` from a sender with no session returns `org.bluez.Error.Failed: No discovery started`, so there is nothing for a retry to do.

If it helps, I can open a PR with the change above plus a test that fakes an `InProgress` reply from `StopDiscovery` and asserts that `stop()` returns cleanly and still resets the filter.

Related: #958 (same traceback, closed 2022), #1689 (generic InProgress reports, mostly StartDiscovery), #1282 (InProgress retry on GATT operations), bluez/bluez#807 (Discovering left true).

cgoudie:
Fixes #2021.

### The change

In the BlueZ backend, `BleakScanner.stop()` now treats `org.bluez.Error.InProgress` from `StopDiscovery` the same way it already treats `org.bluez.Error.NotReady`: the error is swallowed and `stop()` returns normally. One condition and a comment; no other behaviour changes. The `SetDiscoveryFilter({})` reset stays in the `else` branch exactly as it does for `NotReady`, so this path skips it too; the next scan sets its own filter, so nothing depends on it.

### Why swallowing is correct

`InProgress` on `StopDiscovery` is bluetoothd relaying `MGMT_STATUS_REJECTED` from the kernel, which the kernel returns whenever discovery is not in the `FINDING` or `RESOLVING` state (`net/bluetooth/mgmt.c: stop_discovery`, `hci_core.c: hci_discovery_active`). Before it replies, bluetoothd has already run `discovery_remove(client)` in `src/adapter.c: stop_discovery_complete()`, so our session is gone when the error arrives. The stop we asked for has happened in every sense that matters to the caller. Raising turns a completed scan into an exception, and the caller cannot retry because the scanner nulls `_stop` before awaiting it.

The points raised in the issue thread, one at a time.

### 1. "I would rather see an integration test that does it for real"

I tried, and I could not make one that fails without the fix, so I have not included one. The reason is in how the two halves interact. bluetoothd forwards a client's `StopDiscovery` to the kernel only while its own `discovery_enable` flag is set (`discovery_stop`, whose idle branch returns without any MGMT round trip otherwise), and it clears that flag when it processes the kernel's `Discovering(false)` event (`discovering_callback`). So the kernel's rejection needs a stop to be forwarded after the kernel has left `FINDING` (for an LE-only scan, at its own `DISCOV_LE_TIMEOUT` of 10.24 s) but before bluetoothd has processed that event. That is a scheduling race inside bluetoothd, not a time offset a test can aim at. Measured on BlueZ 5.72: stops at 10, 12, 15, 15.2, 15.4 and 16 s after start were accepted 18 of 18 times, and five `find_device_by_address(..., timeout=15.0)` teardowns stopped cleanly. I also tried inducing it from the vhci controller by holding the reply to bluetoothd's automatic restart so the kernel sits in `STARTING` while `stop()` runs; it does not work, because bluetoothd only sets `discovery_enable` in `start_discovery_complete`, so during its own restart it still reads idle and stops cleanly without asking the kernel. What is included instead is a unit test that fakes the bus reply (see Test below). It pins bleak's handling; it does not exercise BlueZ.

### 2. "We should just add some comments to it that link to the BlueZ bug or code that causes the problem"

Done. The comment above the condition names the bluetoothd path (`stop_discovery_complete()` calls `discovery_remove()` before checking the status, then relays `MGMT_STATUS_REJECTED` as `InProgress`) and links this issue and bluez/bluez#807.

### 3. "It sounds like we should make a patch to BlueZ to fix this"

Agreed, and I intend to send one to linux-bluetooth after this lands. The shape: in `stop_discovery_complete()`, treat `MGMT_STATUS_REJECTED` as "already stopped" rather than returning early, so `discovery_enable`, `discovering` and the `Discovering` property are reset the same way they are on success. bleak's change stands on its own regardless, since every 5.x user gets the error as things are.

### 4. "I guess this means we can't listen for a property change to tell us that the scanning stopped early"

Correct, twice over. When the kernel stops an LE scan on its own timer, `discovering_callback` only flips bluetoothd's internal `discovery_enable`; `adapter->discovering` stays true and no `PropertiesChanged` is emitted, so the early kernel stop is invisible over D-Bus. And after a failed `StopDiscovery`, the early return leaves `Discovering` true with no clients (bluez/bluez#807), so there is nothing to subscribe to for this case either.

### 5. "Also see #1146 which is related to catching that scanning stopped before we requested it"

Related but a different case. From bleak's side the scan never stopped early: bluetoothd kept our session and cycled the kernel scan internally, and the error arises at our own requested stop. An early-stop hook would not fire (no property change, per point 4) and is not needed for this fix. The two are complementary.

### 6. "It seems like it could affect starting the next scan since the discovering property is now wrong?"

Yes. After the failed stop, `update_discovery_filter()` sees equal filters and `discovering != 0` and returns without issuing a kernel start, so the next `StartDiscovery` on that adapter is accepted but does not actually scan, until a later successful stop or a power cycle resets the state. In our field data this showed up as clusters of consecutive timeouts on one adapter. This PR does not fix that; the bluetoothd patch in point 3 does. It only stops bleak from raising on the part that already succeeded.

### What we measured

24 occurrences on one Cerbo GX (BlueZ 5.72) in a single day, every one on the `find_device_by_address(..., timeout=15.0)` teardown for an absent device, none on a scan that found its device early. bluetoothd was crash-looping that day for an unrelated reason, which is consistent with the race widening under load; in the seven days since that was fixed there have been none. #958 and #1689 show other users hitting it. So the condition is real and provable from source, intermittent in practice, and not reproducible on demand.

### Test

`tests/backends/bluezdbus/test_issue_2021.py` runs `active_scan()` against a fake bus that answers `StopDiscovery` with `org.bluez.Error.InProgress`, and asserts that `stop()` returns, that the advertisement and device-removed callbacks are removed, and that the three calls went out in order. A second case answers with `org.bluez.Error.Failed` and asserts that still raises `BleakDBusError`, so the swallow stays scoped to the one error whose semantics are established above. It needs `dbus_fast` and skips where that is not installed.


cgoudie:
Updated for review. The PR description now answers each of your points under its own heading, so I will only summarise here.

On the integration test: I tried to make one that fails without the fix and could not. bluetoothd only forwards a client's stop to the kernel while its own `discovery_enable` flag is set, and clears it when it processes the kernel's `Discovering(false)` event, so the rejection needs a stop to land in the gap between the two. That is a scheduling race, not a time offset: stops at 10, 12, 15, 15.2, 15.4 and 16 s after start were accepted 18 of 18 times on 5.72, as were five `find_device_by_address` timeouts. Holding the vhci controller's reply to bluetoothd's restart does not help either, because bluetoothd stops cleanly whenever it still reads idle. So the PR carries a unit test that fakes the `InProgress` reply and pins bleak's handling, plus the negative case that other errors still raise. If you see a way to reach the race from the harness, I will gladly add it.

The code comment now names the bluetoothd path as well as the bug. On your question about the next scan: yes, it is affected, and the description spells out how; that part is bluetoothd's and I will send a patch to linux-bluetooth after this. Marking ready for review.

dlech:
> On your question about the next scan: yes, it is affected

Hmm... then ignoring this error might will just lead to a silent scan not working problem later, correct? I'm tempted to say, let's keep the error just to let people know bluetoothd is broken and probably needs to be restarted.

cgoudie:
Good question, and the answer is that there are two different situations here that produce the same `InProgress`, and this PR is about only the first one.

1. The one-off case. The kernel stops an LE scan on its own after `DISCOV_LE_TIMEOUT` (10.24 s). If our `stop()` lands just after that, the kernel answers the redundant stop with REJECTED, and bluetoothd has already removed our client in `stop_discovery_complete` before it relays the error. So our discovery session is gone, the scan we asked to end has ended, and the very next `StartDiscovery` from the same connection starts a fresh scan normally. This is the common case, it is transient, and scanning is not broken afterward. Swallowing the error here is correct: the operation succeeded, and raising turns a successful stop into a spurious exception the caller cannot retry (the scanner nulls `_stop` before awaiting it).

2. The stuck case. Separately, bluetoothd can get into a state where `discovery_enable` stays 1 while the kernel is `DISCOVERY_STOPPED`, and then it stays there: every last-client stop is rejected and never clears the belief, and every new filtered start takes the "queue a stop" branch whose completion callback is NULL, so no MGMT start is ever sent again. The card then scans nothing until it is reset. We have observed this on our own hardware (a card that ran 192 clean scan windows, then returned `InProgress` on every stop and moved zero RX bytes across nine accepted windows afterward). This is the case you are worried about, and you are right that it needs the adapter reset.

The point I would make is that raising the error does not help with case 2. bleak sees the identical `InProgress` in both, so raising cannot tell them apart: it would fail the common, successful, self-recovering case in order to surface the rare stuck one, and even then the generic `InProgress` does not tell a user "restart bluetoothd." The stuck state is a bluetoothd bug (it never clears `discovering` on a rejected stop, bluez/bluez#807), and the real fix is the BlueZ patch I plan to send: clear the belief on REJECTED, which both makes case 1 a non-event and stops case 2 from ever latching. bleak swallowing is the right interim behavior for case 1 today.

If a caller does want to detect the stuck state, the reliable signal is not a single `InProgress` but a persistent one paired with an empty scan, which is an application-level policy rather than something `stop()` should decide. Happy to add a `logger.warning` on the swallow so the condition is at least visible in logs without failing the call, if you would like that.

dlech:
> 1. The kernel stops an LE scan on its own after `DISCOV_LE_TIMEOUT` (10.24 s).

Does this mean that any scan that a user requests in Bleak longer than this gets cut short and the rest of the time it is not actually scanning?



> bleak sees the identical `InProgress` in both, so raising cannot tell them apart:

Can we look at the `Discovering` property? Even in cases where there is more than one client that has requested a scan, does this timeout stop them all? If that is so, then we might have an out here.

> Happy to add a `logger.warning` on the swallow so the condition is at least visible in logs without failing the call, if you would like that.

This sounds like the best we can do if the "Discovering" property can't be used. So info is probably good enough. And we can add something to the docs about the known bug in BlueZ.

If the "Discovering" property is false at this point though, we could probably demote it from a warning to a debug since that likely means that the next scan will work. Would still want it to be warning in other case though.

cgoudie:
> Does this mean that any scan that a user requests in Bleak longer than this gets cut short and the rest of the time it is not actually scanning?

Not cut short, but not continuous either. When the kernel's `DISCOV_LE_TIMEOUT` ends the LE scan it sends `MGMT_EV_DISCOVERING` with `discovering = 0`. bluetoothd's `discovering_callback` handles that: if any discovery clients remain it re-arms the scan with `trigger_start_discovery(adapter, IDLE_DISCOV_TIMEOUT)`, where `IDLE_DISCOV_TIMEOUT` is 5 seconds (`src/adapter.c`, unchanged between 5.72 and current master). So a 60 second `BleakScanner` session on BlueZ is really a train of roughly 10 second scan windows separated by roughly 5 second idle gaps, for as long as at least one client is registered. Advertisements that arrive during a gap are missed, and a device advertising at a slow interval can take a couple of windows to show up, but the session does run for the full duration the caller asked for.

The kernel timeout is per adapter, not per client. All clients share one kernel discovery, so it ends for all of them at once and bluetoothd restarts it for all of them together as long as any remain.

> Can we look at the `Discovering` property?

I checked the path in `adapter.c`, and it does not separate the two cases at the moment we see the error. `stop_discovery_complete` sends our `InProgress` reply first and only clears `adapter->discovering` and emits the `Discovering` change on the success path; on the error path it returns before that. `discovering_callback`, which handles the kernel's `discovering = 0` event, updates only the internal `discovery_enable` flag and never touches `adapter->discovering` or emits a property change. So after a rejected stop the `Discovering` property reads `true` in both the one-off case and the stuck case. That is exactly bluez/bluez#807. A rule of the form "demote to debug when `Discovering` is false" would therefore never fire.

What does separate them is what happens on the next stop, and that gives us a clean test that needs no property read.

In the one-off case the kernel's `discovering = 0` event lands a moment after our rejected stop and `discovery_enable` drops to 0. The next `StartDiscovery` with the same filter is accepted but does not reach the kernel, because `update_discovery_filter` returns early when the filters are equal and `adapter->discovering` is still set. That scan sees nothing. Its `StopDiscovery` then takes the idle branch of `discovery_stop` (`discovery_enable == 0`), which succeeds, clears `discovering` and emits the property change, and the scan after that runs normally. So one rejection is always followed by a successful stop, and the cost is one empty scan.

In the stuck case `discovery_enable` stays 1, so every later start takes the "queue up a stop" branch in `start_discovery_timeout` whose completion callback is NULL, and every last-client stop is rejected again. No scan reaches the kernel until the adapter is reset.

So two `InProgress` rejections in a row on one adapter, with no successful stop between them, cannot come from the race and can only mean the stuck state. That is what the PR now does:

- The first rejection is swallowed, logged at INFO, and remembered per adapter on the `BlueZManager`.
- A successful `StopDiscovery` clears that memory.
- A second rejection with the memory still set is logged at WARNING, naming the adapter and the remedy (power cycle or re-plug the adapter, or restart bluetoothd), and the original `BleakDBusError` is raised, so the caller learns the adapter is not scanning rather than getting silent empty scans.

That gives you the split you asked for, INFO for the harmless case and a raised error for the broken one, and keeps your original point that a broken bluetoothd should not be hidden. Two limits, both stated in the new troubleshooting section: the memory is per process, so a process that scans once never escalates; and if another scanner is active on the same adapter, bluetoothd answers our stop without asking the kernel, which counts as a success, so the detection only works while we are the last client. The tests cover the first rejection, the second consecutive one raising with the warning, and a clean stop between two rejections resetting the count.

The one empty scan in the harmless case is bluetoothd's bug and stays. Clearing `discovering` when the kernel rejects the stop is the fix on that side, and I will send it to linux-bluetooth as a separate patch once this lands.

bleak/backends/bluezdbus/manager.py
```diff
@@ -502,9 +505,51 @@ async def stop() -> None:
...
+                            # because it is not actively scanning at that
+                            # moment (already stopped, or mid-restart); BlueZ
+                            # has already removed our discovery session by the
+                            # time it replies, so nothing is left to stop:
+                            # bluetoothd's stop_discovery_complete() calls
+                            # discovery_remove() before checking the status,
+                            # then relays MGMT_STATUS_REJECTED as InProgress.
+                            # See https://github.com/hbldh/bleak/issues/2021
+                            # and https://github.com/bluez/bluez/issues/807.
+                            if ex.dbus_error not in (
```
dlech, line 518: I would change this to a match statement.

Indent is getting pretty deep here too, so maybe put everything in the `except` in a function to help with readability?

CHANGELOG.rst
```diff
@@ -21,6 +21,7 @@ Changed
 Fixed
 -----
 * Fixed handling empty notification payloads in BlueZ backend when using "AcquireNotify". Fixes #1982.
+* Fixed ``BleakScanner.stop()`` raising ``BleakDBusError`` with ``org.bluez.Error.InProgress`` in BlueZ backend when the kernel had already stopped scanning. A single rejection is now logged and tolerated; two in a row on the same adapter still raise, since that means bluetoothd's discovery state is stuck. Fixes #2021.
```
dlech, line 24: I like to keep these short and let people go to the issue for the details.

dlech (commented):
This seems like the best we can do given what we know about the bug.

dlech:
I let Fable review this locally and it thinks the heuristic of failing twice is a bit fragile. And it offered the suggestion of rather trying to start a scan again as the test.

I.e. If we get the InProgress error, call StartDiscovery and StopDiscovery again right then. If that succeeds, then it should be the harmless case and if it fails, we know it is the stuck case.

That eliminates needing to track all of the things that happened between the first and second tries.

Issue #1278, opened 2023-04-13
Disconnect causes "Event loop is closed"


jacobbreen25:
* bleak version: 1.2.0
* Python version: 3.11.2
* Operating System: Windows 10

### Description

I am trying to use bleak without using the async context manager (I have found that with the bluetooth device I am using it causes a few issues with latency) as I will later use this to implement into a GUI. 

While using the disconnect function in the BleakClient, while it does disconnect, it consistently gives this error:

`Traceback (most recent call last):
    loop.call_soon_threadsafe(handle_session_status_changed, args)
    self._check_closed()
  File "C:\redacted\AppData\Local\Programs\Python\Python311\Lib\asyncio\base_events.py", line 519, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed`

### Code

Here is the program I ran to get this error

```python
from bleak import BleakScanner, BleakClient, backends 
import asyncio

class bluetooth:
    def __init__(self):
        self.bleClient = None
    async def callback(self, characteristic, d : bytearray):
        data_array = d.decode().rstrip().split('\t')
        print("Calling!")

    async def disconnect_callback(self, client):
        print("Diconnected")

    async def find_and_connect(self, name : str):
        devices = await BleakScanner.discover()
        for d in devices:
            if d.name == name:
                self.bleClient = BleakClient(d, disconnected_callback= lambda : self.disconnect_callback(), timeout=5.0)
                try:
                    await self.bleClient.connect()
                except Exception as e:
                    print(e)
                print("Connected")
                return
        raise Exception("No Device")

    async def retrieve(self):
        if self.bleClient != None:
            if self.bleClient.is_connected:
                print("is connected")
                try:
                    await self.bleClient.start_notify("eb523251-6513-11ed-9c9d-0800200c9a66".format(0xFFE1), self.callback)
                    await asyncio.sleep(0.1)
                    await self.bleClient.stop_notify("eb523251-6513-11ed-9c9d-0800200c9a66".format(0xFFE1))
                    await self.bleClient.disconnect()
                except Exception as e:
                    print(e)
            else:
                print("not connected")

ble = bluetooth()
try:
    asyncio.run(ble.find_and_connect("FlexibleSensorBLE-100"))
except Exception as e:
    print(e)
    print("Closing")
    exit
asyncio.run(ble.retrieve())
```

PR #2023, opened 2026-09-03
backends/winrt: use try_call_soon_threadsafe() in event handlers

nebkat:
WinRT event handlers run on a thread pool thread and hand off to the asyncio loop with `call_soon_threadsafe()` without checking that the loop is still open. One landing after the loop is closed raises `RuntimeError('Event loop is closed')` on a non-Python thread where nothing can catch it, so it surfaces as an unraisable exception at shutdown (under pytest, a `PytestUnraisableExceptionWarning`). The usual trigger is disconnect at teardown: `handle_disconnect()` removes the session-status token, but a callback already in flight races the loop shutdown and loses.

Same failure #1855 fixed for CoreBluetooth, so this applies the existing `try_call_soon_threadsafe()` helper to all six cross-thread hops in `client.py` and `scanner.py`.

Related: #1278, the same traceback from 2023.

https://claude.ai/code/session_01Y2188WRhSN8xiHTeKNDmoV

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -21,6 +21,7 @@ Changed
 Fixed
 -----
 * Fixed handling empty notification payloads in BlueZ backend when using "AcquireNotify". Fixes #1982.
+* Fixed crash in WinRT backend if a WinRT event handler is called after the asyncio event loop is closed.
 
 `3.0.2`_ (2026-05-02)
 =====================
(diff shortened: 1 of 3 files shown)


dlech:
Good one. I thought we had done this already. But I guess we made Bleak so good about cleaning up that we didn't hit the issue until #1768 change the behavior that leads to the reported observation.

PR #2029, opened 2026-09-11
Keep deprecated adapter arguments from mutating BlueZ defaults

### LINKED ISSUE #2028: BleakScanner/BleakClient: deprecated adapter= kwarg is written into the shared mutable bluez={} default, so every later scanner/client in the process inherits the first adapter
### Description

In 3.0.x the deprecation shim for the old `adapter=` keyword copies the
value into the `bluez` argument:

```python
# bleak/__init__.py, BleakScanner.__init__ (same block in BleakClient.__init__)
bluez: BlueZScannerArgs = {},
...
adapter_kwarg = kwargs.get("adapter")
if adapter_kwarg is not None:
    warn("the 'adapter' keyword argument is deprecated, ...", DeprecationWarning, stacklevel=2)
    if "adapter" not in bluez:
        bluez["adapter"] = adapter_kwarg
```

`bluez` defaults to a module-level `{}`, so when the caller did not pass
`bluez=` the assignment mutates the **shared default object**. From then
on every `BleakScanner()` / `BleakClient(addr)` constructed in the process
without an explicit `bluez=` silently runs on that first adapter, even
callers that never mentioned an adapter at all.

The DeprecationWarning is hidden by default outside `__main__`, so nothing
tells the user this happened.

### What happened to us

A long-running service pins its scanners to a USB dongle with
`adapter="hci1"` (the pre-3.0 spelling). When the dongle is unplugged the
service's supervisor re-runs discovery with **no** adapter argument,
expecting bleak's default (the first powered adapter, `hci0`). Instead every
new scanner raised `adapter 'hci1' not found` — `get_default_adapter()` was
never consulted because `bluez` already carried `{"adapter": "hci1"}` from
the polluted default. Passing `adapter="hci0"` explicitly did not help
either: `"adapter" in bluez` was already true, so the new value was ignored
and the scanner still targeted `hci1`.

### Minimal reproduction (Linux, no hardware or event loop needed)

```python
import inspect, warnings

warnings.simplefilter("ignore", DeprecationWarning)
scan_default = inspect.signature(BleakScanner.__init__).parameters["bluez"].default
client_default = inspect.signature(BleakClient.__init__).parameters["bluez"].default
print("before:", scan_default, client_default)

BleakScanner(adapter="hci1")                      # one old-style caller
BleakClient("AA:BB:CC:DD:EE:FF", adapter="hci1")

print("after: ", scan_default, client_default)
s2 = BleakScanner()                               # never asked for an adapter
print("second scanner adapter:", s2._backend._adapter)
```

Output on bleak 3.0.2:

```
before: {} {}
after:  {'adapter': 'hci1'} {'adapter': 'hci1'}
second scanner adapter: hci1
```

### Expected

A scanner or client constructed without `bluez=`/`adapter=` uses the
manager's default adapter regardless of what earlier callers passed.
Passing a different `adapter=` later should take effect.

### Suggested fix

Never mutate the caller's (or the default) dict:

```python
if adapter_kwarg is not None:
    warn(...)
    bluez = {"adapter": adapter_kwarg, **bluez}   # explicit bluez["adapter"] still wins
```

or use `bluez: Optional[BlueZScannerArgs] = None` and `bluez = dict(bluez or {})`
at the top of both constructors. The same pattern applies to
`BleakClient.__init__`. (`cb`, `winrt` and the `start_notify` `bluez`
defaults are also shared `{}` objects; they are not written to today, so
they are only a latent hazard.)

### Environment

- bleak 3.0.2 (also present on `develop` @ 764f40e)
- Raspberry Pi 5, Raspberry Pi OS, kernel 6.18.39+rpt-rpi-2712, BlueZ 5.82,
  Python 3.13.5 (also reproduced with Python 3.14 on the constructor path)
- Two adapters: onboard `hci0` + USB dongle `hci1`

Gonghan-Princess:
Fixes #2028.

The deprecated `adapter=` shim currently writes into the shared `bluez={}` default in both `BleakScanner` and `BleakClient`. After one instance requests `hci1`, later instances can inherit that adapter or ignore a new `adapter=` value. The same assignment also changes a caller-provided BlueZ dictionary.

Build a new dictionary when applying the deprecated argument, following the approach endorsed in #2028. An explicit `bluez["adapter"]` retains precedence. Add parameterized regression tests for both constructors and the required changelog entry. The tests replace backend constructors, so they do not need Bluetooth hardware or an event loop.

Validation:

- Before the fix: four mutation regression cases failed; two precedence cases passed.
- The six new cases pass on Windows CPython 3.10 and 3.14 with pytest 8.4.2 and warnings treated as errors.
- Complete adapter-deprecation test file on Python 3.12: 8 passed.
- Full Windows Python 3.12 suite: 19 passed, 103 skipped (hardware-dependent integration and other-platform tests).
- isort, Black, flake8, mypy (92 source files), Pyright, Sphinx with warnings treated as errors, and sdist/wheel builds passed.

The remaining full Python-version and OS matrix, including Linux BlueZ integration with real or virtual controllers, still needs CI. The tests here verify argument isolation before backend construction and do not exercise Bluetooth hardware.

AI assistance: OpenAI Codex helped investigate the issue, implement the maintainer-endorsed approach, write regression tests, run checks, and prepare this description.

### CI coverage follow-up

All 20 build/test jobs and the format/lint check passed. Codecov reports 100% coverage for the two modified coverable lines; `bleak/__init__.py` coverage increased from 80.64% to 81.39%.

The project check reports 52.95% against a 53.00% baseline. Its entire decrease comes from unchanged `bleak/backends/bluezdbus/manager.py`: line 1115 changed from fully to partially covered, and lines 1116–1117 changed from covered to uncovered. These lines handle an adapter `InterfacesRemoved` signal; the baseline Python 3.10 BlueZ integration run covered this cleanup path, while the current run did not. No coverage threshold or unrelated tests were changed.

The branch already includes current `develop`. Codecov's one-commit-behind warning refers to its coverage baseline: the intervening commit only changes `.readthedocs.yml` and does not trigger the build/test workflow. [Full Codecov comparison](https://app.codecov.io/gh/hbldh/bleak/pull/2029).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,6 +20,7 @@ Changed
 
 Fixed
 -----
+* Fixed deprecated ``adapter`` keyword argument mutating shared defaults and caller-provided ``bluez`` arguments in ``BleakScanner`` and ``BleakClient``. Fixes #2028.
 * Fixed handling empty notification payloads in BlueZ backend when using "AcquireNotify". Fixes #1982.
 
 `3.0.2`_ (2026-05-02)
--- a/bleak/__init__.py
+++ b/bleak/__init__.py
@@ -139,8 +139,7 @@ class BleakScanner:
                 stacklevel=2,
             )
 
-            if "adapter" not in bluez:
-                bluez["adapter"] = adapter_kwarg
+            bluez = {"adapter": adapter_kwarg, **bluez}
 
         self._backend = PlatformBleakScanner(
             detection_callback,
@@ -645,8 +644,7 @@ class BleakClient:
                 stacklevel=2,
             )
 
-            if "adapter" not in bluez:
-                bluez["adapter"] = adapter_kwarg
+            bluez = {"adapter": adapter_kwarg, **bluez}
 
         self._backend = PlatformBleakClient(
             address_or_ble_device,
--- a/tests/test_adapter_deprecation.py
+++ b/tests/test_adapter_deprecation.py
@@ -1,6 +1,65 @@
+from collections.abc import Callable
+from typing import Any
+from unittest.mock import Mock
+
 import pytest
 
 from bleak import BleakClient, BleakScanner
+
+
+@pytest.fixture(params=["scanner", "client"])
+def backend_bluez_args(
+    request: pytest.FixtureRequest, monkeypatch: pytest.MonkeyPatch
+) -> Callable[..., dict[str, Any]]:
+    backend = Mock()
+    monkeypatch.setattr(
+        f"bleak.get_platform_{request.param}_backend_type",
+        lambda: (backend, "test"),
+    )
+
+    def create(**kwargs: Any) -> dict[str, Any]:
+        if request.param == "scanner":
+            BleakScanner(**kwargs)
+        else:
+            BleakClient("00:11:22:33:44:55", **kwargs)
+        return backend.call_args.kwargs["bluez"]
+
+    return create
+
+
+def test_adapter_kwarg_does_not_change_defaults(
+    backend_bluez_args: Callable[..., dict[str, Any]],
+):
+    with pytest.deprecated_call(match="the 'adapter' keyword argument is deprecated"):
+        first = backend_bluez_args(adapter="hci1")
+    with pytest.deprecated_call(match="the 'adapter' keyword argument is deprecated"):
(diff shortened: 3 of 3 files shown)


dlech:
Thanks!

PR #2031, opened 2026-09-13
backends/bluezdbus: match device removed callbacks on the whole adapter path

bdraco:
Device removed callbacks are registered per adapter and matched against the removed object path with a plain prefix check, so a callback for `/org/bluez/hci1` also fires for devices on `hci10` through `hci19`. This surfaced while testing with a dozen adapters. The match now requires the trailing slash.

Split out of #2012 at the reviewer's request since it is a pre-existing issue. #2012 will need a small rebase once this merges.

Per the guidelines: CHANGELOG.rst is updated, a unit test covers the `hci1` versus `hci10` case, and the branch is a single commit.


dlech:
Thanks!

